PR #5, opened 2015-05-20
Updated readme with partial installation steps

thejoshcrawford:
I had problems getting this working because the embedded files weren't found so that's why step 8 is empty, but the majority of these steps should get someone started.


styfle:
Yes please :+1: 
I found this package by just adding to my project.json and it wasn't clear how to use it.
Then I found this github project but there was no "getting started" or wiki to be found.

PR #11, opened 2015-07-19
added swagger path details attribute to add more details (see description)

mburumaxwell:
added swagger path details attribute since we still have no means to generating XML files yet one may need the summary and implementation notes
In the image below, see the summary added and the implementation details

swagger path


domaindrivendev:
Xml files are now possible - https://github.com/domaindrivendev/Ahoy/blob/master/test/WebSites/Basic/Startup.cs#L51

mburumaxwell:
Okay. Great news to a one-year old PR.

PR #34, opened 2015-10-26
AppVeyor and Travis-CI scripts for CI

mcliment:
Added scripts for building and running tests on both CI platforms if you are interested. Once integrated badges can be added to the README.md file.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,15 @@
+language: csharp
+solution: Ahoy.sln
+sudo: false
+mono:
+  - latest
+install:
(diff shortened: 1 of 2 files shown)


cwe1ss:
This PR could be closed because it is dnvm/dnx based which no longer exists in RC2. We could create new scripts on top of #105 if you accept that and you want to integrate with AppVeyor/TravisCI

domaindrivendev:
See https://github.com/domaindrivendev/Ahoy/pull/105#issuecomment-232286924

PR #61, opened 2016-02-15
SwaggerGen : set byte format for properties of type byte[]

simonferquel:
When a model contains a byte[] property, the generated schema does not specify the format as byte, and thus, client generator Tools generate a model with a string property.
With this fix, generators like autorest generate byte arrays proxy models and correctly serialize them as base64


domaindrivendev:
Looks like this change was already made at some point - https://github.com/domaindrivendev/Ahoy/blob/master/src/Swashbuckle.SwaggerGen/Generator/SchemaRegistry.cs#L240

PR #65, opened 2016-02-19
format "uuid" for System.Guid to allow clients to generate proxies with Guid

cwe1ss:
Right now, Guid parameters are rendered with type "string" and with no "format". This pull request adds "uuid" for the format field. This allows clients tools to generate it back to Guid. I used "uuid" because that's the example from the specification. "Guid" is just one implementation of uuid. https://en.wikipedia.org/wiki/Globally_unique_identifier

this could also be backported to the older versions ( https://github.com/domaindrivendev/Swashbuckle/issues/527 )

### DIFF
--- a/src/Swashbuckle.SwaggerGen/Generator/SchemaRegistry.cs
+++ b/src/Swashbuckle.SwaggerGen/Generator/SchemaRegistry.cs
@@ -208,7 +208,8 @@ namespace Swashbuckle.SwaggerGen.Generator
             { typeof(sbyte), () => new Schema { Type = "string", Format = "byte" } },
             { typeof(bool), () => new Schema { Type = "boolean" } },
             { typeof(DateTime), () => new Schema { Type = "string", Format = "date-time" } },
-            { typeof(DateTimeOffset), () => new Schema { Type = "string", Format = "date-time" } }
+            { typeof(DateTimeOffset), () => new Schema { Type = "string", Format = "date-time" } },
+            { typeof(Guid), () => new Schema { Type = "string", Format = "uuid" } }
         };
     }
 }
--- a/test/Swashbuckle.SwaggerGen.Test/Generator/SchemaRegistryTests.cs
+++ b/test/Swashbuckle.SwaggerGen.Test/Generator/SchemaRegistryTests.cs
@@ -30,6 +30,7 @@ namespace Swashbuckle.SwaggerGen.Generator
         [InlineData(typeof(DateTime), "string", "date-time")]
         [InlineData(typeof(DateTimeOffset), "string", "date-time")]
         [InlineData(typeof(string), "string", null)]
+        [InlineData(typeof(Guid), "string", "uuid")]
         public void GetOrRegister_ReturnsPrimitiveSchema_ForSimpleTypes(
             Type systemType,
             string expectedType,


RehanSaeed:
Could really use this feature right now. Pasting a GUID in where Ahoy says string is getting real old.

daviddunkley:
Any update on when this VERY useful change will be released?

gitfool:
@daviddunkley You can use the following in the meantime:
```c#
config.EnableSwagger(x =>
{
    x.MapType<Guid>(() => new Schema { type = "string", format = "uuid" });
});
```
Alas, it gets even more complicated if you want to distinguish `Guid` from `Guid?` (Nullable\<Guid\>).

daviddunkley:
@gitfool many thanks for the suggestion. It works well and now my AutoRest Clients can actually have the correct types!

For completeness I got Guid? working like this...
```
var vendorExt = new Dictionary<string, object> { { "x-nullable", true } };
c.MapType<Guid?>(() => new Schema
{
    type = "string", format = "uuid", vendorExtensions = vendorExt
});
```

gitfool:
@daviddunkley I do exactly the same thing, but I'm also trying to map all primitives, including all common value types, which is currently missing `System.TimeSpan`:
```c#
config.EnableSwagger(x =>
{
    var xnullable = new Dictionary<string, object> { { "x-nullable", true } };
    x.MapType<bool?>(() => new Schema { type = "boolean", vendorExtensions = xnullable });
    x.MapType<byte?>(() => new Schema { type = "integer", format = "int32", vendorExtensions = xnullable });
    x.MapType<sbyte?>(() => new Schema { type = "integer", format = "int32", vendorExtensions = xnullable });
    x.MapType<short?>(() => new Schema { type = "integer", format = "int32", vendorExtensions = xnullable });
    x.MapType<ushort?>(() => new Schema { type = "integer", format = "int32", vendorExtensions = xnullable });
    x.MapType<int?>(() => new Schema { type = "integer", format = "int32", vendorExtensions = xnullable });
    x.MapType<uint?>(() => new Schema { type = "integer", format = "int32", vendorExtensions = xnullable });
    x.MapType<long?>(() => new Schema { type = "integer", format = "int64", vendorExtensions = xnullable });
    x.MapType<ulong?>(() => new Schema { type = "integer", format = "int64", vendorExtensions = xnullable });
    x.MapType<float?>(() => new Schema { type = "number", format = "float", vendorExtensions = xnullable });
    x.MapType<double?>(() => new Schema { type = "number", format = "double", vendorExtensions = xnullable });
    x.MapType<decimal?>(() => new Schema { type = "number", format = "double", vendorExtensions = xnullable });
    x.MapType<DateTime?>(() => new Schema { type = "string", format = "date-time", vendorExtensions = xnullable });
    x.MapType<DateTimeOffset?>(() => new Schema { type = "string", format = "date-time", vendorExtensions = xnullable });
    x.MapType<TimeSpan>(() => new Schema { type = "string", format = "time-span" });
    x.MapType<TimeSpan?>(() => new Schema { type = "string", format = "time-span", vendorExtensions = xnullable });
    x.MapType<Guid>(() => new Schema { type = "string", format = "uuid" });
    x.MapType<Guid?>(() => new Schema { type = "string", format = "uuid", vendorExtensions = xnullable });
    x.SchemaFilter<RequiredSchemaFilter>();
})
```
This needs to be combined with the following to manipulate the Swagger doc to get better code generation with AutoRest:
```c#
public static class SchemaExtensions
{
    public static bool IsNullable(this Schema schema)
    {
        object value;
        return (schema.vendorExtensions != null) && schema.vendorExtensions.TryGetValue("x-nullable", out value) &&
            (value is bool) && (bool)value;
    }
}

public class RequiredSchemaFilter : ISchemaFilter
{
    public void Apply(Schema schema, SchemaRegistry schemaRegistry, Type type)
    {
        var required = schema.properties
            .Where(x => !x.Value.IsNullable() &&
                ((x.Value.type == "boolean") || (x.Value.type == "integer") || (x.Value.type == "number") ||
                (x.Value.format == "date-time") || (x.Value.format == "time-span") || (x.Value.format == "uuid") ||
                (x.Value.@enum != null)))
            .Select(x => x.Key)
            .ToArray();

        if (required.Length > 0)
        {
            if (schema.required == null)
            {
                schema.required = new List<string>();
            }
            foreach (var property in required)
            {
                schema.required.Add(property);
            }
        }
    }
}
```
Now I'm closer to where I want to be, but there's still room for improvement:

- I'm not sure it's safe to invent your own format, so maybe `TimeSpan` should be done using another vendor extension, say `x-format`, so it would become:
```c#
var xformatTimeSpan = new Dictionary<string, object> { { "x-format", "time-span" } };
var xformatNullableTimeSpan = new Dictionary<string, object> { { "x-format", "time-span" }, { "x-nullable", true } };
x.MapType<TimeSpan>(() => new Schema { type = "string", vendorExtensions = xformatTimeSpan });
x.MapType<TimeSpan?>(() => new Schema { type = "string", vendorExtensions = xformatNullableTimeSpan });
```

- This does not handle nullable enums. This should be handled generally for all enums, and that can only happen by modifying [Swashbuckle.Core/Swagger/SchemaRegistry.cs](https://github.com/domaindrivendev/Swashbuckle/blob/master/Swashbuckle.Core/Swagger/SchemaRegistry.cs#L134), since 

- The `RequiredSchemaFilter` helps direct AutoRest to generate better code, but `required` and `nullable` are not exactly the same thing. (See OpenAPI-Specification issue [#229](https://github.com/OAI/OpenAPI-Specification/issues/229) for related discussions.)

@domaindrivendev Would you accept a PR to change [SchemaRegistry.CreatePrimitiveSchema](https://github.com/domaindrivendev/Swashbuckle/blob/master/Swashbuckle.Core/Swagger/SchemaRegistry.cs#L129) to also include `x-nullable` and `x-format` (or similar for `TimeSpan`)? It should be harmless to add these vendor extensions and will enable much better code generation. In fact, given such changes, I could submit a PR to enlighten AutoRest to recognize them. :wink:

gitfool:
@domaindrivendev Any thoughts on the above so I know what to do next? I'm currently on Swashbuckle, not Ahoy. Does it make sense to fork Swashbuckle and submit a PR now with the above suggestions, or is the Ahoy merge imminent and I should wait?

gitfool:
After further thought and experimentation, I created a PR anyway (see above) to put this all together and illustrate the primitive type enlightenment I'm after.

draptik:
The suggestion using `MapType` doesn't work for me. 

Here is my sample REST call:
```
public string Get(Guid id)
{
       return "value";
}
```

And in SwaggerConfig:
```
config.EnableSwagger(c => {
    // ...
    c.MapType<Guid>(() => new Schema { type = "string", format = "uuid" });
}
// ...
```
The GUID property is still displayed as data type string.

Am I using the method correctly?

gitfool:
@draptik Do you mean it still displays as a string in the Swagger UI? `MapType` won't change that AFAIK; it only works for code generation tools that look for `uuid` format strings.

draptik:
@gitfool Thanks for the quick reply. Yes, that is what I meant: The UI still displays the GUID as data type string.
image

Is there a way to change this?

RehanSaeed:
@gitfool You can use an `IModelFilter` for that:

```
[SwaggerModelFilter(typeof(LogModelFilter))]
public class Log
{
    public Guid Id { get; set; }
}

public class LogModelFilter : IModelFilter
{
    public void Apply(Schema model, ModelFilterContext context)
    {
        model.Default = new Log()
        {
            Id = Guid.NewGuid()
        };
    }
}
```

draptik:
@RehanSaeed Thanks! Your suggestion using a `SwaggerModelFilter` did not work for me. According to https://github.com/domaindrivendev/Swashbuckle#transitioning-to-swashbuckle-50 the `IModelFilter` was replaced by a `ISchemaFilter`. Is there a different approach to map Guids without creating a filter for each model having a Guid?

cwe1ss:
seems like my original PR was already integrated. I guess the general topic about further types etc should go into a separate issue/PR.

fleed:
Any answer to last @draptik question? Is it possible to correctly handle `Guid` properties out of the box or should I use @RehanSaeed 's suggested _workaround_?

PR #71, opened 2016-03-09
Upgrade to ASP.NET Core RC2 Daily Builds

cwe1ss:
I updated Swashbuckle to work against a current RC2 daily build. The tests don't work though because I wasn't able to get a working dnx runner for the daily builds. 

I currently have my own feed for this but it would be great if you could create official daily build versions of Swashbuckle as well. I guess more and more people will switch to a daily build of ASP.NET Core because the RC1 release is already quite old and there have been many breaking changes since then.


cwe1ss:
closed since there are already two newer PRs :)

PR #73, opened 2016-03-24
Fix Response Headers collection type to match Swagger spec

kyschouv:
Changes `Response.Headers` collection type to `IDictionary<string, Header>`. This fixes the output `SwaggerDocument` to match the Swagger schema.


kyschouv:
@domaindrivendev Any chance you can merge this pull? I'd like to return to mainline builds =).

PR #80, opened 2016-04-25
Fixed ref schema error message

tillig:
Updated the error message used in CreateRefSchema to indicate the fix for duplicate schema IDs given the new configuration mechanism.


domaindrivendev:
This issue was addressed with this commit - bd5d8f1cbd13651ba6664343a674900566fdd74c

PR #88, opened 2016-05-22
.NET Core RC2 update including UnitTests

TsengSR:
Today I did a more complete port to RC2. 

The previous pull requests didn't ported the UnitTests. I have managed to get all the tests pass, except for 2.

I had to use RC3 of xunit runner, as the RC2 contains a but where tests fail on x64 as mentioned in this issues

https://github.com/xunit/xunit/issues/843
https://github.com/dotnet/cli/issues/3069#issuecomment-220332924
https://github.com/xunit/xunit/issues/850#issuecomment-220832417

Failing Test is GetSwagger_SetsResponseStatusCodeAndSchema_AccordingToActionReturnType with `[InlineData(nameof(FakeActions.ReturnsActionResult), "200", false)]` and `[InlineData(nameof(FakeActions.ReturnsVoid), "204", false)]`. I don't know exactly why they fail, but the `SupportedResponseTypes`in SwaggerProvider.cs, line 138 is empty and the loop never entered so the call to `responses.Add(...)` is never called and the array in the test is null. 

I wasn't sure what the expected behaviour would be, when there is more than one supported type or none at all, so I left it out as is so @domaindrivendev can have a look at it.

Hope you'll be able to get it compile on your machine (I got VS 2015 Update 2 with .NET Core Preview Tooling - took me a while to get the old tools uninstalled and get rid of dnx remain). 

I was unable to find a way to replace the `_appEnv.Configuration,`, so for now I used preprocessor directives. It's not very clean way to solve it, but `IApplicationEnvironment` was removed/placed into the static  `PlatformServices` class, but there is no `Configuration` property anymore.

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,7 +1,7 @@
 {
   "sources": [ "src", "test" ],
   "sdk": {
-    "version": "1.0.0-rc1-final",
+    "version": "1.0.0-preview1-002702",
     "runtime": "clr",
     "architecture": "x86"
   },
--- a/src/Swashbuckle.SwaggerGen/Annotations/XmlCommentsOperationFilter.cs
+++ b/src/Swashbuckle.SwaggerGen/Annotations/XmlCommentsOperationFilter.cs
@@ -1,6 +1,6 @@
 ﻿using System.Linq;
 using System.Xml.XPath;
-using Microsoft.AspNet.Mvc.Controllers;
+using Microsoft.AspNetCore.Mvc.Controllers;
 using Swashbuckle.SwaggerGen.Generator;
 
 namespace Swashbuckle.SwaggerGen.Annotations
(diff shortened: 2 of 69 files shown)


gencebay:
You can check it!
https://github.com/gencebay/Ahoy/blob/6.0.0-rc2-final/src/Swashbuckle.SwaggerGen/Generator/ApiDescriptionExtensions.cs

shep1987:
There is a slight bug in the SwaggerUI project leading to the ui not being embedded
Your buildOptions in project.json should be like this:

"buildOptions": {
    "embed": {
      "include": [ "SwaggerUi", "bower_components/swagger-ui/dist" ]
    }
  },

Thanks for the work on this!

TsengSR:
@shep1987: Thanks. Fixed it and added a test to verify it. Now only the `[InlineData(nameof(FakeActions.ReturnsActionResult), "200", false)]` test is failing

domaindrivendev:
Closed in favor of #89

PR #89, opened 2016-05-24
RC2 Migration and Updates

gencebay:


domaindrivendev:
Thanks for this. Did you get all tests passing? I just pulled down and there still appears to be a lot of red tests. Also, looks like the Integration tests aren't working either

gencebay:
Thanks, reviewed and commit! should work now.

gencebay:
Unit Tests are ok! and also added Enum IdTextPair serialization and show it on Model - Model Schema sections on Swagger UI but I could not find a proper way to set enum integer value as default when click for set as parameter value. [Tests - GetOrRegister_SupportsOptionToDescribeAllEnumsAsIdPair]

domaindrivendev:
Awesome thanks for this. I omitted the enum description changes because they break the semantics of the JSON Schema and hence the Swagger specification. The schema enum property should be the "valid values" (http://json-schema.org/latest/json-schema-validation.html#anchor76) for the JSON property that's being described.

Let's continue this discussion in a separate ticket. For now let's focus on RC2 support. Thanks again!

src/Swashbuckle.SwaggerGen/Annotations/SwaggerAttributesOperationFilter.cs
```diff
@@ -2,6 +2,7 @@
 using System.Linq;
 using System.Net;
 using Swashbuckle.SwaggerGen.Generator;
+using Swashbuckle.SwaggerGen.Extensions;
```
domaindrivendev: The extensions (ApiDescriptionExtensions, JsonContractExtensions etc.) are specific to the "Generator" functionality and so I would like to keep them in the same namespace. It's just a styling preference - I prefer to namespace by functionality rather than technical groupings.

TsengSR:
Is this version packaged and pushed on the official nuget feeds? I don't see it there yet. Or at least a private myget feed?

gencebay:
Yes, you can use it as shown below
[https://channel9.msdn.com/Events/dotnetConf/2016/Building-Secure-Web-APIs-with-ASPNET-Core](dotnetConf) after on 0:21:15 in timeline

PR #95, opened 2016-06-05
Fix SwaggerProvider to use "formData"

RemyArmstro:
Swagger provider was output "form" when using [FromForm] annotation.  This was not valid Swagger spec.  Updated code to use "formData" instead.


domaindrivendev:
Went with #207

PR #103, opened 2016-06-18
Updates to project-files for RC2

cwe1ss:
Hi Richard,

I made a few updates to the project-files and to the wording:

* AspNet Core -> ASP.NET Core in docs/nuget package descriptions (that's the official wording)
* Nightly NuGet feeds for ASP.NET and XUnit are no longer necessary (everything is on NuGet)
* DNX -> DotNet in xproj files (see https://github.com/aspnet/Templates/blob/dev/src/BaseTemplates/StarterWeb/StarterWeb.xproj )
* launchSettings.json fixed ( https://github.com/aspnet/Templates/blob/dev/src/BaseTemplates/StarterWeb/Properties/launchSettings.json )
* Tests: Copy TestFixtures/XmlComments.xml to output so that "dotnet test" works properly

### DIFF
--- a/NuGet.config
+++ b/NuGet.config
@@ -1,8 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <configuration>
   <packageSources>
-    <add key="AspNetCiRelease" value="https://www.myget.org/F/aspnetcirelease/api/v3/index.json" />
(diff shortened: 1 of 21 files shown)


cwe1ss:
I close this since the code is already on RTM. I'll check if there are some parts of this PR that are still relevant and create a separate PR if so.

PR #104, opened 2016-06-18
*.user files should not be in git

cwe1ss:
People use different settings which leads to unnecessary commits


cwe1ss:
this was included in the merged RTM pull request

PR #105, opened 2016-06-18
psake-based build script for dotnet cli install, restore, build, test & package

cwe1ss:
Adds a psake (PowerShell) based build script that installs the dotnet-cli (if not present), restores, builds, tests and packs everything. All artifacts (test results, nuget packages) will be placed in an "artifacts"-folder.

The build process can be started by calling `.\build.ps1` from a PowerShell window.

Notes:
* psake 4.6.0 was taken from https://github.com/psake/psake/releases/tag/v4.6.0
* In order to get a successful build, this requires #103 (fixed nuget.config for dotnet-restore, fixed XmlComments.xml). Just remove "dotnet-test" from the "Default"-Task in build-definition.ps1 (line 29) if you want to run it without tests.
* I moved the "Swashbuckle" project from nuspec to project.json to be consistent with other projects and to get the nuget-packages through the same pipeline.
* It does not yet set a dynamic version number because the projects don't use the *-character right now. I would fix this, if you want to use this PR.
* This build script is not x-plat because it requires PowerShell

As soon as you have such a build script, you could easily integrate this into e.g. AppVeyor (which is free for open-source) to have Continuous Integration. AppVeyor would also give you an easy way to publish the packages to NuGet.org


cwe1ss:
rebased for RTM

domaindrivendev:
Hey there - thanks for this build script. Would you be interested in submitting a PR to get integrated with AppVeyor/Travis-CI?

PR #106, opened 2016-06-18
Respect PathBase for Swagger-UI.

cwe1ss:
The SwaggerMiddleware already adds the Request.PathBase property to its path. The SwaggerUi-project does not yet do this. This results in a broken UI because the path to swagger can't be resolved properly.

I've added a simple middleware to the Basic-website to test this. This middleware allows to make a request with a "X-Forwarded-PathBase" header, which gets set as the PathBase.

We have this exact setup with our APIs. They are hosted behind a reverse-proxy which sends the path under which they are hosted as a HTTP header.


cwe1ss:
rebased for RTM

mkamonster:
Interesting but what about the portnumber associated to the endpoint? Does it respect the X-Forwarded-Host completely? So the combination of hostname and portnumber?

PR #107, opened 2016-06-20
Fix null reference in XmlCommentsModelFilter

### LINKED ISSUE #100: Null reference in GetCommentIdForProperty()
First, wanted to say that this is a great tool, thanks for putting it out there.

Running 6.0.0-beta9 on RC2, I'm finding some null reference exceptions in GetCommentIdForProperty(). jsonProperty.PropertyInfo() appears to never to get set. Changing null check seems to get things running again.
https://github.com/domaindrivendev/Ahoy/blob/master/src/Swashbuckle.SwaggerGen/Annotations/XmlCommentsModelFilter.cs

        public void Apply(Schema model, ModelFilterContext context)
        {
            var commentId = XmlCommentsIdHelper.GetCommentIdForType(context.SystemType);
            var typeNode = _xmlNavigator.SelectSingleNode(string.Format(MemberXPath, commentId));

            if (typeNode != null)
            {
                var summaryNode = typeNode.SelectSingleNode(SummaryTag);
                if (summaryNode != null)
                    model.Description = summaryNode.ExtractContent();
            }

            foreach (var entry in model.Properties)
            {
                var jsonProperty = context.JsonObjectContract.Properties[entry.Key];
                //if (jsonProperty == null) continue;
                if (jsonProperty.PropertyInfo() == null) continue;

                ApplyPropertyComments(entry.Value, jsonProperty.PropertyInfo());
            }
        }

applypropertycomments

gewoonrik:
fixes #100

### DIFF
--- a/src/Swashbuckle.SwaggerGen/Annotations/XmlCommentsModelFilter.cs
+++ b/src/Swashbuckle.SwaggerGen/Annotations/XmlCommentsModelFilter.cs
@@ -34,7 +34,11 @@ namespace Swashbuckle.SwaggerGen.Annotations
                 var jsonProperty = context.JsonObjectContract.Properties[entry.Key];
                 if (jsonProperty == null) continue;
 
-                ApplyPropertyComments(entry.Value, jsonProperty.PropertyInfo());
+                var propertyInfo = context.JsonObjectContract.UnderlyingType.GetProperty(jsonProperty.UnderlyingName);
+                if (propertyInfo != null)
+                {
+                    ApplyPropertyComments(entry.Value, propertyInfo);
+                }
             }
         }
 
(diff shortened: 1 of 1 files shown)


src/Swashbuckle.SwaggerGen/Annotations/XmlCommentsModelFilter.cs
```diff
@@ -34,7 +34,11 @@ public void Apply(Schema model, ModelFilterContext context)
                 var jsonProperty = context.JsonObjectContract.Properties[entry.Key];
                 if (jsonProperty == null) continue;
 
-                ApplyPropertyComments(entry.Value, jsonProperty.PropertyInfo());
+                var propertyInfo = context.JsonObjectContract.UnderlyingType.GetProperty(jsonProperty.UnderlyingName);
```
domaindrivendev: At first glance, it seems the above line could be reduced to:

    var propertyInfo = jsonProperty.PropertyInfo();

Maybe I'm missing something and if so, perhaps a unit test would help make the intention here a  little clearer

gewoonrik:
`jsonProperty.PropertyInfo();` did not compile for me. There is no PropertyInfo method on the JsonProperty object. But maybe I'm missing something?

gewoonrik:
My bad, I first fixed this code in my own project and then copied it back. I didn't know that PropertyInfo was a extension method.

Issue #111, opened 2016-06-27
Update to 1.0


Tazer:
Hey! :) Doing the upgrade to .net core 1.0 soon ?

tillig:
First, thanks to @domaindrivendev for this great package. It's one of the core things for REST on .NET that's pretty indisposable at this point, at least for me.

While I recognize the challenges of keeping up with the .NET Core stuff (I'm having similar challenges with [Autofac](https://github.com/autofac/Autofac)), I do see that there are several issues and pull requests here that could help speed things along in getting this out for ASP.NET Core. As it stands, my team is consuming custom builds of this package off our own internal NuGet server.

@domaindrivendev If you're interested, I and my team would like to help. I'd prefer that to forking and having to own the whole thing ourselves. If I/we could accept PRs on your behalf or, at the very least, get this up and running on the RTM bits and published to NuGet in beta, that'd be awesome. If you're interested, drop me a line - my email is my username at `paraesthesia.com`. Until then, I'll be updating all of this to run on ASP.NET Core in my fork since it appears I'm running into some type load issues due to breaking interface changes in the RTM bits.

No pressure, and again, tons of thanks for all the great work so far.

npnelson:
I don't know if this helps, but here is one of the things that changed between RC2 and RTM that impacts this project. https://github.com/aspnet/Routing/pull/303.  I'm not sure how to fix it though, I haven't had the opportunity to work with TemplateMatcher.

I also want to say that this is an awesome package and has become indispensable for my work as well.

bradygaster:
+1000 to what @tillig said regarding the project and how awesome it is. i'd love to see the 1.0 release coming out soon. I'm the project's biggest fan - @domaindrivendev can back me up on this. :)

tillig:
For reference / anyone interested, [I have a branch in my fork building against .NET Core RTM](https://github.com/tillig/Ahoy/tree/rtm-project-update).

domaindrivendev:
@tillig - any chance you could submit a PR for this and I'll review and try get it pushed through by EOD tomorrow.

tillig:
Added PR #112 based on my working fork/branch. Let me know if there are questions or if there's more work to do.

IanElsinga:
Hi, AddSwaggerGen fails with Core RTM with the following exception:  

> 
[...]

PR #112, opened 2016-06-28
Update for .NET Core RTM

tillig:
This is for #111.

I had to comment out one of the unit tests about setting defaults (marked with a TODO) because it didn't validate.

I also sorted/tidied the project.json files for easier diff comparisons.

### DIFF
--- a/src/Swashbuckle.Swagger/project.json
+++ b/src/Swashbuckle.Swagger/project.json
@@ -1,9 +1,27 @@
 ﻿{
+  "version": "6.0.0-rc-001",
+  "description": "Core components for Swashbuckle - a Swagger toolchain for ASP.NET Core",
+  "packOptions": {
+    "licenseUrl": "https://github.com/domaindrivendev/Swashbuckle/LICENSE",
+    "owners": [
+      "Richard Morris"
+    ],
+    "projectUrl": "https://github.com/domaindrivendev/Swashbuckle",
+    "summary": "Provides a Swagger object model and middleware that can be used to expose Swagger descriptions for your API.",
+    "tags": [
+      "swagger",
+      "documentation",
+      "discovery",
+      "help",
+      "webapi",
+      "aspnet",
+      "aspnetcore"
+    ]
+  },
   "dependencies": {
     "Microsoft.AspNetCore.Mvc.Core": "1.0.0",
     "Microsoft.AspNetCore.Mvc.Formatters.Json": "1.0.0"
   },
-  "description": "Core components for Swashbuckle - a Swagger toolchain for AspNet Core",
   "frameworks": {
     "net451": {},
     "netstandard1.6": {
@@ -12,17 +30,5 @@
         "portable-net451+win8"
       ]
     }
-  },
-  "packOptions": {
-    "licenseUrl": "https://github.com/domaindrivendev/Swashbuckle/LICENSE",
-    "owners": [
-      "Richard Morris"
-    ],
-    "projectUrl": "https://github.com/domaindrivendev/Swashbuckle",
-    "summary": "Provides a Swagger object model and middleware that can be used to expose Swagger descriptions for your API",
-    "tags": [
-      "Swagger Documentation Discovery Help WebApi AspNet AspNetCore"
-    ]
-  },
-  "version": "6.0.0-rtm1"
+  }
 }
--- a/src/Swashbuckle.Swagger/project.json
+++ b/src/Swashbuckle.Swagger/project.json
@@ -1,27 +1,28 @@
 ﻿{
-  "version": "6.0.0-beta9",
-  "description": "Core components for Swashbuckle - a Swagger toolchain for AspNet Core",
-  "authors": [ "Richard Morris" ],
-
-  "packOptions": {
-    "tags": [ "Swagger Documentation Discovery Help WebApi AspNet AspNetCore" ],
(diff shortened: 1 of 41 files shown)


tillig:
Oh, one thing - I had to switch the unit tests to only test one framework rather than multiple frameworks. There's a bug in VS tooling and xunit such that VS won't let you run tests that target multiple frameworks. If you can figure that one out... Well, I spent a couple hours on it and have up. :)

src/Swashbuckle.Swagger/project.json
```diff
@@ -1,27 +1,28 @@
 ﻿{
```
domaindrivendev: I kinda liked the existing layout for the project.json files :( I'd prefer to have properties grouped by meaning as opposed to alphabetically. Any chance we could change these back?

src/Swashbuckle.Swagger/project.json
```diff
@@ -1,27 +1,28 @@
...
+    "owners": [
+      "Richard Morris"
+    ],
+    "projectUrl": "https://github.com/domaindrivendev/Swashbuckle",
+    "summary": "Provides a Swagger object model and middleware that can be used to expose Swagger descriptions for your API",
+    "tags": [
+      "Swagger Documentation Discovery Help WebApi AspNet AspNetCore"
+    ]
+  },
+  "version": "6.0.0-rtm1"
```
domaindrivendev: Version should be 6.0.0-beta10. Not ready to call this baby rtm or even close just yet

test/WebSites/Basic/Swagger/AddCartDefault.cs
```diff
@@ -17,9 +18,9 @@ public void Apply(Schema model, ModelFilterContext context)
         {
             _logger.LogInformation("Applying an awesome IModelFilter that leverages Dependency Injection");
 
-            model.Default = new
+            model.Default = new Cart
             {
-                Id = "myCartId"
+                Id = 123
```
domaindrivendev: Right now - anonymous types are better for setting examples because the Swagger-specific serializer may not reflect the API's serialization settings (camelCasing etc.). At least with the anonymous types, you can control the casing yourself.

test/WebSites/CustomizedUi/project.json
```diff
@@ -1,81 +1,52 @@
...
     "Microsoft.VisualStudio.Web.CodeGeneration.Tools": {
-      "version": "1.0.0-preview1-final",
-      "type": "build"
+      "type": "build",
+      "version": "1.0.0-preview2-final"
     },
-    "Swashbuckle.SwaggerGen": "6.0.0-beta9",
-    "Swashbuckle.SwaggerUi": "6.0.0-beta9"
+    "Swashbuckle.SwaggerGen": "6.0.0-rtm1",
+    "Swashbuckle.SwaggerUi": "6.0.0-rtm1"
```
domaindrivendev: 6.0.0-beta10

src/Swashbuckle.Swagger/project.json
tillig: I can definitely do that. I changed them because in a solution with a lot of projects that all have to change at the same time it's much harder to diff and compare what's missing when they're in that grouped format. Having them in a predictable format, while potentially less human-readable, fixes a lot of problems with respect to long-term maintainability.

But, sure, I'll change them back. It'll take a bit.

src/Swashbuckle.Swagger/project.json
tillig: Unfortunately it can't be `beta10`. The problem is that NuGet uses the suffix _alphabetically_ to determine what's new. So, alphabetically, `beta10` is _older_ than `beta9`. That's actually why they switched early on from releasing `-rc2` style suffixes to `-rc2-final` - because people using early drops couldn't _upgrade_ to an _older_ version (`-rc2` < `-rc2-12345` < `-rc2-final`). When I have CI hooked up, I usually use the build number (~3 digits) as the beta suffix so I don't run into this issue.

I arbitrarily picked `rtm` so when I published to our internal NuGet server it would be seen as newer. For now I'll switch to `-rc-001` so it'll work with NuGet but won't be seen as RTM ready.

tillig:
I reformatted the project.json files to hopefully be closer to the originals. I did notice from the ASP.NET Core team's projects that their tags are all lower case and they're not space-delimited anymore, so I updated that, too.

I went with `-rc-001` for the version to fix the `beta` vs `rtm` issue. NuGet should see the `rc` version as later than `beta` yet it still indicates not quite ready for prime time.

Finally, I added an `.editorconfig` to help keep your style settings in place. If you don't use the [EditorConfig](http://editorconfig.org/) add-ins it won't matter, but if you do it'll help. I work across a lot of projects that have different indentation requirements (tabs vs. spaces, four spaces vs. two spaces, etc.) and if you use the plugin it'll automatically adjust on a per-project basis for you.

domaindrivendev:
@tillig - thanks for this great contribution. Great catch on the version ordering. However, the current state of this project is best reflected through a "beta" version. There's still quite a few missing features and there's also one or two areas where I foresee the need for breaking changes. It's a little ugly, but I ended up going with the following - 6.0.0-beta901

domaindrivendev:
Nuget shipping shortly ...

tillig:
Sounds great! Glad I could help.

another-guy:
Guys, don't want to make noise in this PR, but be sure your efforts are appreciated!

PR #113, opened 2016-06-28
Migrate to .NET Core RTM

kant2002:
Just migration to .NET Core RTM

While you wait when this PR will be accepted, you could add NuGet feed 

    <add key="NetCoreMigration" value="https://www.myget.org/F/netcore-migration/api/v3/index.json" />
to your Nuget.config file i nthe root of your solution


sirkirby:
Seems that there is already a PR in progress that addresses this... #112

PR #114, opened 2016-06-29
Upgraded from RC2 to .NET Core 1.0

ericwj:
The version was bumped to `6.0.0-beta10` in all NuGet package projects.

There is one test failing, but I believe it is due to [swagger.io/validator](http://online.swagger.io/validator/debug) not being spec compliant.

The test is the following one, which is tested to return an empty JSON file, but instead returns an error complaining that `definitions/Cart/default` is not of type `string`...which it shouldn't. It should be a valid cart, which it is.

Albeit I don't quite understand why its `Id` property is set to a string value in `AddCartDefault.cs` although it is of type `integer`. That should have this test fail if the validator is truly spec compliant. Not that it matters for the test result today - it fails with the same error if I set it to `-1`.

```
[Theory]
[InlineData(typeof(Basic.Startup), "/swagger/v1/swagger.json")]
public async Task SwaggerRequest_ReturnsValidSwaggerJson(...)
```


domaindrivendev:
The original PR for this - #112 has already been merged in.

PR #115, opened 2016-06-29
converted to dotnetcore 1.0

mihevc:
Updated to core 1.0 and tooling preview 2.  I hope this is helpful.

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 {
   "projects": [ "src", "test", "test/WebSites" ],
   "sdk": {
-    "version": "1.0.0-preview1-002702"
+    "version": "1.0.0-preview2-003121"
   }
 }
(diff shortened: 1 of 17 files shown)


AlexCuse:
I was about to look into this myself.  Would be happy to help with testing or any remaining work.

AlexCuse:
@mihevc have you tried this at all?  We are finding a lot of breaking changes to json serialization in the move to 1.0, thinking about pushing a temporary package to nuget so we can use the UI to work through them.

mihevc:
The only issues I ran into with serialization that now they camel case json by default and they changed up the constructors a bit.  I was able to work around them.  The constructor one caused me more issues.

domaindrivendev:
The original PR for this - #112 has already been merged in.

AlexCuse:
Ah great, thank you!

PR #116, opened 2016-07-02
The Docs pointed to the wrong app.Use  method

SteveStrongBAH:
this works with .net core 1.0


domaindrivendev:
Thanks - good catch

PR #126, opened 2016-07-11
- add SwaggerResponseRemoveDefaultsOperationFilter

maord3:
Add the ability to remove all default operations as a filter instead of declaring SwaggerResponseRemoveDefaultsAttribute on each and every controller class.
The usage is useful when returning IActionResult from a controller, by default swashbuckle add NoContent as a relevant response which is not true in my case.
I found it really useful in my case, it makes the code much cleaner.


BernieWhite:
There doesn't appear to be a tests in the contribution.

PR #133, opened 2016-07-14
Strong name assemblies

domaindrivendev:
@rynowak - would be great to get your :eyes: on this.


rynowak:
Sorry, this got missed somewhere in my inbox - looking now

rynowak:
Usually you'd just use a single key for a whole group of assemblies - is there a reason why you ended up with three? Other than that, looks good!

PR #166, opened 2016-09-02
Support for including custom stylesheets and scripts into default UI

andyalm:
This adds back some support for including custom stylesheets and scripts to the default UI. This feature existed in previous versions of Swashbuckle and I still find it useful. I decided not to include embedded resource support, as that can be a concern of the individual applications to use the EmbeddedFileProvider if they need something like that.


andyalm:
I'm not sure why these diff's are showing whole file differences. I did this work on a Mac. I wonder if it's line ending differences.

andyalm:
Please hold off on merging this, I just realized that I did not implement the javascript portion here correctly and the scripts run too early. I will add a commit that ensures the custom scripts don't execute until after the swaggerUI has initialized.

andyalm:
I just updated the script injection so that it worked the same as in Swashbuckle 5 where the custom scripts aren't executed until the UI has initialized.

PR #177, opened 2016-09-15
Added support for DefaultValue data annontation attribute

spacedsweden:
I had the need to populate with some some interesting data in all the properties of my Models, thought it would be useful for others as well


domaindrivendev:
Thanks for this - I wasn't aware of that attribute.

spacedsweden:
awesome!

PR #202, opened 2016-10-21
fix the the missed 'host' properites.

MienDev:
The host is missed, while invoking `GetSwagger`  in the middleware. (the original parameter just replaced by `null` value.)
One needs the `host` property while making swagger.json working well with the third-party tools, such as `Postman`. 
> if 'host' property missed, the imported entry in postman would generate the url like `http://foo/bar`, while urls like `http://foo.com:8888/foo/bar` with host was the correct one indeed.


MienDev:
My fault,  for a month I hav''nt use it.
I should have know that it's configurable now.

PR #204, opened 2016-10-25
Set oauth2 SecurityScheme type in constructor

moander:


domaindrivendev:
If doing this for the OAuth2 scheme, I'd like to be consistent and also do it for the other ones - e.g. BasicAuth and ApiKey

PR #205, opened 2016-10-25
Added support for a custom index.html Stream.

Sumo-MBryant:
This enables the existing templating and configuration support to be
performed but to have a custom index.html for features not exposed in
the configuration.


CumpsD:
Looking forward to this :) 👍

CumpsD:
Looking forward to this :) 👍 

@domaindrivendev any plans to support something like this?

Issue #173, opened 2016-09-12
ProducesResponseTypeAttribute not as functional as SwaggerResponseAttribute


tillig:
With the removal of `SwaggerResponseAttribute` in favor of the base MVC `ProducesResponseTypeAttribute` some ability to annotate the API seems to have disappeared (please do correct me if I'm wrong).
- **All responses require a type.** The `ProducesResponseTypeAttribute` allows you to specify a response code but you also must _also_ specify a non-null `System.Type` that goes with the response. What sort of type comes back with a 404? (It looks like @domaindrivendev [has already filed an issue on this](https://github.com/aspnet/Mvc/issues/4863), but there isn't a corresponding discussion here, so it may be worth talking about from a workaround perspective.)
- **There is no per-response description.** The `ApiExplorer.ApiResponseType` object has a status code and type associated, but no plaintext description. The current generator code [maps the description to hardcoded strings](https://github.com/domaindrivendev/Ahoy/blob/d871059b1898b97c9ab9d31ecc40ab15d183d119/src/Swashbuckle.SwaggerGen/Generator/SwaggerGenerator.cs#L216) but there's no great way to get that info back in there.

My current workaround is basically to replicate the original `SwaggerResponseAttribute` and put my own operation filter in place to get the full, richer set of data back in there.

I'm curious if there is a plan longer-term to put more metadata markup ability back into Swashbuckle or if this will always be a custom extension. If you don't want it in the main Swashbuckle packages, perhaps a new package `Swashbuckle.Metadata` or something with the attributes put back?

vas6ili:
@tillig You can use a custom tag `<response code="200">description here</response>` in XML docs to add a description to each response type ([this example in one of the test websites](https://github.com/domaindrivendev/Ahoy/blob/master/test/WebSites/Basic/Controllers/ResponseTypeAnnotationsController.cs)). While there are

domaindrivendev:
You can also use typeof(void) for responses that don't contain a payload

tillig:
Using XML docs doesn't seem to work in ASP.NET Core if you have a modular application where some of your REST API controllers are in separate assemblies. The XML docs don't appear to get copied to the build output / package directory.

`typeof(void)` is a trick I didn't think of to handle the type. Still doesn't take care of the per-response description, but it's a step closer.

vas6ili:
@tillig That is strange. I have request/response models and controllers in separate assemblies and it seems to work just fine.

[...]

PR #236, opened 2016-11-22
Add SwaggerProducesResponseTypeAttribute and operation filter

tillig:
This is a fix for #173 - putting a response type attribute in place that allows a description to be added to the response type for the purposes of Swagger docs.

I chose to derive the `SwaggerProducesResponseTypeAttribute` from the base ASP.NET core `ProducesResponseTypeAttribute` so you could avoid redundant attributes. Which is to say...

```csharp
// This is redundant, so you don't have to do it.
[ProducesResponseType(typeof(void), 204)]
[SwaggerProducesResponseType(204, description: "No content is returned.")]
public IActionResult MyAction()
{
  // ...
}

// This is all you need - it will work with both standard ASP.NET API description
// as well as Swagger generation.
[SwaggerProducesResponseType(204, description: "No content is returned.")]
public IActionResult OtherAction()
{
  // ...
}
```

I did consider making it an operation filter attribute, but given it led to redundant markup I opted for this way. If that's going to be troublesome or cause confusion, it wouldn't be hard to switch it to an operation filter attribute instead.


src/Swashbuckle.SwaggerGen/Annotations/SwaggerProducesResponseTypeAttribute.cs
```diff
@@ -0,0 +1,50 @@
+﻿using System;
+using System.Linq;
+using Microsoft.AspNetCore.Mvc;
+
+namespace Swashbuckle.SwaggerGen.Annotations
+{
```
domaindrivendev: Purely a style preference, but I'm really not a big fan of XML comments, or comments in general for that matter. My philosophy is - strive for expressive code that reveals it's intention on it's own, reserving comments for hacks or non-intuitive workarounds that can't be expressed well with code alone. My bad for not having some contributor guidelines, but any chance you could remove the XML comments :)

src/Swashbuckle.SwaggerGen/Annotations/SwaggerProducesResponseTypeAttribute.cs
```diff
@@ -0,0 +1,50 @@
...
+    /// <para>
+    /// This attribute derives from <see cref="ProducesResponseTypeAttribute"/> so you don't
+    /// have to put duplicate/redundant attributes on a controller or action to serve both Swagger
+    /// and standard ASP.NET API description purposes - just use the extended
+    /// Swagger response attribute and it works for both.
+    /// </para>
+    /// </remarks>
+    /// <seealso cref="Microsoft.AspNetCore.Mvc.ProducesResponseTypeAttribute" />
+    [AttributeUsage(AttributeTargets.Class | AttributeTargets.Method, AllowMultiple = true, Inherited = true)]
+    public class SwaggerProducesResponseTypeAttribute : ProducesResponseTypeAttribute
```
domaindrivendev: I like the idea of extending ProducesResponseType, but would prefer to still call the attribute _SwaggerResponseAttribute_ as it reads a little better and is more consistent with the existing _SwaggerOperationAttribute_

src/Swashbuckle.SwaggerGen/Annotations/SwaggerProducesResponseTypeFilter.cs
```diff
@@ -0,0 +1,87 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.AspNetCore.Mvc.ApiExplorer;
+using Swashbuckle.Swagger.Model;
+using Swashbuckle.SwaggerGen.Generator;
+
+namespace Swashbuckle.SwaggerGen.Annotations
+{
```
domaindrivendev: Remove XML comments as per above, it's hard to read the code with all the comments in the way :)

src/Swashbuckle.SwaggerGen/Annotations/SwaggerProducesResponseTypeFilter.cs
```diff
@@ -0,0 +1,87 @@
...
+        private static void ApplyAttribute(Operation operation, OperationFilterContext context, SwaggerProducesResponseTypeAttribute attribute)
+        {
+            var key = attribute.StatusCode.ToString();
+            Response response;
+            if (!operation.Responses.TryGetValue(key, out response))
+            {
+                response = new Response();
+            }
+
+            response.Description = attribute.Description;
```
domaindrivendev: I'm wondering (don't have access to .NET dev environment right now to verify) if the response schema will already be there by virtue of the attribute extending _ProducesResponseType_. Prior to executing filters, the generator will have added responses based on the presence of ApiDescription.SupportedResponseTypes. It depends on the implementation of ApiExplorer, but it's possible that applying the _SwaggerProducesResponseTypeAttribute_ attributes, which extend _ProducesResponseType_, will cause SupportedResponseTypes to be populated. Wondering if you considered this?

domaindrivendev:
A few small comments - but other than that looks awesome!

src/Swashbuckle.SwaggerGen/Annotations/SwaggerProducesResponseTypeAttribute.cs
tillig: Before I remove them... XML comments are where IntelliSense sources its doc. While it adds to the code, consumers aren't reading the code at this level.

Are you sure you want them gone?

src/Swashbuckle.SwaggerGen/Annotations/SwaggerProducesResponseTypeFilter.cs
tillig: In unit tests for the filter the whole integration pipeline hasn't run so you get a NullReferenceException without the check. Given I didn't want the filter to assume core behavior lest something inadvertently break, I figured the null check was cheap and reasonable.

src/Swashbuckle.SwaggerGen/Annotations/SwaggerProducesResponseTypeAttribute.cs
domaindrivendev: Yep I'd still like to remove them - thanks.

tillig:
Updated per review comments.

PR #238, opened 2016-11-25
Updated SwaggerUI to 2.2.8

alexvaluyskiy:
Fixed https://github.com/domaindrivendev/Ahoy/issues/229

### DIFF
--- a/src/Swashbuckle.SwaggerUi/bower.json
+++ b/src/Swashbuckle.SwaggerUi/bower.json
@@ -2,7 +2,7 @@
     "name": "swashbuckle",
     "private": true,
     "dependencies": {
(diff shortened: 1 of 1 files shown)


domaindrivendev:
Thanks for this. Wondering if you compared the Swashbuckle version of index.html with the latest from swagger-ui? It has lots of SB specific stuff so will be different but sometimes there's changes  (e.g. new/updated JavaScript script tags) that need to be added to the SB version

alexvaluyskiy:
@domaindrivendev No, they have not changed index.html since 10 August

domaindrivendev:
Awesome - thx for the contribution

PR #240, opened 2016-12-03
Fixed exception when generating actions with model binding.

LammyL:
Fixed bug where swagger generator would throw an exception if the parameters were bound on modelBinding, services or custom before allowing the operation filters to remove the parameters.  I moved the validation check to happen after the filters get applied.

### DIFF
--- a/src/Swashbuckle.SwaggerGen/Generator/SwaggerGenerator.cs
+++ b/src/Swashbuckle.SwaggerGen/Generator/SwaggerGenerator.cs
@@ -159,7 +159,7 @@ namespace Swashbuckle.SwaggerGen.Generator
 
             //the operation filters may remove parameters that are not supported such as services injections or CancellationToken
             //we need to validate rule this AFTER we apply the filters
-            var allowedParameterSources = new[] { "fromData", "body", "header", "path", "query" };
+            var allowedParameterSources = new[] { "formData", "body", "header", "path", "query" };
             if (operation.Parameters != null && operation.Parameters.Any(c => !allowedParameterSources.Contains(c.In, StringComparer.OrdinalIgnoreCase)))
             {
                 throw new NotSupportedException(string.Format("Ambiguous location (path, query etc.) for one or more parameters in action - {0}. " +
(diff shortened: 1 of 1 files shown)


domaindrivendev:
@LammyL - thanks for this PR. You've definitely identified an issue that's been introduced with the raised exceptions on unknown binding types. However, I'm not 100% sold on the solution you've proposed ...

It's always been my intention that the generator will generate/enforce valid Swagger and then filters can be used to extend and enhance the generated document if necessary. Your solution now allows the generator to return invalid Swagger, giving filters an opportunity to correct it. Practically it makes sense, but philosophically it kinda breaks the intended separation of concerns.

Going back to the problem your trying to solve - I think parameters that are bound to services will be automatically ignored because parameters are filtered by _Source.IsFromRequest=true_. I also think, Swashbuckle should automatically detect CancellationToken parameters and ignore them. I would be happy to add this change.

So, that just leaves the third case you mentioned - parameters with "custom" model binders. These are described as bindings from "unknown data sources" and therefore shouldn't have the IsFromRequest flag because all of the request sources are known as per the HTTP specification. Thoughts?

LammyL:
Personally I was trying to get it working again with the cancellation token.  I have this implemented as a filter and some changes broke it, hence the pull request.

My use case is really just this one at this point.  I added the other types in this PR mostly because they were supported by the enum returned from the framework.

I'm happy to send you another pull request to address just the cancellation token if you can suggest a good place in the code to filter out the cancellation token.  It's a bit messy because the one cancellation token parameter is split into 4 separate fields by the framework, but it have it working cleanly as a filter so it would be easy to reimplement.

domaindrivendev:
Hey I just added handling of the CancellationToken here - bb4f070d4107f3378b2809361338c8dbabd57c29. You're correct - it is a little messy. Maybe my implementation is a little flaky would be interested to see what your filter looks like?

LammyL:
I like your implementation.  It is very clean and elegant.  My filter checked for pretty much exactly the same things you had.  

I'll close this pull request for now since it isn't really required anymore.

PR #243, opened 2016-12-06
bump up version of netcore framework to 1.1.0

gcristofol:
Bump up version of netcore framework to 1.1.0 across all projects

### DIFF
--- a/src/Swashbuckle.Swagger/project.json
+++ b/src/Swashbuckle.Swagger/project.json
@@ -1,4 +1,7 @@
 ﻿{
+  "runtimes": {
+    "win10-x64": {}
+  },
   "version": "6.0.0-*",
   "description": "Core components for Swashbuckle - a Swagger toolchain for ASP.NET Core",
 
@@ -31,6 +34,12 @@
 
   "frameworks": {
     "net451": {},
+    "netcoreapp1.1.0": {
+      "imports": [
+        "dotnet5.6",
+        "portable-net45+win8"
+      ]
+    },
     "netstandard1.6": {
       "imports": [
         "dnxcore50",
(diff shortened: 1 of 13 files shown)


LammyL:
i am wondering what the repercussions are with leaving this as netstandard1.6 rather than bumping up the library?  It would seem to me that changing it to the latest version would prevent older apps from running this library or linking to this library.

gcristofol:
Probably I could add netstandard on the top level project.json. The only incompatibiliy I did find was dotnet-test-xunit so (limited to the environment I have for testing) it should run just fine.

src/Swashbuckle.Swagger/project.json
```diff
@@ -31,6 +34,12 @@
 
   "frameworks": {
     "net451": {},
+    "netcoreapp1.1.0": {
```
alexvaluyskiy: It is not necessary to add netcoreapp to library projects

timotyweis:
Please update also the following references in `SwaggerUI`
- "Microsoft.AspNetCore.StaticFiles": "1.0.0" --> 1.1.0
- "Microsoft.Extensions.FileProviders.Embedded": "1.0.0" --> 1.1.0

timotyweis:
Please ensure that also `Microsoft.AspNetCore.StaticFiles` and `Microsoft.Extensions.FileProviders.Embedded` are moved to 1.1.0.

dewrox:
Is there an ETA for this to be resolved?  Without Core 1.1 integration being complete, uptake of this project will be hampered.

jarroda:
As has been mentioned before, it is not necessary for this library to be updated to use it in a 1.1 app.  We're using it in multiple apps today without any issues.

dewrox:
Has anyone used this (as Core 1.0) with Identity Server 4.0 (Core 1.1.0) and had the Authorize work?  Identity Server 4.0 group says it is a client issue... the client in this case is Swagger.

domaindrivendev:
Re-iterating @jarroda's comments. This project depends on Core 1.0 at a minimum but absolutely works if you upgrade to 1.1.

PR #246, opened 2016-12-14
Remove dependency on MVC in SwaggerUi

cwe1ss:
I'd like to use the SwaggerUi library outside of ASP.NET Core MVC. The library currently has a dependency on AspNetCore.Mvc.Core even though it doesn't need it.

This PR would remove this dependency and add back only the required dependencies.


domaindrivendev:
Sweet thx!

Issue #63, opened 2016-02-18
Produces attribute not appearing in swagger output


chrizy:
using 6.0.0-rc1-final

When a Produces attribute is added ether to the class or method,
 [Produces("application/json")]

the swagger.json "produces" element is always empty, 

"produces":[]

Thanks.

chrizy:
Confusing but Produces need to be the return data type,

senj:
I can set the return data type, but then my own return type won't get listed.
If I set [Produces("application/vnd.my.v1+json")] it doesn't work.

alexidsa:
I'm working on fixing it in this PR: https://github.com/domaindrivendev/Swashbuckle.AspNetCore/pull/292

RehanSaeed:
Can this issue please be reopened, it was never addressed.

domaindrivendev:
Looks to be working in the latest version of SB for ASP.NET Core - Swashbuckle.AspNetCore.1.1.0 ...

**Code**
```csharp
[Route("/products")]
[Produces("application/json", "application/vnd.my.v1+json")]
public class ProductsController
{
    ...
}
```

**Generated Swagger**
```json
produces: [
  "application/json",
  "application/vnd.my.v1+json",
]
```

ljfraney:
I upgraded to 1.1.0 and I have an action that returns an image, but I'm still getting the empty `"produces": []` array.
```
[Route("api/v1/[controller]")]
public class ProductController : Controller
{
    [HttpGet("Picture/{productId:int}", Name = "Picture")]
    [Produces("image/jpeg")]
    public async Task<IActionResult> Picture(int productId)
    {
        var picture = await GetPicture(productId);
        return File(picture, "image/jpeg");
    }
}
```
I get the following JavaScript error in swagger-ui.min.js: 

> Uncaught TypeError: Failed to execute 'createObjectURL' on 'URL': No function was found that matched the signature provided.

clawrenceks:
@domaindrivendev Something is definitely not quite right here. I can never get the produces attribute to populate the produces section. 

More importantly however, having written my own IOperationFilter to do the population of the produces array for me, I can now see entries in produces in the swagger.json, but my swagger UI never shows me the "Response Content Type" dropdown box.

What is controlling the appearance of this dropdown box? Is there any specific information I can provide to help with debugging? I am using asp.net core 2.0.5 and swashbuckle.aspnetcore 1.1.0

[...]

PR #292, opened 2017-02-01
Produces attribute is not working

siberianguy:
There's a long lasting reported issue about Produce attribute not working: https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/63

I'll work on fixing it in this PR. I've managed to create a failing test so far. I'm completely new to Swashbuckle codebase, so if anyone can hint me the direction where to go, I would be extremely grateful.


RehanSaeed:
I'm now hitting an issue caused by produces being empty (See https://github.com/domaindrivendev/Swashbuckle.AspNetCore/pull/292). Have you had any luck at finding a solution.

Thuran:
just replace FromBody with FromForm

Thuran:
Just replace FromBody with FromForm

PR #296, opened 2017-02-07
Allow the SwaggerUI Document title to be configurable

dennisroche:
By default, the Swagger UI will have a generic document title (e.g. 'Swagger UI'). When you have multiple Swagger pages open, it can be difficult to tell them apart.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# Set the default end-of-line to UNIX
+* text=auto eol=lf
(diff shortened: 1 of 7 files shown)


dennisroche:
- [x] Feature
- [x] Unit test added
- [x] Documentation updated

Revised from original approach of using JS to update the title to existing template inject `%(DocumentTitle)` as it was testable.

PR #306, opened 2017-02-20
Fix #305

### LINKED ISSUE #305: Swagger UI: JSConfig bootstrapping serialization error
Swagger UI's window.JSConfig bootstrapping is affected by JsonConvert's Default Settings, and this can cause runtime javascript errors. Here's my application setup:

At the top of `ConfigureServices`, I modify JsonConvert default settings for my application purposes:
```C#
// Serialization settings
var jsonSettings = new JsonSerializerSettings(); 
jsonSettings.ContractResolver = new CamelCasePropertyNamesContractResolver();
JsonConvert.DefaultSettings = () => jsonSettings;
```

As a result of `camelCase` settings above, the generated window.JSConfig line will become:
```javascript
window.JSConfig = JSON.parse('{"swaggerEndpoints":[{"url":"/swagger/v1/swagger.json"}')...
```

**Note the camelCased `swaggerEndpoints` and `url`, etc.**

As a result, **JSConfig.SwaggerEndpoints[0].Url** throws an exception and an empty window is shown.

JSConfig must be serialized reliably regardless of any JsonConvert default settings that the user might have chosen. 

image

ArashMotamedi:
In case user has overridden `JsonConvert.DefaultSettings` naming strategy (e.g. camelCase, etc.), it needs to be reset here to produce correct window.JSConfig serialized settings.

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerUI/Template/IndexSettings.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerUI/Template/IndexSettings.cs
@@ -11,6 +11,9 @@ namespace Swashbuckle.AspNetCore.SwaggerUI
         private static readonly JsonSerializerSettings jsConfigSerializationSettings = new JsonSerializerSettings
         {
             ContractResolver = new Newtonsoft.Json.Serialization.DefaultContractResolver()
+            {
+                NamingStrategy = new Newtonsoft.Json.Serialization.DefaultNamingStrategy()
+            }
         };
 
         public IList<StylesheetDescriptor> Stylesheets { get; private set; } = new List<StylesheetDescriptor>();


domaindrivendev:
Hey thx for this - great catch! Just wondering if, at this point, we should avoid JsonConvert altogether and use an isolated instance of JsonSerializer:

```csharp
private string GetJSConfigJson()
{
    var serializer = new JsonSerializer { NullValueHandling = NullValueHandling.Ignore };
    using (var writer = new StringWriter())
    {
        serializer.Serialize(writer, JSConfig);
        return writer.ToString();
    }
}
```

ArashMotamedi:
Yep, that's probably the safer approach that can avoid any other side effects from the static converter.

domaindrivendev:
Closing in favor of 8377c006378a7db46186a425f082b798d63c35e9

PR #307, opened 2017-02-21
Use app serialization for 'example' and 'default' fields

domaindrivendev:


src/Swashbuckle.AspNetCore.Swagger/Application/SwaggerContractResolver.cs
```diff
@@ -1,26 +1,50 @@
...
-        public override JsonContract ResolveContract(Type type)
+        protected override JsonProperty CreateProperty(MemberInfo member, MemberSerialization memberSerialization)
         {
-            var defaultContract = base.ResolveContract(type);
-            if (defaultContract is JsonDictionaryContract) return defaultContract;
+            var jsonProperty = base.CreateProperty(member, memberSerialization);
 
-            return _camelCasePropertyNamesContractResolver.ResolveContract(type);
+            if (member.Name == "Example" || member.Name == "Examples" || member.Name == "Default")
+                jsonProperty.Converter = _applicationTypeConverter;
```
domaindrivendev: Good enough for the response examples case?

PR #313, opened 2017-02-27
Update README.md

Procee:
Update documentation to reflect naming of the project components.
Add semicolon so the code sample is compilable.


domaindrivendev:
The published Nuget package was out-of-date as opposed to the docs. As of 1.0.0-rc2 (just release), _SwaggerUi_ was renamed to _SwaggerUI_ to be more consistent with .NET naming conventions.

PR #321, opened 2017-03-03
Optimized XmlCommentsOperationFilter logic

xperiandri:
Optimized `XmlCommentsOperationFilter.ApplyResponsesXmlToResponses` and `XmlCommentsOperationFilter.ApplyPropertiesXmlToPropertyParameters` search code.

Nothing changed in behavior

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,7 @@ Once you have an API that can describe itself in Swagger, you've opened the trea
 
     _Now you can restart your application and check out the auto-generated, interactive docs at "/swagger"._
 
-# Swashbuckle & ApiExplorer #
+# Swashbuckle, ApiExplorer, and Routing #
 
 Swashbuckle relies heavily on _ApiExplorer_, the API metadata layer that ships with ASP.NET Core. If you're using the _AddMvc_ helper to bootstrap the MVC stack, then _ApiExplorer_ will be automatically registered and SB will work without issue. However, if you're using _AddMvcCore_ for a more paired-down MVC stack, you'll need to explicitly add the Api Explorer service:
 
@@ -75,6 +75,28 @@ Swashbuckle relies heavily on _ApiExplorer_, the API metadata layer that ships w
 services.AddMvcCore()
     .AddApiExplorer();
 ```
+
+Additionally, if you are using _[conventional routing](https://docs.microsoft.com/en-us/aspnet/core/mvc/controllers/routing#conventional-routing)_ (as opposed to attribute routing), any controllers and the actions on those controllers that use conventional routing will not be represented in ApiExplorer, which means Swashbuckle won't be able to find those controllers and generate Swagger documents from them. For instance:
+
+```csharp
+app.UseMvc(routes =>
+{
+   // SwaggerGen won't find controllers that are routed via this technique.
+   routes.MapRoute("default", "{controller=Home}/{action=Index}/{id?}");
+});
+```
+
+You **must** use attribute routing for any controllers that you want represented in your Swagger document(s):
+
+```csharp
+[Route("example")]
+public class ExampleController : Controller
+{
+    [HttpGet("")]
+    public IActionResult DoStuff() { /**/ }
+}
+```
+Refer to the [routing documentation](https://docs.microsoft.com/en-us/aspnet/core/mvc/controllers/routing) for more information.
 
 # Components #
 


src/Swashbuckle.AspNetCore.SwaggerGen/Annotations/XmlCommentsOperationFilter.cs
```diff
@@ -98,7 +99,13 @@ private void ApplyResponsesXmlToResponses(IDictionary<string, Response> response
             {
                 // Check for a corresponding  API parameter (from ApiExplorer) that's property-bound?
                 var propertyParam = apiDescription.ParameterDescriptions
-                    .Where(p => p.ModelMetadata?.ContainerType != null && p.ModelMetadata?.PropertyName != null)
```
xperiandri: @Bjego, here double check is done that can be eliminated.
If you check for null than do something once `?.` is nice but if you apply `?.` to the same property more than once it's a waste of resources.
This short form does not add any value in readability too.

Bjego: Ok yes I think your right ;) I didn`t think about the resources.

src/Swashbuckle.AspNetCore.SwaggerGen/Annotations/XmlCommentsOperationFilter.cs
```diff
@@ -98,7 +99,13 @@ private void ApplyResponsesXmlToResponses(IDictionary<string, Response> response
...
+                    .Where(p =>
+                    {
+                        var modelMetadata = p.ModelMetadata;
+                        return modelMetadata != null
+                            && modelMetadata.ContainerType != null
+                            && modelMetadata.PropertyName != null;
+                    })
```
domaindrivendev: what's the advantage of this change vs just removing the second (redundant use of ?)? In other words, why not just replace:

    Where(p => p.ModelMetadata?.ContainerType != null && p.ModelMetadata?.PropertyName != null)

with

    Where(p => p.ModelMetadata?.ContainerType != null && p.ModelMetadata.PropertyName != null)

Is it just a style preference? If so, I still prefer the one-liner.

xperiandri: Nothing about style, only performance.

xperiandri: If you prefer the second one, that's fine too.
My option is more readable in my opinion.

src/Swashbuckle.AspNetCore.SwaggerGen/Annotations/XmlCommentsOperationFilter.cs
```diff
@@ -80,9 +80,10 @@ private void ApplyResponsesXmlToResponses(IDictionary<string, Response> response
             while (responseNodes.MoveNext())
             {
                 var code = responseNodes.Current.GetAttribute("code", "");
-                var response = responses.ContainsKey(code)
-                    ? responses[code]
-                    : responses[code] = new Response();
+
+                Response response;
+                if (!responses.TryGetValue(code, out response))
+                    responses[code] = response = new Response();
```
domaindrivendev: I like this change

src/Swashbuckle.AspNetCore.SwaggerGen/Annotations/XmlCommentsOperationFilter.cs
domaindrivendev: I don't like this change and would prefer to stick with the following:

    Where(p => p.ModelMetadata?.ContainerType != null && p.ModelMetadata.PropertyName != null)

domaindrivendev: So we're hitting subjective discussion about readability now. Like I said, I prefer the one-liner.

domaindrivendev: @xperiandri - do you want to update as discussed and then I'll be happy to merge. Thanks again for the contributions.

xperiandri: OK

PR #331, opened 2017-03-06
Added document name format

xperiandri:
Added ability to set document name format to allow usage like this 
```C#
options.DocNameFormat("Version {0}");

options.SwaggerDoc("Version 1", new Info
{
    Title = "Ha Ha Taxi API version 1",
    Version = "1",
});
```

Which is absolutely logical where name is set to value meaning name and version is set to value meaning version.

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/Generator/SwaggerGenerator.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/Generator/SwaggerGenerator.cs
@@ -25,20 +25,23 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
         }
 
         public SwaggerDocument GetSwagger(
-            string documentName,
+            string documentVersion,
             string host = null,
             string basePath = null,
             string[] schemes = null)
         {
             var schemaRegistry = _schemaRegistryFactory.Create();
 
+            var documentNameFormat = _settings.DocumentNameFormat;
+            string documentName = documentNameFormat == null ? documentVersion
+                                                      : string.Format(documentNameFormat, documentVersion);
             Info info;
             if (!_settings.SwaggerDocs.TryGetValue(documentName, out info))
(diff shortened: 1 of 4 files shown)


src/Swashbuckle.AspNetCore.SwaggerGen/Generator/SwaggerGenerator.cs
```diff
@@ -25,20 +25,23 @@ public class SwaggerGenerator : ISwaggerProvider
         }
 
         public SwaggerDocument GetSwagger(
-            string documentName,
+            string documentVersion,
```
domaindrivendev: documentVersion is too specific. Versioning is just one reason why you would define multiple documents but there may be others

domaindrivendev:
Sorry might be missing something but I really don't understand what this is solving. Can you

xperiandri:
What is the purpose of document name? If version is a unique identifier.

src/Swashbuckle.AspNetCore.SwaggerGen/Generator/SwaggerGenerator.cs
xperiandri: Like separate partitions?

erendrake: I have thought about using this for multi-tenancy where you want to do custom work that leads to a contract change

domaindrivendev:
documentName is the unique identifier. In many cases, you may want to use the version as you're document name but there's other cases where you wouldn't. For example, you may want different docs for your "Order Processing API" vs your "Product Catalog API". The tenancy scenario that @erendrake described is another good example.

domaindrivendev:
I'm closing this PR as I still don't understand what problem it's actually trying to solve

xperiandri:
Agree

PR #334, opened 2017-03-08
Fix SecurityRequirementsOperationFilter

RehanSaeed:
The `SecurityRequirementsOperationFilter` is incorrectly mapping a policy name to claim types, there is a comment about this but it would be much better to map to the actual claim types which we can get quite easily from the AuthorizationOptions which is populated with policies like so:

```
services.AddAuthorization(
    options =>
    {
        options.AddPolicy("ReadPolicy", x => x.RequireClaim("read-customer"));
    });
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -637,29 +637,39 @@ services.AddSwaggerGen(c =>
 // SecurityRequirementsOperationFilter.cs
 public class SecurityRequirementsOperationFilter : IOperationFilter
 {
+    private readonly IOptions<AuthorizationOptions> authorizationOptions;
+
+    public SecurityRequirementsOperationFilter(IOptions<AuthorizationOptions> authorizationOptions)
+    {
+        this.authorizationOptions = authorizationOptions;
+    }
+
     public void Apply(Operation operation, OperationFilterContext context)
     {
-        // Policy names map to scopes
-        var controllerScopes = context.ApiDescription.ControllerAttributes()
+        var controllerPolicies = context.ApiDescription.ControllerAttributes()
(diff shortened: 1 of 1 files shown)


domaindrivendev:
Nice! Also good to see the filter DI-awareness being leveraged :)

RehanSaeed:
Cool. I've done a bit more work on this and am wondering if you would be willing to take another PR:

- The `ActionAttributes` and `ControllerAttributes` use reflection. I've found a way of getting authorization requirements without without doing that.
- I've split the `SecurityRequirementsOperationFilter` into three `IOperationFilter`s:
    1. `UnauthorizedResponseOperationFilter` - Add Unauthorized response when `DenyAnonymousAuthorizationRequirement` authorization requirement` is found.
    2. `ForbiddenResponseOperationFilter` - Add Forbidden response when any of the other authorization requirements are found i.e. `ClaimsAuthorizationRequirement`, `NameAuthorizationRequirement`, `RolesAuthorizationRequirement` and `AssertionRequirement`.
    3. `OAuth2ClaimsOperationFilter` - Adds `oauth2` claims to `Operation.Security` from any `ClaimsAuthorizationRequirement`s found.
- I've written some basic tests for the above.

This is probably too much code to put in the ReadMe. Perhaps it should live within the project itself?

DavidStrickland0:
Not sure The "Filter Di Awareness" is working as intended.  
In line 657 
+            .Select(x => this.authorizationOptions.Value.GetPolicy(x))

markti:
I get the same error as @DavidStrickland0 ... not sure what's missing

Issue #335, opened 2017-03-09
Move to new project setup


ruant:
Since VS2017 is out now, is it time to move everything over to the new .csproj setup?

ruant:
csproj changes: https://github.com/ruant/Swashbuckle.AspNetCore/commit/5bcbcf859c189ba0656a76d3790d95589321a8f8

I also updated everything to asp.net core 1.1 and the dependencies to the latest version: https://github.com/ruant/Swashbuckle.AspNetCore/commit/9ac44438dc1b6258d40ab12d1ade398649ddd1b4

For me it builds and runs fine, so if you want me to create a pull request let me know.

The only thing I noticed was that one of the test projects complained about a constructor, but that must have been there from the start?

Also, I've never migrated a project like this before, so if there something that should be changed let me know 😃

domaindrivendev:
Have it done on a local branch and now trying to resolve some build warnings

xperiandri:
Cool!

ruant:
@domaindrivendev Awesome, let us know if you need help with anything 😃

timotyweis:
@domaindrivendev any ideas of a new deadline for the `v1.0.0` release?

Issue #143, opened 2016-07-30
Core, with net452 framework?


domaindrivendev:
Moving the original issue - https://github.com/domaindrivendev/Swashbuckle/issues/780 here as it's related to SB 6x and ASP.NET Core ...

I'm not seeing a way to install Swashbuckle on an Asp.Net Core application using the net452 framework.

Swashbuckle 6.0 will only install on dnx framework.

Swashbuckle 5.0 is set up for OWIN

Is it possible to get Swashbuckle on a Core/net452 application

domaindrivendev:
@rynowak - any ideas? Is it as simple as adding net452 here https://github.com/domaindrivendev/Ahoy/blob/master/src/Swashbuckle.Swagger/project.json#L32

JSkimming:
We're installing into an  ASP.NET Core application targeting .NET 4.6.1 and it works just fine.

Given you're already targeting net451 it should work from from .NET 4.5.1 through to .NET 4.6.1.

rynowak:
@domaindrivendev yes - it's that simple. We go all the way down to 4.5.1 https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.Core/project.json#L59

You may need some ifdefs though - `#if NET451` and `#if NETSTANDARD1_6` (and similar) are provided for you.

mkontula:
I was trying this today, and with current rc3 package, the test project fails with following information. The test project and api application project both target `net452`. When I change the targets to `netcoreapp1.1`, the exception goes away. To me it seems something is trying reference 1.0.0.0 version of  `Microsoft.Extensions.DependencyInjection.Abstractions`, while the version being used by `Microsoft.AspNetCore.Mvc.Core` & others is 1.1.1.0 or similar.

```
=== Pre-bind state information ===
LOG: DisplayName = Microsoft.Extensions.DependencyInjection.Abstractions, Version=1.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60
 (Fully-specified)
LOG: Appbase = file:///C:/work/Tests452/bin/Debug/net452
LOG: Initial PrivatePath = NULL
Calling assembly : Swashbuckle.AspNetCore.SwaggerGen, Version=1.0.0.0, Culture=neutral, PublicKeyToken=d84d99fb0135530a.
===
LOG: This bind starts in default load context.
[...]

PR #339, opened 2017-03-13
Convert to msbuild

domaindrivendev:


domaindrivendev:
Addresses #335

domaindrivendev:
Closing in favor of 2c3e02334efe3d27c8118ff2afd7708443eee192

mkontula:
In 2c3e023 core library references point to 1.0.* versions. Ruant's comment and commit in issue #335 addresses new packages. RC4 targeting current package versions would be greatly appreciated.

domaindrivendev:
@mkontula - perhaps I'm missing something with the new csproj semantics but it's my understanding that the following ref ...

    <PackageReference Include="Microsoft.AspNetCore.Mvc.Core" Version="1.0.0" />

specifies a dependency supporting any version of Core >= 1.0.0. In fact the docs [here](https://github.com/NuGet/Home/wiki/PackageReference-Specification) say exactly that:

_This means the same as 3.6.0 in project.json today, which is >=3.6.0, preferring the lowest_

As per the [Nuget best practices for package developers](https://docs.microsoft.com/en-us/nuget/create-packages/dependency-versions), which states ...

_Developers most commonly specify a minimum version, which is a version number without adornmants as shown above, like 6.1.0. This allows NuGet to install that version or later_

I want to depend on a minimum version so that consumers can upgrade at their own pace.

At least that's my current understanding of things and what I consider to be best practice. @rynowak - if you have any input here it would be much appreciated (as always)?

mkontula:
In a comment in #143 I found problems with rc3 when targeting net452. 

In my test project there's a reference to Microsoft.AspNetCore.Hosting.Abstractions. VS2017 shows, that version 1.1.1 is referenced. My API project references rc3 of this library. When the testhost builds starts the API project, this library is unhappy when it founds 1.1.1.0 and not 1.0.0.0 which gets referenced because rc3 references 1.0.0 of Microsoft.AspNetCore.Mvc.Core. 

I tried to bump versions of this libary up like @ruantt did as well, and the test project compiles & runs tests with no problem. Hopefully somebody with more understanding of this could help. Curiously enough, when I target netcoreapp1.1 with the API & Test project, no problems exists at all.

ruant:
@mkontula I would love if someone could clarify how this should work as well.
What I did was a more try and fail approach 😆 until it worked.

I do agree with you, @domaindrivendev , it should depend on the minimum version it works with.

mkontula:
I kind of repeated your work today @ruant and figured out these things: 
- running against full .net framework (in my case net452) assembly binding is different from netcoreapp1.1
- with net452 any reference which points to NET Core 1.0.* fails if the running environment (in my case xunit test project targeting net452) has references to 1.1.* packages.
- since SwashBuckle.AspNetCore.SwaggerGen references Microsoft.AspNetCore.Mvc.Core 1.0.0 it also wants the .NET Core packages with assembly version 1.0.0.0 when it's run against full framework. In my case the .NET Core packages were all from 1.1.* era and xunit was very unhappy like the snippet shows in #143.

This is more than weird and I'll raise a question in aspnetcore repo. If it won't be fixed, it should certainly be more clearly documented. As a workaround I bumped all versions up to current and stashed the binaries to our own myget feed.  Not pretty but it works.

PR #357, opened 2017-03-27
Added filter for applying data annotations to parameters.

rthunbo:


src/Swashbuckle.AspNetCore.SwaggerGen/Annotations/DataAnnotationAttributesOperationFilter.cs
```diff
@@ -0,0 +1,113 @@
...
+            }
+
+            for (var index = 0; index < operation.Parameters.Count; index++)
+            {
+                var parameter = operation.Parameters[index];
+                if (parameter is NonBodyParameter nonBodyParam)
+                {
+                    var propertyParam = context.ApiDescription.ParameterDescriptions
+                       .Where(p => p.ModelMetadata?.ContainerType != null && p.ModelMetadata?.PropertyName != null)
+                        .FirstOrDefault(p => parameter.Name.Equals(p.Name, StringComparison.OrdinalIgnoreCase));
```
Erikvl87: @rthunbo, forgive me if I'm mistaken, but I believe you can't safely assume that the ApiParameterDescription from the OperationFilterContext is the same as the one from the Operation by only comparing the parameter name.

I understand it is your only hook to do so at the moment, but a Header parameter with the same name as a FromQuery parameter will result in both having (or not having, because of the order returned by FirstOrDefault) the Required property.

I'm guessing that the only way to be sure is to change the IParameter returned by [CreateParameter()](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/0e45a32d2f33c01877332f6da1516eb54a042c0a/src/Swashbuckle.AspNetCore.SwaggerGen/Generator/SwaggerGenerator.cs#L163) to contain a reference to the [paramDescription](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/0e45a32d2f33c01877332f6da1516eb54a042c0a/src/Swashbuckle.AspNetCore.SwaggerGen/Generator/SwaggerGenerator.cs#L165) it was created with. That way, you can retrieve the appropriate metadata directly from the parameter from within the operation, instead of having to lookup the parameter within the OperationFilterContext.

rthunbo: @Erikvl87 well this is the way it is done in XmlCommentsOperationFilter already. See https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/1292e2b9f410bb6ef29c457d685241d2730330c1/src/Swashbuckle.AspNetCore.SwaggerGen/Annotations/XmlCommentsOperationFilter.cs

rthunbo: @domaindrivendev whats your oppinion about this PR. Is it something that you would want in the swashbuckle library.

freeranger:
@rthunbo Great work.   Just wondering why you ignore body parameters?
Specifically for the required attribute - in a POST the body is required and it seems odd that route params will appear as required in the UI but body params do not.
One could add a filter to always mark body as required...but perhaps that isn’t necessarily the case, so using a [Required] param would work here....if only there was a way to communicate that to the UI...which would be via a filter like this!

Cheers

src/Swashbuckle.AspNetCore.SwaggerGen/Annotations/DataAnnotationAttributesOperationFilter.cs
domaindrivendev: @rthunbo - apologies for the late response and thanks for this contribution. You're definitely addressing a gap in the current functionality but I would tend to agree with @Erikvl87 in that it feels like core functionality as opposed to something that lives in a filter.

It's also bothering that this logic has to be essentially repeated (it or something very similar is also in SchemaExtensions.AssignValidationProperties. It all feels like a design smell (not yours but mine).

So, let me dig a little deeper before deciding on next steps. Thanks

domaindrivendev:
Closing as I believe it's covered by #749

PR #405, opened 2017-05-31
Document EmailAddressAttribute with schema.Format = "email"

maxnorth:
Adds swagger documentation for the EmailAddressAttribute data annotation by setting the schema format equal to "email".


domaindrivendev:
As far as I can tell, "email" is not one of the standard format values identified by the Swagger spec. If applications want to expose their own custom formats, then they should use a Schema filter (see readme)

Issue #408, opened 2017-06-01
Support NETStandard.Library 1.6.0


strofimo:
Hi,

Is it possible to lower NETStandard.Library version dependency to 1.6.0?
I'm using Asp.Net Core on AWS and they provide support only for LTS version of Core which is 1.0.
This prevents to use any packages which required  NETStandard.Library 1.6.1 and higher.
So far I'm using Swashbuckle.6.0.0-beta* but it luck of some important configuration capabilities. 

thanks

tillig:
Looks like it might be a transitive dependency since the packages target `netstandard1.6` but don't specify an exact version of `NETStandard.Library`.

strofimo:
Based on info here - https://www.nuget.org/packages/Swashbuckle.AspNetCore/1.0.0-rc3, rc3 doesn't depend on NETStandard.Library at all like Swashbuckle.6.0.0-beta* did. Same for Swagger, SwaggerUI and SwaggerGen components. Why do we have such dependency in release version?

tillig:
Most likely a difference in tooling. Between RC3 (February 27, 2017) and release (April 21, 2017) VS 2017 was officially released and I think there were some .NET Core tooling updates.

Given there's going to be a dependency - transitive or direct - on this (since the MVC bits depend on it, too), it might be time better spent figuring out how you can fix it. Can you add a direct dependency on 1.6.0 to fix it or does the transitive dependency via the MVC stuff still get you? Do you have to back-rev the MVC dependencies to fix it? If so, is that a good idea given the security updates they've been releasing?

In the end, you may be able to submit a PR with a fix that'd work for you... which is a much faster way of getting what you are looking for, I think, than just waiting for it to happen.

strofimo:
I find out that if doesn't specified explicitly, newest version of NETStandard.Library that VS or the CLI knows about will be used. So all 4 projects use 1.6.1 now. U can see in nuget managment for solution. Another thing is that NETStandard.Library package is no longer meant to be upgradable through the package management UI. The only way I find out is adding "<NetStandardImplicitPackageVersion>1.6.0</NetStandardImplicitPackageVersion>" into "PropertyGroup" part of csproj file for each project. After these changes everything builds ok and most of the tests pass (4 test fails but it seems for unrelated to changes reason).

ryanolsonx:
Is there any update on this? I'm looking for this same change.

domaindrivendev:
[...]

PR #409, opened 2017-06-01
Use netstandard.library 1.6.0

strofimo:
Use NETStandard.Library 1.6.0 instead of 1.6.1.
[Issue details](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/408)


wallymathieu:
Well, wouldn't it be better to get rid of NetStandardImplicitPackageVersion ?

PR #413, opened 2017-06-07
Use Additional API Explorer Metadata

### LINKED ISSUE #412: Swagger Generator Does Not Use All API Explorer Metadata
## Overview
The implementation of [SwaggerGenerator.cs](../blob/master/src/Swashbuckle.AspNetCore.SwaggerGen/Generator/SwaggerGenerator.cs) should consider all relevant metadata provided by API explorers. The following information is not currently mapped in all **NonBodyParameter** scenarios:

* **Required** - this can be determined by location **and** by **ApiParameterDescription.RouteInfo.IsOptional**
* **Description** - this can be determined by **ApiParameterDescription.ModelMetadata.Description**, when available
* **Default** - this can be determined by **ApiParameterDescription.RouteInfo.DefaultValue**

## Scenarios
There are undoubtedly numerous scenarios where this woudl be useful. In the case of [API Versioning](https://github.com/Microsoft/aspnet-api-versioning), this will enable discovered API version parameters to be injected into Swagger documents and the UI via its API explorer extensions without requiring additional configuration from service authors. Service authors must currently use custom **IOperationFilter** implementations to bridge this gap.

commonsensesoftware:
Honors additional API explorer metadata, when available.  The information is mapped from:

* ApiParameterRouteInfo.IsOptional
* ApiParameterRouteInfo.DefaultValue
* ModelMetadata.Description

This will negate the requirement to use IOperationFilter when API explorers provide this information.  Fixes #412


domaindrivendev:
@commonsensesoftware - could you take a look at the failing tests when you get a chance?

commonsensesoftware:
Hey @domaindrivendev, I remember seeing these failures before I submitted the PR. I'm not sure if you looked at the build details, but the failures are unrelated. They are caused by **XmlCommentsTextHelper**. I assumed you were already aware of these failures. I switched back to my local _master_ branch and confirmed the tests still fail without my changes.

I've seen this type of issue for the offending tests before. It is the result of different line endings used in the test input strings. Getting the setup right can be maddening. The test input will look ugly, but I recommend swapping out literal line breaks in your strings with **Environment.NewLine**.

I _can_ fix these failing tests, but they aren't really related; however, there's no benefit to be pedantic here. If you'd like me to fix the tests as part of this PR, I'm happy to do so. If I've somehow completely missed the issue, feel free to straighten me. Let me know what you'd like to do. Thanks.

cfeilen:
I think this should be able to merge now that the tests have been fixed.

garrmark:
@domaindrivendev do you have the ability to re-run the checks on this pr or will it need to be goosed to get the checks to pass?

commonsensesoftware:
That would have been _easier_, but I just went ahead and rebased my `master` and this branch with the latest. The push on my side re-triggered the CI build in AppVeyor. Looks like the build is fine now.

dchw:
I am really eager to see this in, @domaindrivendev; any sense on if/when this could be merged?

commonsensesoftware:
I'm not sure I follow. This code is simply saying,

_"The API Explorer provided a default value for parameter [insert name] and I should use that value as the default value for the Swagger parameter."_

If an API Explorer provides a default value (e.g. non-null), why would Swagger care how that value was discovered and why would it not use it? It's also important that the default value is applied **after** the schema model is applied because non-nullable types (ex: int) will pave over the default value if there is one. This behavior is not specific to API versioning. Consider `api/orders/by-date/{date}` which has a default value provided. Developers shouldn't have to further customize the Swagger document and users shouldn't have to fill in a value with a known default value.

As it relates to API versioning, it allows for many flavors of API versioning out-of-the-box:

* By query string
* By URL segment
* By header
* By media type

Even when the URL segment method is used, it must be described as a route parameter such as `api/v{api-version}/resource`. There is no _magic_ string parsing. In all cases, you'd want the value of the `{api-version}` parameter to be automatically populated with the default value provided by the API Explorer. Today, this can only be achieved by creating a custom **IOperationFilter** and filling it in yourself. If Swashbuckle honored this information, it would supplant the need for this particular customization.

There have been some asks here and in API versioning to support substituting the API version value in the route template. I've just added support for that feature, but it's still a choice and you have to opt into it. Furthermore, that doesn't solve other scenarios such as providing the default value when versioning by query string parameter.

commonsensesoftware:
Apologies if I was implying that the semantics between the API Explorer and Swagger are the same; however, Swashbuckle does use a considerable amount of data from the API Explorer to support Swagger. Default values are seemingly innocuous information provided by the API Explorer that Swashbuckle _could_ be picking up. If there is some situation where it should not be picked or honored, I don't think there is an issue with that. I didn't realize there was any situation where a formal Swagger parameter is generated (from the API Explorer information) where that parameter **cannot** have a default value.

While I think there potentially other interesting cases, let's flush how/why this is relevant to Swagger/Swashbuckle and API Versioning together.

### Scenario 1
**Relative Path:** `api/v{api-version}/resource`
**Parameter:** api-version
**Location:** path
**Default Value:** 1.0
**UI-Generated Request URL:** `api/v1/resource`

**Current Behavior**
Swashbuckle generates a Swagger parameter called _api-version_ that has no default value (e.g. blank). The user must fill it in the UI. This is error prone and most service authors do not like it.

**Desired Behavior**
Swashbuckle generates a Swagger parameter called _api-version_ that sets the Swagger parameter _default_  to the default value from the API description provided by an API Explorer, if there is one. This parameter _should_ be read-only if possible.

### Scenario 2
**Relative Path:** `api/resource`
**Parameter:** api-version
**Location:** query
**Default Value:** 1.0
**UI-Generated Request URL:** `api/resource?api-version=1.0`

**Current Behavior**
Swashbuckle generates a Swagger parameter called _api-version_ that has no default value (e.g. blank). The user must fill it in the UI. This is error prone and most service authors do not like it. 

**Desired Behavior**
Swashbuckle generates a Swagger parameter called _api-version_ that sets the Swagger parameter _default_  to the default value from the API description provided by an API Explorer, if there is one. This parameter _should_ be read-only if possible.

### Conclusion
Regardless of whether a parameter is required, I don't see why a Swagger parameter cannot have a default value. In reviewing what the Swagger documentation says, there is some arguably confusing language. In the case of a required path parameter, the default value provided by the API Explorer should be treated as a _constant parameter_. Certainly in the case of a path parameter, I think most people want this value to be _read-only_. Ideally, this would be the case for the query string parameter too, but I don't think there are any current semantics to do that without hooking into the Swagger document generation (which is reasonable).

Swagger seems to have some limitations based on its dogma with respect to path parameters. Plenty of web stacks allow optional path parameters as long as they are at the end of the path. We could all debate whether that's a _good_ practice or not, but I've seen it many times (though I don't personally support it). For example, `api/orders/{id?}`, is a route with a path parameter that is optional because it's at the end. I'm not sure if that even a supported scenario in Swashbuckle, but the Swagger documentation definitely says that it cannot be optional.

Perhaps a more _complete_ implementation should use the following rules:
* If a parameter is required and a default value is provided, it is modeled as a _constant parameter_
* If a parameter is optional and a default value is provided, it is modeled as a _default parameter value_

If I'm not mistaken, this would make the implementation something like:

```c#
if (defaultValue != null)
{
    if (parameter.Required)
        parameter.Enum = new object[]{ defaultValue };
    else
        parameter.Default = defaultValue;
}
```

Thoughts?

domaindrivendev:
I really do appreciate your comments and I apologize for being so pedantic around what looks to be a fairly trivial PR. But, we're still not 100% on the same page. In an effort to gain alignment, I'm going to list out several statements that I believe to be true. Please read through them and let me know which, if any, you disagree with:

1. This PR attempts to populate parameter default from `RouteInfo`. But, ApiExplorer only populates `RouteInfo` on a parameter description if the parameter is a "route" parameter. Therefore, this change will have no effect on query or header parameters.

[If my understanding is incorrect, please provide a code example that I can use to verify]

2. Swagger 2.0 does not support optional "route" parameters, as is clearly stated in [the spec](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#parameter-object):
_If the parameter is in "path", this property is required and its value MUST be true_

3. In Swagger 2.0, the `default` field is used to declare the value the server will use if no value is provided by the client, as is clearly stated in [the spec](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#parameter-object):
_Declares the value of the parameter that the server will use if none is provided_

4. Therefore, the `default` field has no meaning for required parameters because by definition they MUST be provided by the client, as is clearly stated in [the spec](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#parameter-object):
_Note: "default" has no meaning for required parameters_

5. Assuming statements 1-4 to be true, then the proposed change will only ever assign a default value to "route" parameters, which MUST be required, and therefore the assigned `default` value will have "no meaning"

commonsensesoftware:
We're close.

I fully appreciate your comments and commitment to the thought process here. You know the Swagger spec _way_ better than me, so I won't even remotely pretend to be an expert in what it says.

First, a primer. In ASP.NET Core, the only way I've found that any type of _default value_ can be expressed in the API Explorer API is via `RouteInfo.DefaultValue`. I agree that it's kind of a misnomer, but I don't have control of the ASP.NET team's API design on this topic. Arguably, `DefaultValue` should have been put directly on **ApiParameterDescription**.

It might be useful for you to take a peek at [ApiVersionParameterDescriptionContext.cs](https://github.com/Microsoft/aspnet-api-versioning/blob/master/src/Microsoft.AspNetCore.Mvc.Versioning.ApiExplorer/ApiVersionParameterDescriptionContext.cs#L226) in API Versioning to get a sense of how the information is built. From a design perspective, here's how the relevant information is used and intended to be mapped:

* `ApiParameterDescription.Source` - indicates _where_ a parameter is coming from (ex: query, path, header, etc)
* `ApiParameterDescription.RouteInfo.IsOptional` - indicates whether the parameter is required. The API versioning API Explorer will, in fact, mark this true for a path parameter.
* `ApiParameterDescription.RouteInfo.DefaultValue` - indicates the default value for the parameter. The parameter built by the versioned API Explorer will always be the value of the corresponding API version.

I completely agree with your assessment of the Swagger spec as it relates to _default parameter values_. Terminology appears to be key here. In the Swagger [Describing Parameters](https://swagger.io/docs/specification/describing-parameters) page, the mapping we're looking for when paths are involved is **Constant Parameters**, which states:

>You can define a constant parameter as a required parameter with only one possible value.
>...
>**Note:** A constant parameter is not the same as the default parameter value. A constant parameter is always sent by the client, whereas the default value is something that the server uses if the parameter is not sent by the client.

The way _constant parameters_ are mapped in the example is via the `enum` attribute with a single value.

If my understanding is correct, then the rules should be something like:
* If `ApiParameterDescription.Source == BindingSource.Path`, then
  * `ApiParameterDescription.RouteInfo.IsOptional` is irrelevant because the parameter is required
  * If `ApiParameterDescription.RouteInfo.DefaultValue` has a non-null value, then it can be mapped as a _constant parameter_ in the `enum` attribute as a single element array
* Otherwise, `ApiParameterDescription.RouteInfo.IsOptional` determines whether the parameter is required
  * Regardless of whether it's required, if `ApiParameterDescription.RouteInfo.DefaultValue` has a non-null value, then it can be used in the `default` attribute

It appears you're determining whether the parameter is required by using `ModelMetadata.IsRequired`. That's probably _more_ correct than using `ApiParameterDescription.RouteInfo.IsOptional` and something I can update/enhance on my side. The documentation for these APIs ... how do I say it tactfully? ... is sorely lacking.

I think that means that my proposal would be revised once more to:

```c#
if (defaultValue != null)
{
    if (location == "path")
        // this is a 'constant parameter' where the value is a single, required, enum value
        parameter.Enum = new object[]{ defaultValue };
    else
        // this is an optional or required parameter with a 'default parameter value'
        parameter.Default = defaultValue;
}
```

I haven't been able to find anything in the **ModelMetadata** that allows for a default value. If you know of an alternate or _correct_ way to do that, I'm happy to adjust things on my side because it sounds like I might be mapping it incorrectly in that case.

I think there are a couple small tweaks before the PR is taken in its current state, but I want to make sure we're aligned before making any changes just yet. Are we a few steps closer to making an accord?

domaindrivendev:
Please provide inline responses to my 5 statements. That way, we can stay on point. If we’re agreed on all of them, I see no reason for this PR

commonsensesoftware:
1. **False**. The API Explorer, at least by API Versioning, **always** popluates `RouteInfo`. The `RouteInfo.DefualtValue` property is the only place that I know of that a default value can be specified. This is true for path, query, and header parameters.

2. Agreed. If the value comes from the "path", then it's required.

3. Agreed. In Swagger 2.0 and 3.0, the `default` field simply provides an initial value. This would certainly apply to a query string parameter.

4. Agreed that `default` has no meaning for required "path" parameters. However, I disagree that the value has to be provided by the client (e.g. typed in). While the [OpenAPI spec](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#parameter-object) articulates that `default` cannot be be used for required "path" parameters, it makes no mention of _constant parameters_. Both the Swagger [3.0](https://swagger.io/docs/specification/describing-parameters/) and [2.0](https://swagger.io/docs/specification/2-0/describing-parameters/) documentation describe using `enum` for a required "path" parameter. My interpretation of this omission in the spec is simply that the spec does not need to call it out. The definition of `enum` requires that it have at least one element and it will be provided by the server. Using the default value provided by the API Explorer as a single-valued `enum` for a required "path" parameter is not a violation of the spec nor would it fail validation. It's not clear to me if the UI has a different spec, but the wording in the spec with respect to required parameters and default values is _flawed_ in my opinion. There's no reason why a required parameter, of any type, can't have a default value. Despite my reservations about the language, I'm not suggesting to go off-spec; however, by spec, any required parameter, in particular "path" parameters, can use `enum` with a single value. Regardless of whether you think that should only be for "path" parameters or all required parameters, I think that would still satisfy consumers.

5. \# 2 and \# 3 are **true**; however, \# 1 and \# 4 are **false**. The default value provided by the API Explorer will either be mapped to `default` or `enum` depending on whether the parameter is a "path" parameter. Alternatively, all required parameters can be mapped to `enum` if a default value is provided.

domaindrivendev:
OK, now I understand what you’re trying to do. We’ll have to agree to disagree on spec interpretation. For me, default means the value that the server will use if the API consumer doesn’t provide one, and therefore makes no sense for required fields. I think you want to provide “possible” values (even if there’s just 1) and that’s not a “default” value by my interpretation of the spec.

I think the use of enum makes more sense so I’d prefer that solution. But, I’m not happy piggybacking on an ApiExplorer field that’s for a different purpose. It’s a hidden coupling to ApiVersioning, well highlighted by the time it took for me to understand the true purpose of the code in this PR. I’d like to investigate a more appropriate way to represent this information in ApiExplorer

memark:
Any update on this?

commonsensesoftware:
Yes, but there is more work to do. @domaindrivendev and I have chatted. We're on the same page. Unfortunately, I've confirmed with the ASP.NET team that there is an inherent flaw in current design of the new API explorer in ASP.NET Core. This is being tracked as [#7563](https://github.com/aspnet/Mvc/issues/7563) in the ASP.NET Core MVC repo.

To enable the _correct_ design without making concessions in Swashbuckle to the current limitations, it will require the following:

1. Updating the API Explorer design
2. The ASP.NET team needs to accept the PR and publish a new package version
3. API Versioning needs to consume the package, populate the new data members, and publish a new package version
4. Swashbuckle needs to consume the package, consume the new data members, and publish a new package version

Once all of those pieces are in place, we'll be able to re-evaluate and close out this issue. Although a relatively small change, there is a lot of coordination between several repos - unfortunately, but that's the state of the universe.

mrtristan:
@commonsensesoftware how's the state of the universe looking? seems like maybe core 2.2.0 will have the prereq?

domaindrivendev:

ronwarner:
### ### ****

bradygaster:
I'm doing some investigations into how we can improve the overall HTTP API experience further. API Versioning is something i'm looking at, and one of the [samples](https://github.com/microsoft/aspnet-api-versioning/blob/master/samples/aspnetcore/SwaggerSample/SwaggerDefaultValues.cs#L33) specific to API Versioning & Swashbuckle references this issue. Also noticed that @pranavkm PR'ed and closed https://github.com/aspnet/Mvc/issues/7563. is https://github.com/aspnet/Mvc/issues/7563 finished up and if so, I was wondering if @domaindrivendev and @commonsensesoftware have what they need to finish this up and/or close the issue? 

No pressure, but if there's more work to do I'd like to chat with you both to figure out how/if the team can support. Thanks!

commonsensesoftware:
@bradygaster I believe all the pieces are there now. There was a gap where the default value and required did not have the correct meaning/intention (as stated in the original overview). The list of things should be:

- `OpenApiParameter.Description` _should_ map to `ApiDescription.ModelMetadata.Description` when unset
   -  Most people will use XML Comments for descriptions, but this leaves out other edge cases. The API version parameter, for example, is dynamically generated and has no XML Comment to bind. Without this, there is no description in the Open API document.
   - A description _may_ provided by the API Explorer. If it is, it should be used as a fallback - regardless (IMO).
- `OpenApiParameter.Required` maps to `ApiDescription.IsRequired`
  - IMO (and my examples), the result should be a logical _or_ of either as either side can indicate it's required
- `OpenApiParameter.Schema.Default` maps to `ApiDescription.DefaultValue` when unset

Quite a while back, it also came to my attention that there is gap in _deprecation_. API Versioning, Open API, and, hence Swashbuckle, all support indicating that an API is deprecated, but the API Explorer provides no such metadata. It's not a showstopper, but that would be the remaining barrier in having to write an `IOperationFilter` to bridge Swashbuckle and API Versioning. The possible solutions are:

1. Get the ASP.NET team to add the necessary metadata to the API Explorer
    a. This can probably be done, but there will be likely significant lead time before we get a revision that contains it
2. Provide a callback hook (or something) on `OperationFilterContext` to evaluate if something is deprecated
    a. This would probably be something you'd configure in the _options_
    b. API versioning already provides the `ApiDescription.IsDeprecated()` extension method, but there's no way to connect it

I'd ❤️ to see this finally flush out. Unfortunately, I haven't had the extra time to circle back around and knock it out myself.

HunteRoi:
Hey guys, any update on this by any chance? Also, wouldn't it be interesting to look into reopening the PR as the problem hasn't been solved yet?

PR #420, opened 2017-06-13
Add support for XmlSerializer attributes when generating swagger file

pierslawson:
Added support for setting the xml information based on XmlSerializer attributes. These work fine with any primitive based properties (but as with some existing features) are not currently supported for object based properties (due to an issue with OAI spec https://github.com/OAI/OpenAPI-Specification/issues/556)

However, this makes the sample XML a lot closer to the actually required XML. One feature not yet sorted is setting of the namespace of the root object.


domaindrivendev:

pierslawson:
I'm thinking of re-doing the work in the near future... since you have
moved the code base on a fair bit it might be worth starting afresh anyway.

PR #423, opened 2017-06-14
Don't throw  when the ActionDescriptor is not a ControllerActionDescriptor

masterpoi:
This allows for other implementations of ActionDescriptor. (Which we have internally)


domaindrivendev:
Can you address the test failures then I'd be happy to merge?

masterpoi:
These tests fail on master as well, seems like PR #296 broke them: https://ci.appveyor.com/project/domaindrivendev/ahoy/build/1.0.0-build175. Not sure why though.

masterpoi:
Triggered a new build since it was fixed.

PR #432, opened 2017-06-22
Fixes unit tests requiring \r\n to pass

cfeilen:
Looks like the change to .gitattributes in #296 ended up breaking these unit tests.
This reverts that change, however we should probably do something else to guarantee that this doesn't happen again.

### DIFF
--- a/test/Swashbuckle.AspNetCore.SwaggerGen.Test/Annotations/XmlCommentsTextHelperTests.cs
+++ b/test/Swashbuckle.AspNetCore.SwaggerGen.Test/Annotations/XmlCommentsTextHelperTests.cs
@@ -15,7 +15,7 @@ namespace Swashbuckle.AspNetCore.SwaggerGen.Test
 
             var output = XmlCommentsTextHelper.Humanize(input);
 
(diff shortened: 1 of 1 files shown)


cfeilen:
Hmmmm build failed.  Need to look into how to really "clean" the repo, so that all new files will be checked out (not using the cached ones which are all LF)
I don't know AppVeyor very well.  @domaindrivendev, is that something you know how to do?

cfeilen:
Looking at the AppVeyor output:
   git config --global core.autocrlf true
   git clone -q https://github.com/domaindrivendev/Swashbuckle.AspNetCore.git C:\projects\ahoy
   git fetch -q origin +refs/pull/432/merge:
   git checkout -qf FETCH_HEAD
I'm wondering if the initial checkout is doing the real work, and the fetch/checkout of the merge doesn't really update the files.

I've tested this locally, and with my own build on AppVeyor, and it seems to have been fixed. 
https://ci.appveyor.com/project/cfeilen/swashbuckle-aspnetcore/build/1.0.0-build3

PR #453, opened 2017-07-27
Upgrade to Swagger UI 3.0.21

carlmon:
- `BooleanValues` and `JsonEditor` configuration options are no longer supported.
- All other SB configuration options ported to 3.0.21 index.html template.
- No new configuration options introduced.

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerUI/Template/index.html
+++ b/src/Swashbuckle.AspNetCore.SwaggerUI/Template/index.html
@@ -1,144 +1,129 @@
-﻿<!DOCTYPE html>
-<html>
+<!-- HTML for static distribution bundle build -->
+<!DOCTYPE html>
+<html lang="en">
 <head>
   <meta charset="UTF-8">
-  <meta http-equiv="x-ua-compatible" content="IE=edge">
   <title>%(DocumentTitle)</title>
-  <link rel="icon" type="image/png" href="images/favicon-32x32.png" sizes="32x32" />
-  <link rel="icon" type="image/png" href="images/favicon-16x16.png" sizes="16x16" />
-  <link href='css/typography.css' media='screen' rel='stylesheet' type='text/css'/>
-  <link href='css/reset.css' media='screen' rel='stylesheet' type='text/css'/>
-  <link href='css/screen.css' media='screen' rel='stylesheet' type='text/css'/>
-  <link href='css/reset.css' media='print' rel='stylesheet' type='text/css'/>
-  <link href='css/print.css' media='print' rel='stylesheet' type='text/css'/>
+  <link href="https://fonts.googleapis.com/css?family=Open+Sans:400,700|Source+Code+Pro:300,600|Titillium+Web:400,600,700" rel="stylesheet">
+  <link rel="stylesheet" type="text/css" href="./swagger-ui.css" >
   %(StylesheetsHtml)
+  <link rel="icon" type="image/png" href="./favicon-32x32.png" sizes="32x32" />
+  <link rel="icon" type="image/png" href="./favicon-16x16.png" sizes="16x16" />
+  <style>
+    html
+    {
+        box-sizing: border-box;
+        overflow: -moz-scrollbars-vertical;
+        overflow-y: scroll;
+    }
+    *,
+    *:before,
+    *:after
+    {
+        box-sizing: inherit;
+    }
 
-  <script src='lib/object-assign-pollyfill.js' type='text/javascript'></script>
-  <script src='lib/jquery-1.8.0.min.js' type='text/javascript'></script>
-  <script src='lib/jquery.slideto.min.js' type='text/javascript'></script>
-  <script src='lib/jquery.wiggle.min.js' type='text/javascript'></script>
-  <script src='lib/jquery.ba-bbq.min.js' type='text/javascript'></script>
-  <script src='lib/handlebars-4.0.5.js' type='text/javascript'></script>
-  <script src='lib/lodash.min.js' type='text/javascript'></script>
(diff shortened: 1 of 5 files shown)


carlmon:
The failing tests are not due to this PR

bjornharrtell:
Nice work @carlmon. I would think https://github.com/domaindrivendev/Swashbuckle/blob/master/Swashbuckle.Core/Application/SwaggerUiConfig.cs needed some updated too but perhaps not?

carlmon:
Thanks @bjornharrtell, I will update it. Do you prefer [Obsolete] on the unsupported options? Or should I remove the options? I will also update the usage documentation.

bjornharrtell:
Not my call but I would prefer `[Obsolete]` to (hopefully) make this a non-breaking upgrade.

src/Swashbuckle.AspNetCore.SwaggerUI/Template/index.html
```diff
@@ -1,144 +1,129 @@
...
-<div id='header'>
-  <div class="swagger-ui-wrap">
-    <a id="logo" href="http://swagger.io"><img class="logo__img" alt="swagger" height="30" width="30" src="images/logo_small.png" /><span class="logo__title">swagger</span></a>
-    <form id='api_selector'>
-      <div class='input'>
-        <input placeholder="http://example.com/api" id="input_baseUrl" name="baseUrl" type="text" />
-        <div id='auth_container'></div>
-        <select id="select_document"></select>
-        <script type="text/javascript">
-          // Append configured endpoints as options in the document selector
```
dennismaxjung: You removed the "select document" - dropdown. 
Know you can add multiple swagger endpoint but you cant switch between them on the UI.

src/Swashbuckle.AspNetCore.SwaggerUI/Template/index.html
```diff
@@ -1,144 +1,129 @@
...
-          }
-
-          _.each(JSConfig.OnCompleteScripts, function (script) {
-            $.getScript(script);
-          });
-        },
-        onFailure: function(data) {
-          log("Unable to Load SwaggerUI");
-
-          _.each(JSConfig.OnFailureScripts, function (script) {
```
dennismaxjung: Throwing "Uncaught ReferenceError: _ is not defined"

src/Swashbuckle.AspNetCore.SwaggerUI/Template/index.html
```diff
@@ -1,144 +1,129 @@
...
+
+  function log() {
+    if ('console' in window) {
+      console.log.apply(console, arguments);
+    }
+  }
+  
+  // Build a system
+  const ui = SwaggerUIBundle({
+    url: JSConfig.SwaggerEndpoints[0].Url,
```
dennismaxjung: Change url to urls, (see: https://github.com/swagger-api/swagger-ui/blob/master/README.md#parameters) and change the SwaggerEndpoint object to an array of API definition objects {url: "<url>", name: "<name>"}

carlmon: Nice catch, thanks Dennis!

I updated the PR to support both cases of single and multiple endpoints.

src/Swashbuckle.AspNetCore.SwaggerUI/Template/index.html
carlmon: Thanks @dennismaxjung! I fixed this and updated the PR

dennismaxjung:
It's working, but there seems to be some error at the build task. Some NuGet versions conflicts

heldersepu:
Has anyone tested OAuth2 with this change?

cfeilen:
@dennismaxjung, I was able to fix the nuget errors by upgrading the Microsoft.NETCore.Platforms package from 1.0.2 to 1.1.0,
Fix is in test\Swashbuckle.AspNetCore.IntegrationTests\Swashbuckle.AspNetCore.IntegrationTests.csproj, and test\Swashbuckle.AspNetCore.SwaggerGen.Test\Swashbuckle.AspNetCore.SwaggerGen.Test.csproj.

Issue most likely stems from the fact that Appvyor has probably updated to .NETCore2.0, which now respects the Nuget errors (specifically NU1605)

dennismaxjung:
@cfeilen yes that seems to be the problem.
Maybe we could allow to disable error Check at ui by setting this option  not to the default value if it set to null

HideDev:
Hi, any news?

wallymathieu:
I've added a pull request to fix the build failure in the branch: https://github.com/GenasysTechnologies/Swashbuckle.AspNetCore/pull/1

domaindrivendev:
I def want to support 3x but am also concerned that it's going to be a significant change for folks who've become used to the 2x UI. So, for now, I think it might be better to make it opt-in. For this, I would suggest creating a new project in the repo called Swashbuckle.AspNetCore.SwaggerUI3. Then, opt-in would be supported by installing the additional package and then enabling the 3x UI as follows:

```csharp
app.UseSwaggerUI3(c =>
{
    ...
});
```

Also, I'd like to ensure maintain feature parity if possible - with this PR the MultipleVersions and OAuth2Integration examples don't appear to be working.

wallymathieu:
That sounds nicer

dennismaxjung:
@domaindrivendev
You're right. This might be the best solution for keeping the old v2 and providing the new v3.

Maybe the goal should be to share as much code as possible between both projects? 

Great would be to allow the user easy switching between them.

dfoulk:
Has SwaggerUI 3 been added yet? I noticed a pull request by @alexvaluyskiy, but it doesn't appear to be merged...

domaindrivendev:
See #543

PR #455, opened 2017-08-01
Fix unit tests failing due to line ending assumptions.

carlmon:


domaindrivendev:
Awesome - free time is scarce right now so thanks for resolving this

Issue #461, opened 2017-08-10
Missing HttpStatusCode versions of attribute constructors


izzycoding:
This is more of an annoyance than a bug. But can we please get HttpStatusCode versions of Attributes?
It would save me having to manually cast them to int everywhere.

VisualBean:
There.. created a pull request, lets see if it gets added :)

xperiandri:
You just have to use another type for this
https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNetCore.Http.Abstractions/StatusCodes.cs

thmhoag:
@xperiandri Makes sense, but is there any reason not to have an override allowing the `System.Net.HttpStatusCode` type in an overload? Seems like it might be nice to have either way.

@VisualBean Came across this and your PR (#465) after creating my own attribute for a similar overload. Not sure what happened with the CI build as it seems to have failed, but I'd also love to see this implemented to have the option of using the System.Net type (unless there is a compelling reason not to support it in addition to the `Microsoft.AspNetCore.Http.StatusCodes`)

VisualBean:
Yeah I dunno why this PR failed - cant figure out which "check" failed

domaindrivendev:
Microsoft.AspNetCore.Http.Abstractions/StatusCodes appears to be the "Core" way of dealing with status codes. So going to keep SB consistent with the Core and close issue

PR #465, opened 2017-08-15
Update responseattribute to have an overload that uses HttpStatusCode

VisualBean:
just created an overload that takes care of not having to cast statuscode to int every single time 👍 

see https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/461


VisualBean:
created new pr #586 hoping it would fix the weird check issue

Issue #468, opened 2017-08-21
Support .Net Standard 2.0


daniloro:
Hi, it seems interesting that Swashbuckle supports .Net Standard 2.0.

I have new apis that use the new .Net Standard 2.0 but I can not use Swashbuckle.

Regards

daniloro:
When I search by "Swashbuckle.AspNetCore" on Visual Studio 15.3.1 NuGet Package Explorer, Version combobox appears in blank.

I am using .Net Core 2 WebApi project on Visual Studio 15.3.1

![screen shot 2017-08-21 at 09 09 01](https://user-images.githubusercontent.com/2098832/29518616-6ab9c790-8650-11e7-83da-e0102b4d7026.png)

tillig:
Duplicate of #464

The 15.3.1 tooling has introduced a lot of problems in my projects. For example, Roslyn analyzers started throwing exceptions and failing to run. I'd probably blame the tooling for this, especially since Swashbuckle has no control over how VS displays things.

daniloro:
Following the .NET Standard DOC and NetStandard2 GitHub page .NET CORE 2 only support .Net Framework 4.6.1 as platform, and Swashbuckle.AspNetCore only support .Net Framework 4.5.1, this is why the VS not show an option to install Swashbuckle on .NetCore2 projects.

If the Swashbuckle would support .NetFramework 4.6.1, the Swashbuckle would be compatible with .NetCore1 and .NetCore2 projects following .NET Standard DOC's.
[https://docs.microsoft.com/en-us/dotnet/standard/net-standard](url)

This is my vision of the problem, feel free to correct me if I are wrong

![screen shot 2017-08-22 at 08 18 06](https://user-images.githubusercontent.com/2098832/29563256-1c29e448-8713-11e7-985a-144d5ef2881c.png)

![screen shot 2017-08-22 at 08 09 16](https://user-images.githubusercontent.com/2098832/29563266-26bf0de8-8713-11e7-835a-2cb2a7305fa1.png)

tillig:
Yeah, that's not really how it works.

Take `netstandard` out of the picture for a second and just think about .NET framework. If you wrote something in .NET 2.0 it should still work - without recompile - on .NET 4.0. If you wrote it on .NET 4.0, it should still work on .NET 4.5 or .NET 4.6. However, the other way isn't true - if you wrote it in .NET 4.5 it won't work against .NET 2.0.

[...]

Issue #464, opened 2017-08-15
Upgrade to .netstandard 2.0, asp.net core 2.0  


ghd258:
Upgrade to .netstandard 2.0, asp.net core 2.0

tillig:
Is there a reason to do this? By raising the minimum it would become incompatible with earlier versions, forcing all consumers to update all projects to also target netstandard 2.0.

domaindrivendev:
Agreed. For apps it's great to keep your dependencies up-to-date. But for libraries where you want to minimize the barrier for use, I would always advocate lower min versions. SB shouldn't dictate the version of asp.net core in people's applications.

JanEggers:
@domaindrivendev  it would be good if there is a 2.0 compliant version. right now i get a 500 when opening swagger ui because the static file nuget  referenced by swashbuckle still has version 1.0 and is not compatible with 2.0. (missing constructor) i was able to workaround it by explicitly adding staticfiles nuget in version 2.0 to my project. i would prefere a version of swashbuckle that has version 2.x and targets asp.net core 2.0 that way you could still provide bugfixes for 1.1.x but new features are on 2.x

JustinJohnWilliams:
Any news on this? I see it was re-opened. I'm also having to revert to static files to upgrade to .netcore 2.0

tillig:
@JustinJohnWilliams It's not reopened. There is an open pull request somewhat associated with the issue that has not been committed to or accepted. The current official solution to ASP.NET Core 2.0 support is to add your reference to Microsoft.AspNetCore.StaticFiles 2.0.0 and call it a day.

JustinJohnWilliams:
@tillig Thanks! I didn't see the project: https://github.com/domaindrivendev/Swashbuckle.AspNetCore at first. this appears to take care of everything during my upgrade. Awesome job to all of you. Fantastic work!

Issue #438, opened 2017-07-03
Upgrading to ASP.NET Core 2.0 Preview 2 breaks Swagger UI


cmaslen:
After upgrading to .NET Core 2.0 Preview 2 Navigating to the default /Swagger route results in the following exception:

    System.MissingMethodException: Method not found: 'Void Microsoft.Net.Http.Headers.EntityTagHeaderValue..ctor(System.String)'.
       at Microsoft.AspNetCore.StaticFiles.StaticFileContext.LookupFileInfo()
       at Microsoft.AspNetCore.StaticFiles.StaticFileMiddleware.Invoke(HttpContext context)
       at Microsoft.AspNetCore.StaticFiles.DefaultFilesMiddleware.Invoke(HttpContext context)
       at Swashbuckle.AspNetCore.Swagger.SwaggerMiddleware.<Invoke>d__6.MoveNext()

## Steps to reproduce

1. Upgrade .NET Core to 2.0 preview 2. Install latest VS to work with new API.
2. Create new Web API Project.

Change TargetFramework to netcoreap2.0. Manually edit the project file to update the packages as follows so the site runs:

    <PackageReference Include="Microsoft.ApplicationInsights.AspNetCore" Version="2.0.0" />
    <PackageReference Include="Microsoft.AspNetCore" Version="2.0.0-preview2-final" />
    <PackageReference Include="Microsoft.AspNetCore.Mvc" Version="2.0.0-preview2-final" />
    <PackageReference Include="Microsoft.Extensions.Logging.Debug" Version="2.0.0-preview2-final" />
    
Configure Swagger as per the documentation in this repo's README.md

3. Navigate to the /swagger route.

## Expected outcome

The Swagger UI for the default API in the web project

## Actual outcome

The Exception above

tillig:
The exception looks like it comes from the static files middleware rather than Swashbuckle. Have you tried specifically adding a reference to the StaticFiles package to ensure you have the latest pinned? There's all sorts of weird stuff that happens when you try to mix release with preview bits. Happened when they were first coming out with netcore also. Work in progress that has breaking changes they may or may not keep in the next release... Painful.

I'd look at all the packages Swashbuckle references and manually add them all to my project. Ensure they're the preview versions. That will force the latest to be used.

[...]

PR #473, opened 2017-08-24
Update to .netstandard 2.0

kimiko88:
Fix error using netcore 2.0


tillig:
Duplicate of #464.

See #464 for an explanation of why Swashbuckle intentionally has not upgraded to netstandard 2.0. This PR will likely not be accepted.

kimiko88:
Sorry, I didn't see #464. 

I upgrade to netstandard 2.0 for the same reason of @JanEggers in #464.

I understand your explanation but if someone want to use Swashbuckle in a netcore 2.0 project, need an upgrade like this.

tillig:
No, see #468. You can use it in netcore 2 projects without the upgrade. Just add the reference to the static files assembly. This is going to be a common sort of thing in netstandard / netcore - new to .NET people but something folks who use things like Node (lots of transitive dependencies) are used to.

kant2002:
I think this version should not replace support for .netstandard 1.6, but instead of add libraries required for .NET Standard 2.0 and use `#ifdef` to provide support for both versions. That way minimal version still would be the same, but at the same time support would be preserved.

tillig:
I'm not sure there's a lot of value in that. The problem is that this sort of thing pops up when almost any transient dependency gets an update or any new netstandard version gets released. If the "fix" for the problem is to add a direct reference to `Microsoft.AspNetCore.StaticFiles`, that seems easy enough. That's also the official Microsoft solution for any sort of security upgrades they issue - add direct references to the upgraded packages right to your project.

Targeting netstandard 1.6 means it does work with netstandard 2.0. The larger issue is the MVC dependencies, specifically the StaticFiles dependency, right? It's not different code that needs to execute in Swashbuckle.

Again, this is stuff folks who use systems with large numbers of small and transient dependencies are kind of used to. It's a new and unfamiliar pain to .NET folks, but it's the trade-off for having netstandard/netcore.

Anyway, it's up to @domaindrivendev if he wants to maintain that. I'm just saying, this is a solved problem and it doesn't make sense to add support burden at this level.

kant2002:
@tillig Sorry, I probably don't understand from discussion in both issues that all that should be done is just explicitly require `Microsoft.AspNetCore.StaticFiles` which is more then fine for me.

adamhathcock:
Is this being looked at?  Using on a ASP.NET Core 2 project pulls in 1.0 dependencies.

quetzalcoatl:
FYI: for asp-core 2.1 the fix of "just link static-files" does not seem to work anymore. See #438

MedAnd:
Same issue as others, all static files return **500 Internal Server Error** from Kestrel... any recommended workaround?

- Swashbuckle.AspNetCore Version 3.0.0 
- Microsoft.AspNetCore.Server.Kestrel Version 2.1.1
- Microsoft.AspNetCore.StaticFiles Version 2.1.1
- .Net Core 2.1.302 
- Ubuntu 16.04

domaindrivendev:
Upgrade min bound to netstandard2.0 was already merged in #806. Will be available with the next major release - 4.0.0, coming soon.

MedAnd:
Thanks for the follow-up @domaindrivendev .... any ETA?

PR #479, opened 2017-08-31
Removing hard coded reference to Microsoft.NETCore.Platforms

### LINKED ISSUE #478: Fix Nuget NU1605 errors
The test projects reference Microsoft.NETCore.Platforms 1.0.2, which is considered a downgrade, since NetStandard1.6.1 has a dependency on Microsoft.NETCore.Platforms.1.1.0

Removing the reference to Microsoft.NETCore.Platforms fixes this issue.

cfeilen:
Removes the reference to Microsoft.NETCore.Platforms from the test apps.

Fixes #478


domaindrivendev:
I believe this is already happened with organically with some other changes.

PR #500, opened 2017-09-28
Add XML comments of public data fields

PaulDotNet:
JsonProperty represents both properties and public data fields but Swashbuckle uses PropertyInfo to access type info and thus omits data fields.  This behavior creates situation where swagger spec contains public data fields but no comments for them.
Replacing PropertyInfo with MemberInfo fixes problem. Comments of data members are no longer ignored.

Changed unit tests to use XML comments file generated by compiler instead of checked in file. Some comments in the source code did not match ones in XML file. Now XML comment file is generated from the source code and all tests pass.


albator1932:
I'd like to see that happen in the public release, I also have the issue where XML comments for my model members having an empty description field, I worked around this with a custom filter but it forces me to duplicate the XML comments in an attribute which is not ideal and tedious.

loxsmoke:
This request is already 6 months old. It is easier just to build a custom filter with **ahem** something like https://github.com/loxsmoke/DocXml to read XML comments. Duplicating comments in attributes is not a good idea in my opinion.

albator1932:
It's not because it's 6 months old that it shouldn't be done eventually.
I know issues and pull requests on this project have been in slow mode for quite a while last year (probably because @domaindrivendev was busy with something else) and maybe this one slipped through the cracks somehow.
My point is, XML comments work as expected for actions but not for model members, if this is a bug it should be fixed and if not, then at least having some kind of explanation why would be nice.

I'm gonna take a look at your project though because I wondered how I could read the XML comments literally 2 hours ago so I'm glad you mentioned it :)

PR #509, opened 2017-10-09
fix #288 : leave response stream open on swagger json endpoint

### LINKED ISSUE #288: 500 on swagger json endpoint while logging response using middleware
When I tried logging the response with the method provided [here](http://www.sulhome.com/blog/10/log-asp-net-core-request-and-response-using-middleware), I am getting 500 on the swagger endpoint. While I am able to get the response on calling the URI. Same issue mentioned in the comment.

drauly:
Hello,

This should fix #288 by leaving the response stream open.

What do you think ?

### DIFF
--- a/src/Swashbuckle.AspNetCore.Swagger/Application/SwaggerMiddleware.cs
+++ b/src/Swashbuckle.AspNetCore.Swagger/Application/SwaggerMiddleware.cs
@@ -1,5 +1,5 @@
-﻿using System;
-using System.IO;
+﻿using System.IO;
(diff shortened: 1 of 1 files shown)


drauly:
Thanks @domaindrivendev !

Any planned release ?

foobert:
This commit has the side-effect that the generated JSON will now start with a byte-order-mark (BOM) due to the different StreamWriter constructor. (I just found this out the hard way and wanted to leave a comment if someone else is stumped by this)

Issue #74, opened 2016-03-30
Respect the LowercaseUrls Routing Setting


RehanSaeed:
When using lower case URL's in conjunction with the special `[controller]` string in the route like so:

```
services.ConfigureRouting(
    options =>
    {
        options.LowercaseUrls = true;
    });

[Route("bar/[controller]")]
public class FooController : Controller {}
```

The generated URL's that Swagger displays has an upper-case controller name:

> bar/Foo

irontoby:
Wasn't sure whether to comment here or open a new ticket... considering the (default for ASP.NET Core)[https://github.com/aspnet/Announcements/issues/194] is now camel case, shouldn't camel-case controller names now be the default?

I even explicitly have

            services.AddMvc()
                .AddJsonOptions(opt =>
                {
                    opt.SerializerSettings.ContractResolver =
                        new CamelCasePropertyNamesContractResolver();
                });

in my `Startup.cs` yet I'm getting PascalCased controller names in Swagger (though property names are indeed camelCase).

Is there currently a way to get camelCase controller names and still use the generalized `[Route("api/[controller]")]` on my controllers?

andre-artus:
What is the status on this? I'm still getting Pascal case.

jonsa:
I too wished that the document generator followed the routing configuration.

As a workaround I use the suggested [fix](https://gist.github.com/rafalkasa/01d5e3b265e5aa075678e0adfd54e23f) in https://github.com/domaindrivendev/Swashbuckle/issues/834 (it just needs some `using` tweaks)

dataspun:
I found the UI will output api/foo given:

```
[Route("api/[controller]")]
public class fooController : Controller
```

wickdninja:
I was running into the same issue using dotnetcore 2.1, and hacked this solution together.
 
I wouldn't doubt that there is a better way, but this worked for me.

```
// This method gets called by the runtime. Use this method to configure the HTTP request pipeline.
        public void Configure(IApplicationBuilder app, IHostingEnvironment env)
        {
            // Enable middleware to serve generated Swagger as a JSON endpoint.
            app.UseSwagger(c =>
            {
                c.PreSerializeFilters.Add((document, request) =>
                {
                    var paths = new Dictionary<string, PathItem>();
                    foreach (var item in document.Paths)
                    {
                        var key = item.Key.ToLowerInvariant();
                        var value = item.Value;
                        paths.Add(key, value);
                    }
[...]

PR #538, opened 2017-11-20
Take RouteOptions into account when generating path (names)

funkycoding:
Hi,

Implements / solves https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/74. You can opt-in for this behaviour by using `UseRouteOptionsForRouteNames()`

```
services.AddSwaggerGen(c =>
{
    ...
    c.UseRouteOptionsForRouteNames();
    ...
});
```
Currently it only looks at the `RouteOptions.LowercaseUrls`, but happy to extend if needed. I'm going to see if the MVC team is open for a PR on their ApiExplorer code, to have this behaviour by default (e.g. take RouteOptions into account when generating route namings)


creyke:
+1 for this functionality. Is this going to get merged?

domaindrivendev:
@funkycoding - any reason why we wouldn't always honor the route options? I try to avoid the introduction of additional configuration elements unless absolutely necessary.

domaindrivendev:

PR #552, opened 2017-12-08
Added ability to add parameter description attributes

jsheely:
Same feature I put a PR in for the old project. Though I didn't get a chance to test this one. Should work in theory!

https://github.com/domaindrivendev/Swashbuckle/pull/1202


jsheely:
So much for in theory. It doesn't actually work yet. Stand by

jsheely:
Going to repost this. Was a bit more work than expected for netstandard 1.6

PR #553, opened 2017-12-09
Fix parameters with a custom model binder that are passed in the query instead of the path

nmoATusercube:
When you have a custom type with a ModelBinder/JsonConverter used in a route, the ApiGenerator moved it to the query instead of the path.

I have also added a test.


domaindrivendev:

PR #554, opened 2017-12-11
Added ability to add parameter description attributes

jsheely:
Added SwaggerDescriptionAttribute
Added test for SwaggerDescription Attribute

Allows for descriptions on methods, parameters and model properties


domaindrivendev:
@jsheely - apologies for taking so long to respond. I'm typically quick to merge PR's that align closely to my overall vision for Swashbuckle (so long as the implementation looks sound). But, I'm not so responsive when there's some gaps that need to be addressed. This is mainly because the latter requires quite a bit of back and forth and my cycles are extremely limited these days :(

Anyway, you're definitely addressing a need with this PR and #555. But, there's some differences in how I want to address them. I've created a placeholder issue (#694) to capture my thoughts and am going to close your PR's in their current state. But, hopefully you'd be interested in contributing. Please read through the issue I created and let me know your thoughts? Thanks again for your contributions.

PR #558, opened 2017-12-13
Allow ApiDescriptionExtensions.ControllerAttributes() and .ActionAttributes() to return inherited attributes.

### LINKED ISSUE #557: ApiDescriptionExtensions.ControllerAttributes() and .ActionAttributes() don't return inherited attributes.
I'm trying to create my own OperationFilter, which reads a controller's and its action's Authorize attributes and documents these in the Swagger docs. My authorize attributes are however defined in a base class. This means my operationfilter cannot use the methods in the title to obtain the attributes easily.

nphmuller:
This fixes #557


domaindrivendev:
Happy to merge this but without the tests. IMO, they're overkill for a flag switch. Could you remove them from the PR?

nphmuller:
@domaindrivendev Done! Also rebased to the latest version.

Issue #358, opened 2017-03-27
Question: How can we take enums as references?


alexdrl:
Hello, we have an ASP.NET Core API from which we generate a swagger.json from this package (we love it).

One problem that we have is that we want to codegen the same enum not only as strings, but generate it using swagger.json refs. We modified the swagger.json manually and this is possible, we could generate code with NSwag and the enums are generalized.

I annotated the methods which return an enum without luck, how can we achieve that?

mkontula:
We have pretty much the same issue. Some controller methods have enums as their route or query parameters. This leads to multiple definition of enums in generated client code. I'd rather have one enum defined in definitions and referenced as parameter which swagger definition supports. Manually editing the generated swagger.json is a huge antipatterns because humans are involved.

gatimus:
It seems you can't have definitions for query parameters [SO: Swagger: Reusing an enum definition as query parameter](https://stackoverflow.com/a/32256688/6656422) but I have the same issue for body parameters. I made a [sample](https://github.com/gatimus/SwaggerEnums).

If you go to `swagger/v1/swagger.json` you get

```json
    "definitions": {
        "ModelA": {
            "type": "object",
            "properties": {
                "option": {
                    "format": "int32",
                    "enum": [
                        0,
                        1,
                        2
                    ],
                    "type": "integer"
                }
            }
        },
        "ModelB": {
            "type": "object",
            "properties": {
                "option": {
                    "format": "int32",
                    "enum": [
                        0,
                        1,
                        2
                    ],
                    "type": "integer"
                }
            }
        }
    },
```

The desired result would be

```json
    "definitions": {
        "ModelA": {
            "type": "object",
            "properties": {
                "option": {
                    "$ref": "#/definitions/MyEnum"
                }
            }
        },
        "ModelB": {
            "type": "object",
            "properties": {
                "option": {
                    "$ref": "#/definitions/MyEnum"
                }
            }
        },
        "MyEnum": {
            "format": "int32",
            "enum": [
                0,
                1,
                2
            ],
            "type": "integer"
        }
    },
```

I tired `c.MapType<MyEnum>(() => new Schema { Ref = "#/definitions/MyEnum" });` but only got

```json
    "definitions": {
        "ModelA": {
            "type": "object",
            "properties": {
                "option": {
                    "$ref": "#/definitions/MyEnum"
                }
            }
        },
        "ModelB": {
            "type": "object",
            "properties": {
                "option": {
[...]

PR #582, opened 2018-01-25
Generate schema / references for enum action parameters #358

onionhammer:
This basically changes the behavior of `DescribeAllEnumsAsStrings` for Enum types to generate a `Ref` for the parameter rather than generate the supported values inline.

Addresses #358

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/Generator/SwaggerGenerator.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/Generator/SwaggerGenerator.cs
@@ -192,7 +192,12 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
             if (schema == null)
                 nonBodyParam.Type = "string";
             else
+            {
                 nonBodyParam.PopulateFrom(schema);
+
+                if (schema.Ref != null)
+                    nonBodyParam.Ref = schema.Ref;
+            }
 
             if (nonBodyParam.Type == "array")
                 nonBodyParam.CollectionFormat = "multi";
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/Generator/SwaggerGenerator.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/Generator/SwaggerGenerator.cs
@@ -192,12 +192,7 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
             if (schema == null)
                 nonBodyParam.Type = "string";
             else
-            {
                 nonBodyParam.PopulateFrom(schema);
-
-                if (schema.Ref != null)
-                    nonBodyParam.Ref = schema.Ref;
-            }
 
             if (nonBodyParam.Type == "array")
                 nonBodyParam.CollectionFormat = "multi";
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/Generator/SchemaExtensions.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/Generator/SchemaExtensions.cs
@@ -87,6 +87,10 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
             partialSchema.UniqueItems = schema.UniqueItems;
             partialSchema.Enum = schema.Enum;
             partialSchema.MultipleOf = schema.MultipleOf;
+
+            var nonBodyParam = partialSchema as NonBodyParameter;
+            if (nonBodyParam != null)
+                nonBodyParam.Ref = schema.Ref;
         }
     }
 }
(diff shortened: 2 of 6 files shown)


src/Swashbuckle.AspNetCore.SwaggerGen/Generator/SwaggerGenerator.cs
```diff
@@ -192,8 +192,13 @@ private Operation CreateOperation(ApiDescription apiDescription, ISchemaRegistry
             if (schema == null)
                 nonBodyParam.Type = "string";
             else
+            {
                 nonBodyParam.PopulateFrom(schema);
 
+                if (schema.Ref != null)
+                    nonBodyParam.Ref = schema.Ref;
```
domaindrivendev: A very minor change, but could you move the following line into `SchemaExtension.PopulateFrom`:

```csharp
nonBodyParam.Ref = schema.Ref;
```

No need the for the `if` condition either

domaindrivendev:
I'm pretty much ready to merge this - just one minor change proposed.

domaindrivendev:
Also ... on second thoughts, I wonder if we should do this for all enums, regardless of the `DescribeAllEnumsAsStrings` setting. I'd prefer to keep the behavior consistent regardless of how client generators deal with it.

onionhammer:
@domaindrivendev I'll takey a look at changing that tomorrow

Regarding `DescribeAllEnumsAsStrings` I'm fine with that if you are

onionhammer:
Okay; I moved the code to assign the `Ref`, but it involves some unboxing and reboxing the partial schema since `PartialSchema` doesnt have the `Ref` property.

I can eliminate the `DescribeAllEnumsAsStrings` check as well if you like; but doesnt it have the potential to break people's code who upgrade swashbuckle and regenerate their code? I suppose it would force them to use a value thats closer to what they would want to see anyways though.

onionhammer:
@domaindrivendev Want me to update this PR to always generate the enum names? Its a small change, but it could have implications I dont know of right now - even when I get the tests passing again

onionhammer:
@domaindrivendev Okay.. just in case you want it this way I now generate enums as references regardless of the config

onionhammer:
@domaindrivendev I would actually like to make use of this some day, as it's the defacto-standard for aspnet core (even used in Microsoft's own docs), you requested assistance in the referenced issue:

> This should be a relatively straightforward change. Anyone on this thread interest in submitting a PR? This is FOSS afterall :)

And I came through with a PR, but I guess I'll have to write a swagger document by-hand or continue using WebApiProxy for now, as this is such a glaring limitation that would affect so much of my existing (generated) API.

domaindrivendev:
@onionhammer - I guess this isn't as straight-forward as I first thought. After seeing your message above, I made some time to merge the changes, only to discover that they cause the generated Swagger to be invalid.

In a nutshell, Swagger 2.0 does NOT support the `$ref` field on [Parameter Objects](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#parameterObject) and therefore, a shared enum definition is NOT possible for non-body parameters. Full details are discussed [here in the Swagger/OpenAPI repo](https://github.com/OAI/OpenAPI-Specification/issues/564).

So, we're snookered until SB supports v3 of the Swagger spec. This is a much heavier lift. In the meantime, I can merge a partial version of this PR that supports shared enum definitions within body parameters and response objects ONLY.

On a separate note, regarding your concern around backwards compatibility, I think we should put this behavior behind a specific flag that you need to opt into (e.g. `CreateReferencesForEnums`). This way, no-one is going to be surprised by the new behavior. With the next major release, where breaking changes are more acceptable, I might remove the flag and default to the new behavior.

I'll go ahead and make the adjustments. Please let me what you think - I'm assuming you're not going to be too happy about this, but an implementation that breaks the spec (even if the UI honors it) just isn't on the table.

onionhammer:
@domaindrivendev thanks for looking into this, it's a shame that this breaks the spec, but it's better you found out now than later

Looking forward to OpenApi/v3

domaindrivendev:
NOTE: I applied the changes I described above and merged. Because, I applied them to the merge commit, they're not going to show up in the `Files changed` tab for this PR. To see the final set of changes that got merged refer to the merge commit - https://github.com/domaindrivendev/Swashbuckle.AspNetCore/commit/5abcea4f59a9061870581d99dafbca3524e6b2fa.

In a nutshell, you can now opt into referenced definitions for enums that occur within body parameters OR response schemas with the following config option. Unfortunately, inline enum definitions will still be used for non-body parameters due to the Swagger 2.0 limitations described above.

yahorsi:
Hi Guys,

We recently had to move to the .NET Core 3.1 as 2.2. support ended and so, had to move to the latest Swashbuckle. Our models code-gen does not support Open Api 3.0 so, we have to use: UseSwagger(x=>x.SerializeAsV2 = true);

Unfortunately, in the generated swagger spec we see enums as references that is not supported by our code-gen as well.

Is there any way to turn old 4.0.1 behavior?

Thanks for the answers!

PR #583, opened 2018-01-25
Fix/Workaround for Produces Attribute

ljfraney:
Microsoft.AspNetCore.Mvc.ApiExplorer.ApiDescription.SupportedResponseMediaTypes() does not utilize the Produces attribute to determine the content types that a controller action can return. Since this method is being used to populate the produces array property in Swagger.json, when the Produces attribute is the only way to infer the response type, the array is empty. In the case of a Json response, SwaggerUI works fine with an empty produces array. But, in the case of an image or file mime type, SwaggerUI hangs, and a JavaScript error is thrown.

The code in this pull request works around this problem by populating the produces attribute with a disctinct list of content types from both the SupportedResponseMediaTypes() method, and the Produces attributes.

I should note that I have treated Produces attributes as hierarchical, and not stackable. I'm not sure that this is what Microsoft intended, but their documentation is not clear. Therefore, a [Produces("image/png", "image/jpeg")] attribute on an action will override a [Produces("application/json")] attribute on the action's controller. The alternative would require that the action be placed in a different controller if you don't intend for it to report that it supports "application/json".


domaindrivendev:
@ljfraney - this looks like a great addition. Any chance, you could free up a cycle to address latest conflicts?

domaindrivendev:
Thx - can you take a look at the failing CI build?

ljfraney:
I'm started by rebasing onto upstream master. I'm a little confused at why it now shows that I'm trying to merge 26 commits, instead of just the 2 commits that are from me. In my shop, we usually just merge from master onto our branch and squash. Let me know if I need to do it differently.

domaindrivendev:
Not sure why all those other commits are showing up here. If you rebase your branch on top of upstream master, it should be good. Once you've done that, I think you'll need to do a force push (`git push -f`) to clean up your remote branch, which is what's being to drive the PR. But, I will put out the disclaimer that I'm not a git expert :)

domaindrivendev:

PR #587, opened 2018-02-04
Update responseattribute to have an overload that uses HttpStatusCode

VisualBean:
just created an overload that takes care of not having to cast statuscode to int every single time 👍 see #461


VisualBean:
I dont understand why that check keeps failing

domaindrivendev:
This code doesn't compile ...

Annotations\SwaggerResponseAttribute.cs(14,41): error CS0246: The type or namespace name 'HttpStatusCode' could not be found

domaindrivendev:
Microsoft.AspNetCore.Http.Abstractions/StatusCodes appears to be the "Core" way of dealing with status codes. So going to keep SB consistent with the Core and close this PR

PR #618, opened 2018-02-27
Force JsonConvert not to use formatting

TDabasinskas:
If the application has default `JsonConvert` options configured as following, the Swagger-UI completely fails to load:

```csharp
JsonConvert.DefaultSettings = () => new JsonSerializerSettings
{
    Formatting = Formatting.Indented
};
```

This happens because the generated configuration object looks like this:

```javascript
  window.onload = function () {
    var configObject = JSON.parse('{
  "urls": [
    {
      "url": "/swagger/v1/swagger.json",
      "name": "v1"
    }
  ],
  "validatorUrl": null
}');
```

That is, the generated configuration contains new lines, which breaks the JavaScript with `SyntaxError: unterminated string literal` error.

Passing `Formatting.None` to `JsonConvert.SerializeObject` should ensure Swagger-UI does not depend on the external configuration and in turn does not break if the `Indented` formatting is used.

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerUI/SwaggerUIIndexMiddleware.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerUI/SwaggerUIIndexMiddleware.cs
@@ -59,8 +59,8 @@ namespace Swashbuckle.AspNetCore.SwaggerUI
             {
                 { "%(DocumentTitle)", _options.DocumentTitle },
                 { "%(HeadContent)", _options.HeadContent },
-                { "%(ConfigObject)", JsonConvert.SerializeObject(_options.ConfigObject) },
-                { "%(OAuthConfigObject)", JsonConvert.SerializeObject(_options.OAuthConfigObject) }
+                { "%(ConfigObject)", JsonConvert.SerializeObject(_options.ConfigObject, Formatting.None) },
+                { "%(OAuthConfigObject)", JsonConvert.SerializeObject(_options.OAuthConfigObject, Formatting.None) }
             };
         }
     }


jameswilddev:
I'm curious to know why we're using JSON.parse here; is this different to just using the JSON directly, i.e.

```javascript
  window.onload = function () {
    var configObject = {
  "urls": [
    {
      "url": "/swagger/v1/swagger.json",
      "name": "v1"
    }
  ],
  "validatorUrl": null
};
```

domaindrivendev:
Great catch. I wonder if we should use a separate serializer altogether instead of JsonConvert. There’s other setting that could cause issues too - e.g. ignore nulls

TDabasinskas:
Hi, @domaindrivendev ,

I do have ` NullValueHandling = NullValueHandling.Ignore` for the default JsonConvert settings as well, but at least for a while I haven't noticed it causing any issues. Id addition, that one might be actually OK, as I prefer not to see the `null` values in Swagger output as well, I guess.

domaindrivendev:
Actually we need to keep null because it has explicit meaning for swagger-ui config - see validatorUrl. The of the ConfigObject is it’s a server-side version of the JavaScript config for swagger-ui. So, it should be serialized in a very specific way completely independent of how the app serialized it’s JSON

TDabasinskas:
In that case, maybe we can change the code to something like this?

```csharp
{ "%(ConfigObject)", JsonConvert.SerializeObject(_options.ConfigObject, Formatting.None, new JsonSerializerSettings {NullValueHandling = NullValueHandling.Include}) },
{ "%(OAuthConfigObject)", JsonConvert.SerializeObject(_options.OAuthConfigObject, Formatting.None, new JsonSerializerSettings {NullValueHandling = NullValueHandling.Include}) }
```

Issue #541, opened 2017-11-23
[Feature] Generate Swagger file at postbuild without running the app


poulad-rbc:
### Feature Request

It would be great to generate the Swagger file in CI without running the app. I am willing to build such a feature (either in this repo or create another tool). Using reflection, a small program could read all docs and generate the swagger file.

What are your comments about it?

domaindrivendev:
@poulad-rbc - this is definitely a feature that's long overdue. I'm working on some significant refactors to   make the overall architecture more flexible and ultimately allow this feature to "plug-in" more seamlessly. So, you might want to hold off on your PR until then.

In the meantime you could certainly start spiking into what it would take to generate the Swagger outside of the running application from a command line tool. The tricky part will be getting the complete ApiExplorer infrastructure wired up as Swashbuckle is heavily dependent on it. Easiest thing might be an in memory/test server - you could look at the current IntegrationTests for some inspiration.

When you have a design in mind, it may be prudent to talk through it before you start knocking out the PR, just so we're on the same page. Thanks

PaulDotNet:
I did a very small experiment to get swagger in test/WebSites/Basic project.  Here is the code in the main() function:

        public static void Main(string[] args)
        {
            var host = new WebHostBuilder()
                .UseKestrel()
                .UseContentRoot(Directory.GetCurrentDirectory())
                .UseIISIntegration()
                .UseStartup<Startup>()
                .Build();

            // Trying to get swagger provider here
            var sw = (ISwaggerProvider) host.Services.GetService(typeof(ISwaggerProvider));
            // ... and document here. Which throws exception when it tries to add duplicate "x-purpose" in
            // the AssignOperationVendorExtensions.Apply()
            var doc = sw.GetSwagger("v1", null, "/");

            host.Run();
        }

As you can see from my comments this approach currently does not work. GetService() call breaks something and both sw.GetSwagger and host.Run() fails. If I remove GetSwagger() failure remains.  
Any ideas why is this happening?

qmfrederik:
@PaulDotNet Your code actually works for me; and here's how I serialize the swagger documentation:

```csharp
// Serialize
var swaggerString = JsonConvert.SerializeObject(
    swaggerDoc,
    Formatting.Indented,
    new JsonSerializerSettings
    {
        NullValueHandling = NullValueHandling.Ignore,
        ContractResolver = new CamelCasePropertyNamesContractResolver()
    });

[...]

PR #629, opened 2018-03-02
Don't create a new ISwaggerProvider on every resolution (address #541)

saxonww:
The existing behavior has AddSwaggerGen() register the SwaggerGenerator
as a transient; every resolution of ISwaggerProvider generates a new
instance. Each creation causes all Document, Operation, and Schema
filters to be registered again.

This is slow, and requires that filters check before adding an extension
to avoid generating exceptions.

This may address a problem I encountered and reported in #541

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/Application/SwaggerGenServiceCollectionExtensions.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/Application/SwaggerGenServiceCollectionExtensions.cs
@@ -17,7 +17,7 @@ namespace Microsoft.Extensions.DependencyInjection
 
             services.Configure(setupAction ?? (opts => { }));
 
-            services.AddTransient(CreateSwaggerProvider);
+            services.AddSingleton(CreateSwaggerProvider);
 
             return services;
         }


domaindrivendev:
When using DI frameworks, I generally tend to avoid using "Singleton" lifetimes, particularly in web applications where the likelihood of concurrent requests is much more likely. By sticking with "Transient" lifetimes, I can avoid the complexity of writing and maintaining thread-safe code for the `SwaggerGenerator` and it's dependencies.

There's very little happening in the constructors of these services and therefore I'd be astonished if they have any tangible impact on overall response latency. I'd be far more concerned about the service sitting there taking up permanent space in the consuming applications memory, which would be the case if I merged this PR.

I'm aware of the problem you encountered and reported in #541 and have tracked down the real root cause -  the `Clone` method on `SwaggerGeneratorSettings` and `SchemaRegistrySettings` is only performing a shallow copy, causing some state to be incorrectly kept around between invocations. I'm working on a fix as we speak.

mpetito:
Just ran into the `Clone` issue in my netcoreapp2.0 project when attempting to integrate the swashbuckle cli.  Strangely enough it does not appear to be an issue with projects that target netcoreapp2.1.

@domaindrivendev is this as simple as modifying the `Clone` method to copy the lists?

I'm using the following as a workaround for now:

```C#
var descriptor = services.First(sd => sd.ServiceType == typeof(ISwaggerProvider));
services.Replace(new ServiceDescriptor(
  descriptor.ServiceType, 
  descriptor.ImplementationFactory, 
  ServiceLifetime.Singleton));
```

domaindrivendev:
Yeh I think that’s the probably the simplest thing for now and should solve the immediate issue. I’m not particularly happy with the way the settings are currently plumbed through and would love to simplify but that’s prob best left for another day.

Issue #661, opened 2018-03-22
FromQuery BindNever still show as a valid parameter


Nazgaul:
Hi All.
I got an HttpGet api endpoint and in my model i got a parameter with Bind Never attribute
`[BindNever]
        public string Temp { get; set; }`

The api end result json 

{
name: "temp",
in: "query",
required: false,
type: "string"
}

why is that? in real api request the model binding will never bind this parameter so why do i see it?

If I look at 
https://docs.microsoft.com/en-us/aspnet/core/mvc/advanced/custom-model-binding
and i have Author class in the example - and i only want to get the id from the query string ( the other are populated via the db) how can i manage that swagger will only see the Id property and not the entire class property.

thanks.

PR #668, opened 2018-03-25
Honor BindRequired and BindNever attributes

alexvaluyskiy:
Fixes https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/661 and adds support for BindRequired and BindNever attributes for models
```C#
public class BindingType
{
    public string Property1 { get; set; }

    [BindNever]
    public string Property2 { get; set; }

    [BindRequired]
    public string Property3 { get; set; }
}
```

It also should work for action's parameters (but only in ASP.NET Core 2.1 preview-2 and higher)
```C#
public void AcceptsRequiredParameters([Required, BindRequired] string queryParam) { }
```

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/Generator/JsonPropertyExtensions.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/Generator/JsonPropertyExtensions.cs
@@ -3,6 +3,7 @@ using System.ComponentModel.DataAnnotations;
 using System.Reflection;
 using System.Linq;
 using Microsoft.AspNetCore.Mvc;
+using Microsoft.AspNetCore.Mvc.ModelBinding;
 using Newtonsoft.Json.Serialization;
 
 namespace Swashbuckle.AspNetCore.SwaggerGen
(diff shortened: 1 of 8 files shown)


src/Swashbuckle.AspNetCore.SwaggerGen/Generator/JsonPropertyExtensions.cs
```diff
@@ -13,14 +14,20 @@ internal static bool IsRequired(this JsonProperty jsonProperty)
         {
             return jsonProperty.Required == Newtonsoft.Json.Required.AllowNull
                 || jsonProperty.Required == Newtonsoft.Json.Required.Always
-                || jsonProperty.HasAttribute<RequiredAttribute>();
+                || jsonProperty.HasAttribute<RequiredAttribute>()
+                || jsonProperty.HasAttribute<BindRequiredAttribute>();
```
domaindrivendev: As far as I know, `BindRequired` and `BindNever` are not applicable within types that are bound `FromBody`. Instead, serialization/deserialization for these is handled by input/output formatters. I'm not aware of an deserializer that's aware of MVC model binding attributes.

So, if the framework isn't going to honor these attributes when serializing/deserializing payloads, then I don't think SB should honor them when describing those payloads.

domaindrivendev:
This change makes sense for inferring if a parameter's `required` field should be set or not. But, I don't believe these attributes should be honored when describing `Schema` properties. See my inline comments for more details

domaindrivendev:
Also, you should read the following comments in issue https://github.com/aspnet/Mvc/issues/7399#issuecomment-368575076. The snippet provided represents the approach I'd like to take for inferring a value for the `required` flag on parameters.

PR #674, opened 2018-04-04
Fix SwaggerUI to be CSP compatible

### LINKED ISSUE #505: Doesn't work with CSP
Unable to use this due to the inline script, when using a Content Security Policy without enabling unsafe-inline.

Can this be fixed please?

ErazerBrecht:
Should fix #505

I needed this changes because we hosted the SwaggerUI on a CSP protected domain.
If you're interested feel free to check this PR (and eventually merge) 

Thx for creating this project!
Sincerely, Brecht


domaindrivendev:

Issue #678, opened 2018-04-10
Swagger route direct with 2.4.0


RomainWilbert:
Hi ! 
Since 2.4.0 upgrade, When I access Swagger route (xxx/docs/rest), I am redirected to docs/docs/rest.
If I add a "/" at the end (xxx/docs/rest/) then it works fine. 
Is this normal ?
Here is my config in Startup : 
`            // Enable middleware to serve generated Swagger as a JSON endpoint.
            app.UseSwagger(c => { c.RouteTemplate = "docs/rest/{documentName}/swagger.json"; });

            // Enable middleware to serve swagger-ui (HTML, JS, CSS, etc.), specifying the Swagger JSON endpoint.
            app.UseSwaggerUI(c =>
            {
                c.SwaggerEndpoint("/docs/rest/v1/swagger.json", "v1");
                c.DocumentTitle = ApiName;
                c.RoutePrefix = "docs/rest";
            });`

domaindrivendev:
There was a change to some redirect logic but shouldn’t have broken existing stuff. Will investigate and revert

domaindrivendev:
Should be fixed by the following commit - b84ecc20448f6b8db73c13786a488c890c848145. Could you pull down the preview package from myget.org (https://myget.org/feed/domaindrivendev/package/nuget/Swashbuckle.AspNetCore) and try it out?

RomainWilbert:
I still have the issue with the preview.
I must say that I only reproduce this bug on my development server. My app is a docker app build from microsoft/aspnetcore.
When I run the app on localhost in VS code or in docker, it works fine.

domaindrivendev:
You may need to clear browser cache or just try again in an incognito window. The old incorrect redirect may still be cached in your browser

RomainWilbert:
thanks for the tip but the problem is persistent.

domaindrivendev:
Hmmm it’s working for me with your exact setup so not sure what else I can do at this point. Perhaps you could try with a new project from scratch just to see if it still repros? Also confirm you def have the latest preview version

doctorless:
I'm having the same problem. Also tried the preview and had the same issue.

domaindrivendev:
@doctorless - did you use an ingognito window when you tried the preview?

doctorless:
@domaindrivendev I tried an incognito window, a different browser, and even a different device on a different network that had never seen the endpoint before.

[...]

PR #688, opened 2018-04-13
Fixed issues with #678

doctorless:
Verified working on AWS API Gateway and non-proxied environments.

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerUI/SwaggerUIIndexMiddleware.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerUI/SwaggerUIIndexMiddleware.cs
@@ -7,6 +7,7 @@ using System.Linq;
 using Microsoft.AspNetCore.Http;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
+using System;
 
 namespace Swashbuckle.AspNetCore.SwaggerUI
 {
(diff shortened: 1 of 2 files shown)


domaindrivendev:
Thanks for all the digging and it's very helpful that you've identified the root cause. Regarding your solution though, I'm reluctant to revert to the use of absolute paths (this is how it was implemented in the early days). Constructing absolute paths for the client to use based on the incoming request can cause a myriad of issues in proxy environments.

For example, in many proxy environments `httpContext.Request.Host` will be the host name for the "inner" node (i.e. behind the proxy) as opposed to the edge host name that the client interacts with. Or, if the "inner" node is exposed on a different port, then the client ends up trying to redirect to some weird port. All of these issues and more were reported with the absolute path approach and the switch to relative paths meant they simply don't (or shouldn't) matter.

The AWS API Gateway behavior is incorrect because as subtle as it is - there's a distinct difference between URL's with and without trailing slashes. I don't want to revert to an approach that caused many issues because Amazon has a bug in their API Gateway. I'd certainly like to find a solution but would like it to be based on relative paths.

doctorless:
I can definitely agree on your perspective. Am I correct in assuming the reason for the redirect is to prevent issues with relative paths for resources in the UI (CSS, JS, etc.)? If so, perhaps we can avoid any server-side logic altogether about this and do some manipulation via JavaScript on the index.html page? I'd be happy to investigate this and create a new PR if this is a path you would find better. Or if I'm completely off base, feel free to let me know.

Thanks for commenting on this so quickly. I greatly appreciate the feedback.

domaindrivendev:
Yep, the main reason is both the UI assets and the Swagger endpoints, which can also be relative to the UI page. If you're happy to investigate an alternative solution (I think you're on the right track with the client-side manipulation), that would be great! Thanks for all the help

PR #689, opened 2018-04-13
Client side-based fix for issue #678.

doctorless:
Moved path combination logic to purely client-side in JavaScript.


domaindrivendev:
Thanks - let me stew on this for a bit, I was hoping for a little less client-side logic and still _want_ the current solution to work, because I think it's an elegant and correct solution. Still reluctant to change it because of a bug in another proxy.

domaindrivendev:
@doctorless - please see my [comments](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/678#issuecomment-382133965) in the associated issue. I think we can solve this in a slightly simpler way.

PR #699, opened 2018-04-20
Added initial TagAttribute and Filter implementation

esbenbach:
Added initial TagAttribute and Filter implementation


esbenbach:
This still needs some work. 

But basically just wanted to determine if the approach of having a DocumentFilter is the right way to go?

esbenbach:
Yet another question. Is the idea that adding the TagAttribute to a controller should override the Tags on all the operations, or should it add to the existing operation tags. How about the default "ControllerName" tag that is added, should it be preserved?

esbenbach:
I removed the operation filter. Took another look at the XmlCommentsDocumentFilter, and as far as i can tell it does things in a very similar matter as well. If you prefer I can change my implementation to follow the style of that filter, but personally i prefer the way I have arranged the code (but as always its just a matter of taste.)

PR #700, opened 2018-04-23
Added support for operation Produces/Consumes arrays

esbenbach:
Added support for operation Produces/Consumes arrays.

The implementation should override/preceed whatever would be defined by the frameworks ProducesAttribute and ConsumesAttribute.


domaindrivendev:
Looking good bar the failing tests

esbenbach:
Yearh forgot to run all tests locally. I will fix it within a day or two

esbenbach:
So, the failing test is weird. Running tests inside visual studio AND using the ci\test.ps1 script works just fine on my machine - i even went so far as to clone my fork again just to make sure there was no weird leftovers. On AppVeyor it fails though.

PR #717, opened 2018-05-01
Added SwaggerParameterAttribute and related filter

esbenbach:
Added a new swaggerparameterattribute and related operationfilter to the annotations (and related tests)

Made one of the existing ApiDescriptor extension methods public as it was "nice to have" for the filter.

### DIFF
--- a/src/Swashbuckle.AspNetCore.Annotations/SwaggerParameterAttributeOperationFilter.cs
+++ b/src/Swashbuckle.AspNetCore.Annotations/SwaggerParameterAttributeOperationFilter.cs
@@ -1,0 +1,78 @@
+﻿using Microsoft.AspNetCore.Mvc.Controllers;
+using Swashbuckle.AspNetCore.Swagger;
+using Swashbuckle.AspNetCore.SwaggerGen;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+
+namespace Swashbuckle.AspNetCore.Annotations
+{
+    /// <summary>
+    /// An operation filter that adds parameter metadata on an action.
+    /// defined by the <see cref="SwaggerParameterAttribute" /> on controller methods.
+    /// </summary>
+    /// <remarks>
+    /// This filter overrides any existing description and sets the Required option only if set to true, otherwise it leaves it alone.
+    /// </remarks>
+    /// <seealso cref="Swashbuckle.AspNetCore.SwaggerGen.IOperationFilter" />
+    public class SwaggerParameterAttributeOperationFilter : IOperationFilter
+    {
+        public void Apply(Operation operation, OperationFilterContext context)
+        {
+            EnsureParameters(operation);
+
+            foreach (var pair in this.GetParameterAttributes(context))
+            {
+                var param = GetOrAddParameter(operation, pair.Key);
+                param.Description = pair.Value.Description;
+
+                // Only override if required is set to true, meaning that if required is not set it will be whatever the implementation has detected it should be.
+                // This allows for use of description without having to worry about the required parameter as well - though it prevents to option of using it to se a parameter as not required if the implementation thinks it is!
+                param.Required = pair.Value.Required ? true : param.Required;
+            }
+        }
+
+        /// <summary>
+        /// Ensures the parameters exists on the operation
+        /// </summary>
+        /// <param name="operation">The operation.</param>
+        private static void EnsureParameters(Operation operation)
+        {
+            if (operation.Parameters == null)
+            {
+                operation.Parameters = new List<IParameter>();
+            }
+        }
+
+        /// <summary>
+        /// Gets the or add parameter to the operation collection of parameters
+        /// </summary>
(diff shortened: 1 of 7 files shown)


src/Swashbuckle.AspNetCore.Annotations/SwaggerParameterAttributeOperationFilter.cs
```diff
@@ -0,0 +1,78 @@
...
+        }
+        /// <summary>
+        /// Gets the parameter attributes from the operation
+        /// </summary>
+        /// <param name="context">The context.</param>
+        /// <returns>A dictionary of parameter, attribute pairs</returns>
+        private Dictionary<string, SwaggerParameterAttribute> GetParameterAttributes(OperationFilterContext context)
+        {
+            var controllerActionDescriptor = context.ApiDescription.ControllerActionDescriptor();
+            return controllerActionDescriptor?.Parameters?.OfType<ControllerParameterDescriptor>()
```
domaindrivendev: If `controllerActionDescriptor` and/or `Parameters` is null, then this method will return null and cause a null ref up the stack (i.e. in the foreach loop in the Apply method)

src/Swashbuckle.AspNetCore.Annotations/SwaggerParameterAttributeOperationFilter.cs
```diff
@@ -0,0 +1,78 @@
...
+                .Select(param => new { Name = param?.Name, Attribute = param?.ParameterInfo?.GetCustomAttributes(false)?.OfType<SwaggerParameterAttribute>().SingleOrDefault() })
+                .Where(attr => attr?.Attribute != null)
+                .ToDictionary(k => k.Name, v => v.Attribute);
```
domaindrivendev: The excessive use of null coalescing in this method makes it difficult to follow and, IMO, increases the likelihood of a null ref being missed. I generally favor readability over shortcuts - sometimes you can get both but I'm not convinced that's the case here. More of a style pref than anything else and I'm happy to adjust some of the style things before the final merge into master. So, feel free to leave as is for now

src/Swashbuckle.AspNetCore.Annotations/SwaggerParameterAttributeOperationFilter.cs
```diff
@@ -0,0 +1,78 @@
...
+        }
+
+        /// <summary>
+        /// Gets the or add parameter to the operation collection of parameters
+        /// </summary>
+        /// <param name="operation">The operation.</param>
+        /// <param name="name">The name of the parameter</param>
+        /// <returns></returns>
+        private static IParameter GetOrAddParameter(Operation operation, string name)
+        {
```
domaindrivendev: I don't think the conditional addition is necessary here. I can't think of a scenario where you would have a request parameter in the method signature that hasn't already been mapped to a Swagger `Parameter` by the generator. Furthermore, if we were to add parameters here - they would need to have values for `Name` and `In` to be valid.

esbenbach: Actually the only reason for that was the testing. It failed because when i setup the test there was nothing (the generator is not there when testing).

I could just change the test to introduce the parameter, but was uncertain if it could happen in a proper runtime scenario.

src/Swashbuckle.AspNetCore.Annotations/SwaggerParameterAttributeOperationFilter.cs
```diff
@@ -0,0 +1,78 @@
...
+            var parameter = operation.Parameters?.SingleOrDefault(p => p.Name == name);
```
domaindrivendev: Depending on how the action parameter is bound to the request, this may not always locate the parameter correctly. For example, consider the following:

```csharp
public IActionResult GetProducts([FromQuery(Name="kw")]string keywords)
```

In this case, the Swagger `Parameter` will (correctly) have the name `kv` but the name it's being compared to will be `keywords`.

I've had to work around this in the Xml filters and used the following logic to do so:

```csharp
var actionParameter = apiDescription.ActionDescriptor.Parameters
    .FirstOrDefault(p => parameter.Name.Equals(
        (p.BindingInfo?.BinderModelName ?? p.Name), StringComparison.OrdinalIgnoreCase));
if (actionParameter == null) continue;
```

I have a plan to centralize this logic so it doesn't have to be repeated but that requires a broader refactor. For now, I would just repeat it here.

esbenbach: Damn, wasnt aware you could override the name in the decoration - will update it later. Should probably also add a separate test for it

esbenbach:
Sorry for the delay - anyway i went ahead and changed the implementation so it resembles the xml comment filter a bit more (not entirely though). I think this also fixes the above mentioned issue with parameter names not being matched properly.

PR #731, opened 2018-05-14
Generic parameter in generic base controller - part of issue #730

### LINKED ISSUE #730: Inheriting custom base class for Controller crashes swagger
This:

public class UserController : MyClass.DbController

Crashes swagger.

This:

public class UserController : Controller

Works.

So, swagger cannot work if we inherit custom base clases?

image

aaxelm:
Checking if the full name of the type is null before trying to use it to get the name without the type arguments. When using generic parameters in generic base controllers, this is needed because they get here with null full names.

Also added fix for multiple methods with the same name in generic base controllers. Not selecting the source method differently for generic types anymore because it doesn't seem to be needed. From initial tests, the value gets selected correctly from the XML documentation file.

Fixed #730

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/Annotations/XmlCommentsMemberNameHelper.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/Annotations/XmlCommentsMemberNameHelper.cs
@@ -7,13 +7,13 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
 {
     public class XmlCommentsMemberNameHelper
     {
-        public static string GetMemberNameForMethod(MethodInfo method)
+        public static string GetMemberNameForMethod(MethodInfo method, Type[] inputParametersTypes)
         {
             var builder = new StringBuilder("M:");
 
             // If method is from a constructed generic type, use the generic type
             var sourceMethod = method.DeclaringType.IsConstructedGenericType
-                ? method.DeclaringType.GetGenericTypeDefinition().GetMethod(method.Name)
+                ? method.DeclaringType.GetGenericTypeDefinition().GetMethods().FirstOrDefault(x => x.Name == method.Name && x.GetParameters().Select(p => p.ParameterType).SequenceEqual(inputParametersTypes))
                 : method;
 
             builder.Append(sourceMethod.DeclaringType.FullNameSansTypeArguments().Replace("+", "."));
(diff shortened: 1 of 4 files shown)


test/Swashbuckle.AspNetCore.SwaggerGen.Test/Annotations/XmlCommentsMemberNameHelperTests.cs
```diff
@@ -25,7 +26,7 @@ public XmlCommentsMemberNameHelperTests(ITestOutputHelper output)
         [InlineData(typeof(XmlAnnotatedType), "AcceptsArrayOfConstructedGenericType",
             "M:Swashbuckle.AspNetCore.SwaggerGen.Test.XmlAnnotatedType.AcceptsArrayOfConstructedGenericType(System.Nullable{System.Int32}[])")]
         [InlineData(typeof(XmlAnnotatedGenericType<int,string>), "AcceptsTypeParameters",
-            "M:Swashbuckle.AspNetCore.SwaggerGen.Test.XmlAnnotatedGenericType`2.AcceptsTypeParameters(System.Int32,`0,`1)")]
+            "M:Swashbuckle.AspNetCore.SwaggerGen.Test.XmlAnnotatedGenericType`2.AcceptsTypeParameters(System.Int32,System.Int32,System.String)")]
```
domaindrivendev: This is not the desired behavior. You've changed the expected member name to a value that can't possibly be found in an XML comments file. `XmlAnnotatedGenericType<int, string>` is derived from a generic type. It _doesn't_ have it's own class definition and therefore doesn't have it's own XML comments.

If a controller is constructed from a generic type then the desired behavior is to use the XML comments from the generic type because XML comments for the specific type simply won't be available. This feature was implemented to support the common pattern of CRUD actions via a generic base controller:

```csharp
public abstract class GenericResourceController<TResource> where TResource : new()
{
    /// <summary>
    /// creates a resource
    /// </summary>
    /// <param name="resource">The resource representation</param>
    /// <returns></returns>
    [HttpPost]
    public int Create([FromBody, Required]TResource resource)
    {
        return 1;
    }

    ...
}

[Route("orders")]
public class OrdersController : GenericResourceController<Order>
{ }

[Route("products")]
public class ProductsController : GenericResourceController<Product>
{ }
```

In it's current state, this PR breaks this functionality.

src/Swashbuckle.AspNetCore.SwaggerGen/Annotations/XmlCommentsMemberNameHelper.cs
```diff
@@ -11,15 +11,10 @@ public static string GetMemberNameForMethod(MethodInfo method)
         {
             var builder = new StringBuilder("M:");
 
-            // If method is from a constructed generic type, use the generic type
-            var sourceMethod = method.DeclaringType.IsConstructedGenericType
-                ? method.DeclaringType.GetGenericTypeDefinition().GetMethod(method.Name)
-                : method;
```
domaindrivendev: See comments below

PR #732, opened 2018-05-15
Add parameter filters

domaindrivendev:


JonLopezGarcia:
I just realized that I need this feature 👍 
When do you think it will be released ?

PR #735, opened 2018-05-17
This is update to ASP.NET Core 2.0

kant2002:
I see rationale in the #464, but it was justifiable at that time. Right now I think lot of green projects start with .NET Core 2.0, some older one migrate from .NET Core 1.0 and don't have 1.6 dependencies, since major library vendors provide versions for both frameworks separately. For example JSON.Net.

The #473 seems to be created on the older version, and thus stale. I just make updating for @domaindrivendev easier. All dependencies for core projects taken have latest version with security fixes, and as such I have to update dependencies in the tests. This is the reason why not only 4 major projects was changed.


domaindrivendev:
I have a branch already sitting in the wings for this. It’s just a matter of _when_ to pull the trigger. There’s one or two key issues that I would like to fix before cutting off the pre 2.0 folks from future Swashbuckle updates. I’m hoping to get this done in a matter of weeks so hopefully people can live with the trivial workaround until then

PR #781, opened 2018-06-28
Add overload to support HttpStatusCode enum

davidsavagejr:
Adds an overload so that you don't have to cast to int all over the place

`[SwaggerResponse((int)HttpStatusCode.Unauthorized, null, "A valid user session is required")]`

becomes

[SwaggerResponse(HttpStatusCode.Unauthorized, null, "A valid user session is required")]


jonmeyerson:
I just used StatusCodes.Status200OK, from https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.http.statuscodes?view=aspnetcore-2.1

domaindrivendev:

Issue #791, opened 2018-07-05
ISwaggerProvider can't be requested multiple times


0x53A:
This code works on the first request, and fails on the second request:

```C#
                               var svc1 = ctx.RequestServices.GetRequiredService<ISwaggerProvider>();
                               var doc1 = svc1.GetSwagger(doc, null, ctx.Request.PathBase);
```

This fails on the first request on line ``doc2 =``:

```C#
                               var svc1 = ctx.RequestServices.GetRequiredService<ISwaggerProvider>();
                               var doc1 = svc1.GetSwagger(doc, null, ctx.Request.PathBase);

                               var svc2 = ctx.RequestServices.GetRequiredService<ISwaggerProvider>();
                               var doc2 = svc1.GetSwagger(doc, null, ctx.Request.PathBase);
```

> System.ArgumentException: 'Ein Element mit dem gleichen Schlüssel wurde bereits hinzugefügt.' (An element with the same key was already added)

```
 	mscorlib.dll!System.ThrowHelper.ThrowArgumentException(System.ExceptionResource resource)	Unknown
 	mscorlib.dll!System.Collections.Generic.Dictionary<string, Swashbuckle.AspNetCore.Swagger.Header>.Insert(string key, Swashbuckle.AspNetCore.Swagger.Header value, bool add)	Unknown
 	mscorlib.dll!System.Collections.Generic.Dictionary<System.__Canon, System.__Canon>.Add(System.__Canon key, System.__Canon value)	Unknown
 	Swashbuckle.AspNetCore.Examples.dll!Swashbuckle.AspNetCore.Examples.AddResponseHeadersFilter.Apply(Swashbuckle.AspNetCore.Swagger.Operation operation, Swashbuckle.AspNetCore.SwaggerGen.OperationFilterContext context)	Unknown
 	Swashbuckle.AspNetCore.SwaggerGen.dll!Swashbuckle.AspNetCore.SwaggerGen.SwaggerGenerator.CreateOperation(Microsoft.AspNetCore.Mvc.ApiExplorer.ApiDescription apiDescription, Swashbuckle.AspNetCore.SwaggerGen.ISchemaRegistry schemaRegistry)	Unknown
[...]

PR #792, opened 2018-07-05
#791: make ISwaggerProvider singleton

0x53A:
I see you have a CI, let me just test [my theory](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/791) ...

If this doesn't break anything, I'll try to add a unit test for my case.


domaindrivendev:
Your theory is correct - however failure when using transient lifetime is a symptom rather than a cause. See #629 for more details

PR #817, opened 2018-07-17
Prefer actions for response descriptions over controllers'

martincostello:
I recently upgraded a project to `3.0.0` and found that the response descriptions for my project had changed.

I have a base controller that defines some baseline `SwaggerResponse` attributes for various HTTP status codes, but derived controllers and actions would re-define the response for certain status codes if the documentation needed to be different, such as to be more specific based on action or controller specific implementation details.

For example:

```csharp
[SwaggerResponse(400, "The value of the foo header is invalid.")]
public abstract class MyBaseController : Controller
{
}

[SwaggerResponse(400, "The value of the foo or bar header is invalid.")]
public abstract class MyController : MyBaseController
{
    [SwaggerResponse(400, "The value of the foo, bar or baz header is invalid.")]
    public IActionResult MyAction()
    {
        return Ok(new{ foo = "bar" });
    }
}
```

With `3.0.0`, all responses gave the description for `400` as `The value of the foo header is invalid.`, instead of using the more-derived response description where available.

This PR adjusts `AnnotationsOperationFilter` to only set the `Description` if a value has not already been set so that precedence is retained, rather than the current _last-one wins_ behaviour.

I've also fixed the compiler warning generated by the obsolete property usage from `IDocumentFilter` in its implementation in `DocumentFilterContext`.


martincostello:
@domaindrivendev Any chance of a review and merge of this PR please? 😃

martincostello:
How come this has been closed?

domaindrivendev:

martincostello:
New PR: #890

PR #820, opened 2018-07-18
Allow to change a prefixed uri behind a proxy for the swagger enpoints

bhugot:


domaindrivendev:
It’s difficult for me to assess the solution without a clear description of the problem. Is there an issue you can point me to that describes the problem being solved here? If not, it would be good to create one so I can understand a little better. I’ve put a lot of work into supporting relative URIs, both host relative and page relative (i.e. relative to the swagger-ui page, with proxy support in mind.

domaindrivendev:
Closing due to inactivity

bhugot:
Yeah sorry answer because I thought our problem come from this but it was internal tooling that seems to be wrong. I can t work on it for now but as soon I can if still needed I will submit again

PR #832, opened 2018-07-27
MinimumAttribute / MaximumAttribute with tests

mrtristan:
these attributes were added to address the situation where sometimes it makes sense to specify one bound without the other.

An example that drove me to do this is pagination. I want the user to specify a `pageIndex` which has a minimum of 1, but no logical maximum due to an unknown quantity of data.

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/Attributes/MaximumAttribute.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/Attributes/MaximumAttribute.cs
@@ -1,0 +1,21 @@
+﻿using System;
+
+namespace Swashbuckle.AspNetCore.SwaggerGen
+{
+    /// <summary>
+    /// Adds a maximum value to a given property
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Property, AllowMultiple = false)]
+    public class MaximumAttribute : Attribute
+    {
+        public MaximumAttribute(int maximum)
+        {
+            Maximum = maximum;
+        }
+
+        /// <summary>
+        /// The maximum value for the decorated property.
+        /// </summary>
+        public readonly int Maximum;
+    }
+}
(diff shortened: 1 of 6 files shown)


mrtristan:
let me know if there is anything you'd like me to change or if this is ok

domaindrivendev:
At it's core, Swashbuckle is intended to infer accurate API descriptions directly from the code that implements the API. In other words, the "code is truth" and therefore the documentation always remains in sync with "actual" behavior. For example, when you're using the standard data annotations (e.g. `Range`, `Required` etc.) in conjunction with ASP.NET Core's built-in model validation (i.e. `ModelState.IsValid`), you can annotate properties to actually implement validation behavior, and the documentation will be automatically updated to reflect that change. That's the overarching goal of Swashbuckle - you implement your API and it describes exactly what you've implemented.

So, introducing new attributes that are for documentation only (i.e. don't drive any "actual" behavior) has always felt orthogonal to this goal. With that said, I also recognize that it's convenient to have attributes when you need to fill in some gaps. So, I introduced an additional "add-on" package called `Swashbuckle.AspNetCore.Annotations` (in this repo) that consumers can opt-into to use additional Swagger-specific attributes for this purpose.

At present, that library supports `SwaggerOperation`, `SwaggerParameter` and `SwaggerResponse` attributes but NOT a `SwaggerSchema` attribute. It seems to me that the latter would be the appropriate place put this functionality. For example, a possible implementation would allow you to decorate properties with this attribute to add additional metadata (such as minimum or maximum) to the corresponding, generated Schema:

```csharp
public class Page
{
    [SwaggerSchema(Minimum = 1)]
    public int Index { get; set; }

    public int Size { get; set; }
}
```

So, I'm going to close this PR and suggest you look into that approach instead. But, I do think this will require a little more effort so I would also urge you to consider if it's really necessary for your use-case. For example, if you're using a numeric type for the index field such as int, double etc. then there _really_ is an upper bound - if the consumer provides a value that's too large for that type then the deserialization will fail. So something like `[Range(1, int.MaxValue)]` might be sufficient.

PR #834, opened 2018-07-31
Added support for custom WebHostBuilder when using cli

tburi:
This enables the CLI tool to run on projects with custom webhost configurations. 
ex. adding autofac, custom config files etc.

Added an example project with autofac. CliExampleWithFactory


domaindrivendev:

PR #880, opened 2018-09-07
OpenAPI conversion layer

LEsperanca:
Added a layer to transform the resulting Swagger json into and OpenAPI json. It is also possible to return YAML


LEsperanca:
It seams that darrel miller is already doing the same 
 OpenAPI conversion Layer https://github.com/darrelmiller/Swashbuckle.AspNetCore

Issue #881, opened 2018-09-08
Implicitly use output of project when running dotnet swagger 


kant2002:
When running `dotnet swagger tofile` you should always specify assembly where to look for the Open API definitions. Would be good if the path would be automatically constructed based on the 
a) project path
b) project configuration

Since many projects follow conventions when specify names, this allow easier way to specify command 
lines. Also this is more in line with `dotnet` tooling itself. 

## Original command
```
dotnet swagger tofile --output api.json bin/Debug/netcoreapp2.1/Api.dll v1
```

## Proposed variants

```
dotnet swagger tofile --output api.json -p . v1
# dotnet swagger tofile --output api.json ./bin/Debug/{anyfolder}/{DirectoryName}.dll v1
```

```
dotnet swagger tofile --output api.json -p . -c Release v1
# dotnet swagger tofile --output api.json ./bin/Release/{anyfolder}/{DirectoryName}.dll v1
```

domaindrivendev:
I like this proposal - would you be interested in submitting a PR?

kant2002:
@domaindrivendev May I ask you why do you not use some .NET Core CLI packages, like http://github.com/natemcmaster/CommandLineUtils or at least `Microsoft.Extension.Configuration` namespaces. This is allow not manage parsing of command line arguments.

Example .NET Configuration (which already likely pulled into project)

```csharp
var switchMappings = new Dictionary<string, string> 
{ 
    { "-c", "Configuration" }, 
}; 
var configuration = new ConfigurationBuilder() 
    .AddCommandLine(args, switchMappings) 
    .Build(); 
…
var buildConfiguration = configuration["Configuration"] ?? "Debug"; 
```
Not pushing for this, just asking what's your perspective.

PR #882, opened 2018-09-09
Added --project and --configuration parameters

kant2002:
These parameters allow specify location of the project, and build configuration to use for getting location of the assembly from which load Swagger definition.

See #881


domaindrivendev:

PR #887, opened 2018-09-20
Added support for custom WebHostBuilder when using cli

tburi:
This enables the CLI tool to run on projects with custom webhost configurations.
ex. adding autofac, custom config files etc.

Added an example project with autofac. CliExampleWithFactory


kipmaplewave:
This PR is also important for current projects we are working on.  We have a similar patch, but obviously don't want to maintain a fork.  Is their an ETA when this will be added to master?

officeSpacePirate:
@domaindrivendev Is this PR a direction you're good to go with? If so and @tburi doesn't have the time, I would be willing to resolve the conflicts in a new PR. Equivalent support is blocking my team from using the CLI tool.

brooknovak:
Why is this not being merged? 
We were using this nice CLI tool, but its now breaking due to our projects now using the webhost builder configure method for a custom service provider (in our case we are using castle windsor).

michalsteyn:
Would also love this feature

domaindrivendev:
Merged with some minor code cleanup

PR #907, opened 2018-10-11
[REOPEN] Allow generating Swagger specification from multiple Startup classes

itsprdsw:
Please see [comment](https://github.com/itsprdsw/Swashbuckle.AspNetCore/pull/1#issuecomment-427287255) in https://github.com/itsprdsw/Swashbuckle.AspNetCore/pull/1


itsprdsw:
@domaindrivendev could you have a look at the PR? If you're interested in adding this functionality, we will be pleased to update it against master

domaindrivendev:
55 files is a lot of changes for this PR! It's much easier for me to review PR's that are focused on a very specific change. I understand that there's additional changes (framework version updates etc.) that you're trying to piggyback onto this one but IMO, that just adds too much noise. So, if you could boil it down to just the change in question that would be very helpful.

I'm also wondering if the simplest possible approach here would be just adding a `--startupClass` option to the CLI tool?

PR #913, opened 2018-10-15
Add schema to response

bherila:
The swaggerResponseAttribute does not register the `Type`, this PR fixes that bug.

### DIFF
--- a/src/Swashbuckle.AspNetCore.Annotations/AnnotationsOperationFilter.cs
+++ b/src/Swashbuckle.AspNetCore.Annotations/AnnotationsOperationFilter.cs
@@ -19,7 +19,7 @@ namespace Swashbuckle.AspNetCore.Annotations
 
             ApplySwaggerOperationAttribute(operation, actionAttributes);
             ApplySwaggerOperationFilterAttributes(operation, context, actionAndControllerAttributes);
-            ApplySwaggerResponseAttributes(operation, actionAndControllerAttributes);
+            ApplySwaggerResponseAttributes(operation, actionAndControllerAttributes, context);
         }
 
         private static void ApplySwaggerOperationAttribute(
@@ -71,7 +71,8 @@ namespace Swashbuckle.AspNetCore.Annotations
 
         private void ApplySwaggerResponseAttributes(
             Operation operation,
-            IEnumerable<object> actionAndControllerAttributes)
+            IEnumerable<object> actionAndControllerAttributes,
+            OperationFilterContext context)
         {
             var swaggerResponseAttributes = actionAndControllerAttributes
                 .OfType<SwaggerResponseAttribute>();
@@ -86,6 +87,9 @@ namespace Swashbuckle.AspNetCore.Annotations
 
                 if (swaggerResponseAttribute.Description != null)
                     response.Description = swaggerResponseAttribute.Description;
+
+                if (swaggerResponseAttribute.Type != null)
+                    response.Schema = context.SchemaRegistry.GetOrRegister(swaggerResponseAttribute.Type);
 
                 operation.Responses[statusCode] = response;
             }


domaindrivendev:
This is not necessary as this attribute derives from ProducesResponseType and therefore the status and type registration will be automatically applied by the core generator

bherila:
Hmm... it was not applying on my Azure Functions test, including if I used ProducesResponseType directly. Any pointers how I can fix it there?

domaindrivendev:
Are you using the out-of-the-box ApiExplorer or some custom version? If so, then it’s possible the custom version is not honoring the ProducesResponseType attribute.

bherila:
Yes, it's [FunctionApiDescriptionGroupCollectionProvider.cs](https://github.com/bherila/Swashbuckle.AspNetCore/blob/azure-functions/src/Swashbuckle.AspNetCore.AzureFunctions/Providers/FunctionApiDescriptionGroupCollectionProvider.cs) but I'm not doing anything special with either type. Thanks for your help... I'm stumped :(

bherila:
I even searched the whole thing and it's not clear where the ProducesResponse type is applied?

```
Find all "ProducesResponse", Match case, Subfolders, Keep modified files open, Find Results 1, Entire Solution, "*.cs*"
  C:\proj\wip\Swashbuckle.AspNetCore\test\Swashbuckle.AspNetCore.SwaggerGen.Test\TestFixtures\Fakes\FakeApiDescriptionGroupCollectionProvider.cs(101):            descriptor.FilterDescriptors = descriptor.MethodInfo.GetCustomAttributes<ProducesResponseTypeAttribute>()
  C:\proj\wip\Swashbuckle.AspNetCore\test\Swashbuckle.AspNetCore.SwaggerGen.Test\TestFixtures\Fakes\FakeController.cs(123):        [ProducesResponseType(typeof(void), 204)]
  C:\proj\wip\Swashbuckle.AspNetCore\test\Swashbuckle.AspNetCore.SwaggerGen.Test\TestFixtures\Fakes\FakeController.cs(124):        [ProducesResponseType(typeof(IDictionary<string, string>), 400)]
  C:\proj\wip\Swashbuckle.AspNetCore\test\WebSites\Basic\Controllers\ResponseTypeAnnotationsController.cs(18):        [ProducesResponseType(typeof(int), 201)]
  C:\proj\wip\Swashbuckle.AspNetCore\test\WebSites\Basic\Controllers\ResponseTypeAnnotationsController.cs(19):        [ProducesResponseType(typeof(IDictionary<string, string>), 400)]
  C:\proj\wip\Swashbuckle.AspNetCore\src\Swashbuckle.AspNetCore.Annotations\SwaggerResponseAttribute.cs(10):    public class SwaggerResponseAttribute : ProducesResponseTypeAttribute
  Matching lines: 6    Matching files: 4    Total files searched: 187
```

bherila:
Any thought @domaindrivendev ? This PR is important to be able to get my independent Azure Functions working with the stock Swashbuckle nuget packages as dependencies.

bherila:
Since I could not wait any longer I have published 

```
BH.Swashbuckle.AspNetCore.AzureFunctions.1.0.0
BH.Swashbuckle.AspNetCore.Annotations.4.0.0
BH.Swashbuckle.AspNetCore.SwaggerGen.4.0.0
BH.Swashbuckle.AspNetCore.Swagger.4.0.0
```

domaindrivendev:
FunctionApiDescriptionGroupCollectionProvider is a your own implantation correct? You should update this to honor the presence of ProducesResponseType attributes, which it should because this is part of that interface. Then the existing SB filters will work as expected

bherila:
The original code is checking for SwaggerResponseAttribute and not setting the response.Schema. But it is setting the response.Description. Why should it be setting the response.Description and not the response.Schema? It is not clear how response.Schema is getting injected by the default ApiExplorer. Therefore, I don't know how my FunctionApiDescriptionGroupCollectionProvider is supposed to inject it :(

What I *do* know is that the code in this PR is working and is not creating any regressions. I will have to continue using my forked packages in the meanwhile (at least I rebased them to `master` and everything seems to be working fine).

domaindrivendev:
I'm a big fan of what you're doing with the introduction of a `Swashbuckle.AspNetCore.AzureFunctions` package and would love to get on the same page so you don't have to fork the core Swashbuckle packages.

The `SwaggerResponseAttribute` is derived from the `ProducesResponsTypeAttribute` and this is the key to _why_ it doesn't need to set the `response.Schema`, at least with the current implementaion. To help explain, consider the sequence of events:

1. `SwaggerGenerator` retrieves an `ApiDescription` from the injected `IApiDescriptionGroupCollectionProvider` that's built-in to ASP.NET Core

_The built-in provider honors the presence of `ProducesResponseTypeAttribute`s on actions and returns a corresponding list of `ApiResponseType`s with the apiDescription_

2. The `SwaggerGenerator` inspects those response types and uses them to generate `Response` metadata, including a `Schema` that's generated from the corresponding `Type`

3. The `SwaggerGenerator` feeds the generated `Operation` to the configured `AnnotationsOperationFilter`. Due to the previous steps, this filter makes a _reasonable_ assumption that the `Schema` has already been set.

Of course, this breaks down as soon as there's an implementation of `IApiDescriptionGroupCollectionProvider` that doesn't honor the presence of `ProducesResponseTypeAttribute`s. I was initially arguing that the `ProducesResponseTypeAttribute` is an implicit part of the `IApiDescriptionGroupCollectionProvider` interface and should be honored by _all_ implementations.

However, I can see why this could be debatable. If you disagree with my reasoning above and are unwilling to add the attribute honoring behavior to your `FunctionApiDescriptionGroupCollectionProvider`, then I'd be willing to accept this PR if meant we could avoid the introduction of Swashbuckle forks. Let me know your thoughts?

bherila:
Thank you for the extended description, I agree with your rationale that this additional logic should not need to be added since both attributes do extend from `ProducesResponseTypeAttribute`. The trouble is, after exploring this as well as what ApiExplorer code I've been able to find online for quite some time now, I've been unable to figure out how to do this. Could you suggest where I might be able to add the desired behavior, and I will be happy to make those adjustments?

domaindrivendev:
Turns out it's not that simple - deriving from `ProducesResponseTypeAttribute` does provide benefits when the out-of-the-box `IApiDescriptionGroupCollectionProvider` is in use. So, although it's redundant for all cases except yours, I've accepted your initial PR as it shouldn't have any negative side effects either way.

bherila:
I had pushed some wip to that branch so thank you for cherry picking it!

I will test against the latest master branch soon.

bherila:
For anyone who finds this issue, here is the answer: https://www.nuget.org/packages/Swashbuckle.AspNetCore.AzureFunctions/

Issue #845, opened 2018-08-08
"dotnet swagger tofile" causes an error when user home directory contains a space


vdachev:
Hello,

It seems there's an issue with the "Swashbuckle.AspNetCore.Cli" NuGet package (version 3.0.0-beta1) that causes the tool to crash when called by a user whose home directory (in Windows) contains a space.

In my case, the home directory of the user is "_C:\Users\Valery Dachev_". An attempt to execute the following command (in another directory):
```cmd
dotnet swagger tofile --format Indented --output "bin\Debug\netcoreapp2.1\swagger.json" "bin\Debug\netcoreapp2.1\MyWebApp.dll" v1
```
results in the following error:
> The application to execute does not exist: 'C:\Users\Valery'

I was able to successfully call the same command line on another computer where my home directory is just "_C:\Users\Valery_".

boindiil:
There is also a problem when using a space within the arguments

```powershell
# works
dotnet swagger tofile --output '.\bin\Debug\netcoreapp2.1\swagger.json' "bin\Debug\netcoreapp2.1\Web.Host.dll" v1

# does not work
dotnet swagger tofile --output '.\bin\Debug\netcoreapp2.1\Test Space\swagger.json' "bin\Debug\netcoreapp2.1\Web.Host.dll" v1

# does not work
dotnet swagger tofile --output '.\bin\Debug\netcoreapp2.1\swagger.json' "bin\Debug\netcoreapp2.1\Test Space\Web.Host.dll" v1
```

FrozenKiwi:
comment for anyone blocked on this - one workaround is to move your nuget packages folder: https://stackoverflow.com/questions/47860222/change-nuget-package-location-folder

PR #915, opened 2018-10-16
Excape path in cli arguments

Cmouse87:
Reslove issue #845


Cmouse87:
Thanks a lot! When you could release this changes? I make this fix out of nescessity.

PR #917, opened 2018-10-19
枚举类型显示对应的描述

caopengfei:


domaindrivendev:
Please provide a problem statement - what problem does this solve, and how does it solve it

PR #919, opened 2018-10-19
Use ApiParameterDescription.IsRequired when available

pranavkm:
Takes advantage of https://github.com/aspnet/Mvc/blob/release/2.2/src/Microsoft.AspNetCore.Mvc.Abstractions/ApiExplorer/ApiParameterDescription.cs#L56 added to `ApiParameterDescription` in 2.2

### DIFF
--- a/test/Swashbuckle.AspNetCore.SwaggerGen.Test/Generator/SwaggerGeneratorTests.cs
+++ b/test/Swashbuckle.AspNetCore.SwaggerGen.Test/Generator/SwaggerGeneratorTests.cs
@@ -1,11 +1,11 @@
 ﻿using System;
-using System.Linq;
 using System.Collections.Generic;
+using System.Linq;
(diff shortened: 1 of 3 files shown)


test/Swashbuckle.AspNetCore.SwaggerGen.Test/Generator/SwaggerGeneratorTests.cs
```diff
@@ -689,5 +727,12 @@ private class ApiResponseTypeV2 : ApiResponseType
         {
             public bool IsDefaultResponse { get; set; }
         }
+
+        private class ApiParameterDescription22 : ApiParameterDescription
```
odekhtiarenko:
```suggestion
        private class ApiParameterDescriptionStub : ApiParameterDescription
```

PR #946, opened 2018-11-21
Properties which implement KeyedCollection should emit uniqueItems: true

### LINKED ISSUE #945: Properties which implment KeyedCollection should emit uniqueItems: true
Currently Properties who's type are or implement KeyedCollection are emitted as arrays without uniqueItems: true this results in API documentation that does not properly represent the behavior of the type.

This is similar to the problem raised in #430 for properties implementing ISet

freeranger:
Fixes #945 
KeyedCollection by it's nature is a collection of unique items, so uniqueItems: true should be set in the schema for properties of this type or implementing this type


freeranger:
Needs more work

PR #949, opened 2018-11-21
Allow generating Swagger specification from multiple, unknown Startup classes

itsprdsw:
As a continuation of PR  #907:

* Add SwaggerStartupAttribute to denote Startup classes, which has been isolated in a new project (for now) together with related helper methods. This attribute has:

  * A required OpenApiFileName field, which will be used by Swashbuckle.AspNetCore.Cli to generate correspondent OpenApi specification.
   * Optional ClientClassName and ClientNamespace, to allow defining clients names and namespaces.

* Edit Swashbuckle.AspNetCore.Cli to add new --multiplestartups option, so that mentioned functionality remains optional and existing behavior doesn't change.

-----------------------------------

Usage:

`dotnet swagger tofile --multiplestartups true --output bin bin\$(AssemblyName).dll v1.0`

Implementation has been done using an attribute instead of something similar to [suggested `--startupClass` option](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/pull/907#issuecomment-440250717) because this way we can keep cli command independent from concrete projects (besides `$(AssemblyName)` and api version), without having to know how many startups exist.
This is an important point for the usage we're making of `Swashbuckle.AspNetCore.Cli`, together with `ClientClassName` and `ClientNamespace` optional fields, since it allows us to control both Swagger spec. and clients generation directly from our code.

Attribute usage examples:

```
    [SwaggerStartup(
        openApiFileName: "FooPublicOpenApi",
        ClientClassName = "PublicClient",
        ClientNamespace = "Public.Client")]
    public class PublicStartup { /*...*/ }
```

```
    [SwaggerStartup(
        openApiFileName: "FooInternalOpenApi")]
    public class InternalStartup { /*...*/ }
```

-----------------

Both specific tests for `Swashbuckle.AspNetCore.SwaggerStartupAttr` and a sample project with multiple startups have been implemented, we haven't added them (yet) to the PR to keep it as simple as possible for its evaluation, as @domaindrivendev suggested.


domaindrivendev:
Thankyou for taking the time to simplify this PR - it's made it a lot easier for me to understand the changes you're proposing. However, I'm still uncomfortable with the overall approach. It feels awkward to add a separate project just hold an attribute that only really has meaning for the CLI project. Would the CLI project not be a more suitable home for the attribute?

Perhaps we could take a step back from the implementation detail and try to align on what your trying to do? What's your use case? It seems the dynamic discovery of Startup classes is important to you vs enumerating in a config file or something similar. But, I don't have the context to understand why. I'm thinking you may be trying to build some automation (perhaps for CI/CD?) where you can point the same command to an arbitrary project. If that were the case, then I might suggest that the auto discovery of startup class names is a concern of the automation rather than the Swagger CLI tool.

Even if we were to reach agreement on the approach, then there's a couple of changes I might suggest with the current code changes. For example, there's quite a bit of duplicated logic in the paths that fall either side of the `multiplestartups` option. Also, the `output` parameter is optional in one case and required in the other which breaks the principle of least surprise in API design. But, probably best to table that discussion until we're aligned on what you're trying to solve and how best to go about it.

itsprdsw:
### Our use case:
We've a number of web api projects, some of them with multiple startup classes and some of them with a single one, being those numbers modified quite frequently.
We maintain OpenApi specification and clients generation for all of them **centralized in a single place** (used for CI/CD too, as you correctly guessed), that's why we felt the need of being able to change the number of startup classes and choose the name of their spec files and clients to be generated from them without modifying generic, Swagger generation invocations/code.
All that motivated us to modify our Swagger spec generation tool (yours) to allow that flexibility.

### Our motivation about current approach and pull request:
Let's suppose we had chosen suggested **`--startup` option**, which we firstly considered.
We could analyze each assembly in another tool, and as a previous step, and perform multiple invocations to `dotnet swagger --startup`.
In that case your Cli tool would provide a mechanism for generating OpenApi spec from a given startup, but **anyone willing to use it would need to re-implement a tool to analyze their code and extract their startup classes**.
We considered that having a **`--multiplestartups` option** integrated in your Cli, together with that 'startup discovery' mechanism, would add a huge value to it.
It seems a great advantage to us against having to invoke `dotnet swagger` n times and developing another separated tool to extract those classes (in cases similar to ours).

By the way, adding a `--startup` option would make sense for us, as a way of allowing other devs to use their own 'startup discovery' implementation, in case proposed one doesn't fit them.
We can add it together with this pull-request, or as a separate, future one.

### Implementation details
In case we reach an agreement regarding the approach, we would have no problem doing it in mentioned points:
* `--output` option can be made optional, invoking `SerializeSwagger()` with null `outputPath` in that case, as it's done in main flow.
* We can work in removing any duplicated logic in both Program's path, for sure.
* `SwaggerStartupAttribute` has been separated from CLI to be able to keep `Swashbuckle.AspNetCore.Cli` only as a `<DotNetCliToolReference>`, since it seemed more natural to us doing it that way and also thinking in any possible, future case where attributes would need to be referenced without the cli.
```
  <ItemGroup>
    <PackageReference Include="Swashbuckle.AspNetCore.SwaggerStartupAttr" />
    <DotNetCliToolReference Include="Swashbuckle.AspNetCore.Cli" />
    <!--...-->
  </ItemGroup>
```
However, if you feel it's better to include it in Cli project and we find a way to make it work that way, we won't go against that modification.

tburi:
Just to add my 2 cents, I think you could achieve the same functionality with my PR #887 
by moving the Startup discovery to your own project.

Basically the PR gives an alternative entry point for the CLI where you can build your IWebHost the way you want, by implementing ISwaggerWebHostFactory

```
 public interface ISwaggerWebHostFactory
 {
        IWebHost BuildWebHost(); 
 }
```

keycad:
I'm also working on a project where we are exposing two different api's, and it would be useful for us if Swashbuckle could provide some different-api's dynamic discovery. 
With regards to the approaches you've suggested, our project could fit both with Startup or WebHost.

domaindrivendev:
Closing as #887 has been merged to add custom WebHost discovery, and is aailable as of `v5.1.0`. Here's the relevent section of the docs - https://github.com/domaindrivendev/Swashbuckle.AspNetCore#use-the-cli-tool-with-a-custom-iwebhost-configuration

PR #960, opened 2018-12-06
Example from <example> xml comment use the correct type

### LINKED ISSUE #959: Example from <example> xml comment always has 'string' type
For the model:
```csharp
public class Model
{
    /// <summary>
    ///     The maximum number of elements in a page.
    /// </summary>
    /// <example>10</example>
    [JsonProperty("size")]
    public int Size { get; set; }
}
```
Swashbuckle.AspNetCore produces the following example:
```json
{
  "size": "10"
}
```

As you can see, the type for the `"size"` property is a `string`, not `int`. The correct example should be:
```json
{
  "size": 10
}
```

This bug can be reproduced in the latest stable version 4.0.1.

ar7z1:
In this PR I've fixed #959: for some types I've added convertation to appropriate OpenApi type.

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsSchemaFilter.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsSchemaFilter.cs
@@ -1,9 +1,10 @@
 ﻿using System;
+using System.ComponentModel;
 using System.Xml.XPath;
 using System.Reflection;
+using Microsoft.OpenApi.Any;
 using Newtonsoft.Json.Serialization;
 using Microsoft.OpenApi.Models;
-using Microsoft.OpenApi.Any;
 
 namespace Swashbuckle.AspNetCore.SwaggerGen
 {
@@ -69,46 +70,20 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
             }
         }
 
-        private static IOpenApiAny ConvertToOpenApiType(string value, Type type)
+        private static IOpenApiPrimitive ConvertToOpenApiType(string value, Type type)
         {
-            var convertedType = ConvertToType(value, type);
-            if (convertedType == null)
-                return new OpenApiString(value);
+            object typedExample;
 
-            switch (Type.GetTypeCode(type))
-            {
-                case TypeCode.Int32:
-                case TypeCode.Int16:
-                    return new OpenApiInteger((int) convertedType);
-                case TypeCode.Int64:
-                    return new OpenApiLong((long) convertedType);
-                case TypeCode.Double:
-                    return new OpenApiDouble((double) convertedType);
-                case TypeCode.Single:
-                    return new OpenApiDouble((double) new decimal((float) convertedType));
-                case TypeCode.Decimal:
-                    return new OpenApiDouble((double) (decimal) convertedType);
-                case TypeCode.Byte:
-                    return new OpenApiByte((byte) convertedType);
-                case TypeCode.Boolean:
-                    return new OpenApiBoolean((bool) convertedType);
-                case TypeCode.DateTime:
-                    return new OpenApiDate((DateTime) convertedType);
-                default:
-                    return new OpenApiString(value);
-            }
-        }
-
-        private static object ConvertToType(string value, Type type)
-        {
             try
             {
-                return Convert.ChangeType(value, type);
+                typedExample = TypeDescriptor.GetConverter(type).ConvertFrom(value);
             }
             catch (Exception)
             {
-                return null;
+                return new OpenApiString(value);
             }
+
+            return OpenApiPrimitiveFactory.CreateFrom(typedExample) ?? new OpenApiString(value);
         }
     }
 }
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsSchemaFilter.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsSchemaFilter.cs
@@ -1,4 +1,5 @@
-﻿using System.Xml.XPath;
+﻿using System;
+using System.Xml.XPath;
 using System.Reflection;
 using Newtonsoft.Json.Serialization;
 using Microsoft.OpenApi.Models;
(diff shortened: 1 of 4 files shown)


test/Swashbuckle.AspNetCore.SwaggerGen.Test/Swashbuckle.AspNetCore.SwaggerGen.Test.csproj
```diff
@@ -20,6 +20,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <PackageReference Include="FluentAssertions" Version="5.5.3" />
```
ar7z1: I've added [FluentAssertions package](https://www.nuget.org/packages/FluentAssertions/) to easily compare OpenApi objects. Another way was to write very complicated code to check if OpenApi objects are equivalent.

ar7z1:
@domaindrivendev I've backported this bugfix for 4.0.1 in [69a06e03a051](https://github.com/ar7z1/Swashbuckle.AspNetCore/commit/69a06e03a0512f9adc8a6aba099dffeeaf79b9a1). 
If this PR is ok and if it is not very difficult, it would be great to see these changes in version 4.0.2.

src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsSchemaFilter.cs
```diff
@@ -61,7 +62,53 @@ private void ApplyPropertyComments(OpenApiSchema propertySchema, MemberInfo memb
 
             var exampleNode = memberNode.SelectSingleNode(ExampleXPath);
             if (exampleNode != null)
-                propertySchema.Example = new OpenApiString(XmlCommentsTextHelper.Humanize(exampleNode.InnerXml));
+            {
+                var exampleString = XmlCommentsTextHelper.Humanize(exampleNode.InnerXml);
+                var memberType = (memberInfo.MemberType & MemberTypes.Field) != 0 ? ((FieldInfo) memberInfo).FieldType : ((PropertyInfo) memberInfo).PropertyType;
+                propertySchema.Example = ConvertToOpenApiType(exampleString, memberType);
+            }
+        }
```
domaindrivendev: There's already a class called `OpenApiPrimitiveFactory` that you should be able to re-use in place of the `ConvertToOpenApiType` method. I _think_ something along the following lines will work:

```csharp
var typedExample = TypeDescriptor.GetConverter(memberType).ConvertFrom(exampleString);
propertySchema.Example = OpenApiPrimitiveFactory.CreateFrom(typedExample)
```

ar7z1: @domaindrivendev Great catch! Fixed in 7a5ebf7.

test/Swashbuckle.AspNetCore.SwaggerGen.Test/XmlComments/XmlCommentsSchemaFilterTests.cs
```diff
@@ -55,14 +57,29 @@ public class XmlCommentsSchemaFilterTests
...
+                new object[] {typeof(XmlAnnotatedType), "BoolField", new OpenApiBoolean(true)},
+                new object[] {typeof(XmlAnnotatedType), "GuidProperty", new OpenApiString("d3966535-2637-48fa-b911-e3c27405ee09")},
+                new object[] {typeof(XmlAnnotatedType), "BadExampleProperty", new OpenApiString("property bad example")},
+            };
+
         [Theory]
-        [InlineData(typeof(XmlAnnotatedType), "Property", "property example")]
-        [InlineData(typeof(XmlAnnotatedType), "Field", "field example")]
-        [InlineData(typeof(XmlAnnotatedSubType), "Property", "property example")]
+        [MemberData(nameof(PropertyExampleTestData))]
```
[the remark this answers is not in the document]
ar7z1: `openApiPrimitive.ToString()` - is default object `ToString`, it returns type (`Microsoft.OpenApi.Any.OpenApiString` for example). Added in 7a5ebf7 a simple method to get `Value` via reflection and removed FluentAssertions.

test/Swashbuckle.AspNetCore.SwaggerGen.Test/XmlComments/XmlCommentsSchemaFilterTests.cs
```diff
@@ -75,8 +84,14 @@ public class XmlCommentsSchemaFilterTests
 
             Subject().Apply(schema, filterContext);
 
-            Assert.IsType<OpenApiString>(schema.Properties[propertyName].Example);
-            Assert.Equal(expectedExample, ((OpenApiString)schema.Properties[propertyName].Example).Value);
+            var openApiPrimitive = (IOpenApiPrimitive)schema.Properties[propertyName].Example;
+            Assert.Equal(expectedPrimitiveType, openApiPrimitive.PrimitiveType);
+            Assert.Equal(expectedValueString, GetOpenApiPrimitiveValue(openApiPrimitive).ToString());
+        }
```
domaindrivendev: Thanks for the quick turnaround. One last change before I merge. Prior to the assertions, you can cast `Example` to the concrete `OpenApiPrimitive` instead of the interface. The concrete type has a `Value` property that you can use instead of the reflection hack. For example:

```csharp
var openApiPrimitive = (OpenApiPrimitive)schema.Properties[propertyName].Example;
Assert.Equal(expectedValueString, openApiPrimitive.Value.ToString());
```

ar7z1: Thank you for the review!
I would like to cast to `OpenApiPrimitive`, but I can't: `OpenApiPrimitive` is generic type (https://github.com/Microsoft/OpenAPI.NET/blob/master/src/Microsoft.OpenApi/Any/OpenApiPrimitive.cs#L14). :-(
That's why I had to add this ugly reflection hack.

domaindrivendev:
Thanks for the contribution.

ar7z1:
Thanks again for the review and merge!
Is there any chance to get the backported changes (from commit [cb795f69](https://github.com/ar7z1/Swashbuckle.AspNetCore/commit/cb795f69bba704ad7a4901cef42442e9ec5cd9df)) in 4.0.2?

MikeBeaton:
> @domaindrivendev I've backported this bugfix for 4.0.1 in [cb795f69bba70](https://github.com/ar7z1/Swashbuckle.AspNetCore/commit/cb795f69bba704ad7a4901cef42442e9ec5cd9df).
> If this PR is ok and if it is not very difficult, it would be great to see these changes in version 4.0.2.

@ar7z1 I've put up [a gist](https://gist.github.com/MikeBeaton/ea9e47109f55fbaec0113756904b900e) containing your patched file (plus one `internal` extension method from the [Swashbuckle source](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/master/src/Swashbuckle.AspNetCore.SwaggerGen/Generator/JsonPropertyExtensions.cs) which it needs) and documented in #1021 how to apply this in a live project without waiting for a new release.

PR #1001, opened 2019-01-06
ReDoc 2.0.0-rc.1

emilssonn:
- Updated ReDoc to version 2.0.0-rc.1. This enables OpenAPI 3.0 support.
- Added documentation in README-v5 for ReDoc options.
- Removed ReDoc options that are not in 2.0.0-rc.1. (While doing this I also noticed why the "SupressWarnings" do not work in pre version 5 Swashbuckle, it should be "SuppressWarnings".)
- Added extension methods for ReDoc options.

Since ReDoc is not yet in stable 2.0 this PR should (if considered ok) be merged before Swashbuckle 5.0 is released (since ReDoc 1.x does not work with OpenAPI 3) or when ReDoc 2.0 is released.

### DIFF
--- a/README-v5.md
+++ b/README-v5.md
@@ -19,7 +19,7 @@ Once you have an API that can describe itself in Swagger, you've opened the trea
 
 |Swashbuckle Version|ASP.NET Core|Swagger / OpenAPI Spec.|swagger-ui|ReDoc UI|
 |----------|----------|----------|----------|----------|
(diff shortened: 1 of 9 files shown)


vankooch:
I've merged your branch into my fork, you can use this for testing
```https://www.myget.org/F/kiarash-dev/api/v3/index.json```

PR #1019, opened 2019-01-25
[Bugfix] Ignore missing properties when mapping metadata

PavelZykan:
This code:
```
var jsonProperty = jsonObjectContract.Properties[entry.Key];
if (jsonProperty == null) continue;
```
will fail with at the first line with an exception when entry.Key is not present in the dictionary. The check should be done beforehand. I added a rudimentary test that fails without this change, but passes after the change is introduced. Thanks.


PavelZykan:
Hi @domaindrivendev,
is there anything I can do to have this PR merged? Thanks.

jimcomiskey:
Hi @domaindrivendev is there an ETA on when Swashbuckle.ASpNetCore 5.0.1 RC will become an Official? Just wondering since it will contain this fix. For now I'm using 4.0.0.  Thanks.

Issue #1027, opened 2019-02-01
Write only (`set` only) properties are not being tagged as such in Swashbuckle models


MikeBeaton:
`set`-only properties appear in the generated model output, even though these are not serialized by the .NET Core Web API.

For example with this class:

```
public class TestClass
{
    // none of these appear in the swagger output - correct
    private int UnmaskedPrivateField;
    private int UnmaskedPrivateGetProperty { get; }
    private int UnmaskedPrivateGetSetProperty { get; set; }
    private int UnmaskedPrivateSetProperty { set { } }

    // all of these appear in the swagger output - the last one should not
    public int UnmaskedPublicField;
    public int UnmaskedPublicGetProperty { get; }
    public int UnmaskedPublicGetSetProperty { get; set; }
    public int UnmaskedPublicSetProperty { set { } }

    // all of these appear in the swagger output - the last one should not
    [JsonProperty("Json_MaskedPublicField")] public int MaskedPublicField;
    [JsonProperty("Json_MaskedPublicGetProperty")] public int MaskedPublicGetProperty { get; }
    [JsonProperty("Json_MaskedPublicGetSetProperty")] public int MaskedPublicGetSetProperty { get; set; }
    [JsonProperty("Json_MaskedPublicSetProperty")] public int MaskedPublicSetProperty { set { } }

    // all of these appear in the swagger output - the last one should not
    [JsonProperty("Json_MaskedPrivateField")] private int MaskedPrivateField;
    [JsonProperty("Json_MaskedPrivateGetProperty")] private int MaskedPrivateGetProperty { get; }
    [JsonProperty("Json_MaskedPrivateGetSetProperty")] private int MaskedPrivateGetSetProperty { get; set; }
    [JsonProperty("Json_MaskedPrivateSetProperty")] private int MaskedPrivateSetProperty { set { } }
}
```

and this controller method:

```
[HttpGet]
public ActionResult<TestClass> GetTest()
{
    return new TestClass();
}
```

then the actual output of a call to the endpoint is:

```
{
  "unmaskedPublicField": 0,
  "Json_MaskedPublicField": 0,
  "Json_MaskedPrivateField": 0,
  "unmaskedPublicGetProperty": 0,
  "unmaskedPublicGetSetProperty": 0,
  "Json_MaskedPublicGetProperty": 0,
  "Json_MaskedPublicGetSetProperty": 0,
  "Json_MaskedPrivateGetProperty": 0,
[...]

PR #1036, opened 2019-02-06
Fix Swashbuckle not tagging write-only properties

MikeBeaton:
Fix, with tests, for #1027.

Note that the API test for `ComplexType` is now returning six properties (i.e. ignoring the new write-only property) automatically; it was only Swashbuckle that needed to be changed to also ignore it.

I hope I've put the check for non-readable properties in a sensible place, but let me know if not!

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/Generator/SchemaRegistry.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/Generator/SchemaRegistry.cs
@@ -231,6 +231,7 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
         private OpenApiSchema CreateObjectSchema(JsonObjectContract jsonContract, Queue<Type> referencedTypes)
         {
             var applicableJsonProperties = jsonContract.Properties
+                .Where(prop => prop.Readable)
                 .Where(prop => !prop.Ignored)
                 .Where(prop => !(_options.IgnoreObsoleteProperties && prop.IsObsolete()))
                 .Select(prop => prop);
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/Generator/SchemaRegistry.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/Generator/SchemaRegistry.cs
@@ -231,7 +231,6 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
         private OpenApiSchema CreateObjectSchema(JsonObjectContract jsonContract, Queue<Type> referencedTypes)
         {
             var applicableJsonProperties = jsonContract.Properties
-                .Where(prop => prop.Readable)
                 .Where(prop => !prop.Ignored)
                 .Where(prop => !(_options.IgnoreObsoleteProperties && prop.IsObsolete()))
                 .Select(prop => prop);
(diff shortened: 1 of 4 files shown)


MikeBeaton:
PS See my first comment on #1027 for why anybody might so much as want write-only properties.

Though hopefully the fact that adding a write-only property to the `ComplexType` test class immediately created a mismatch between what Swashbuckle says the API will output and what it actually outputs is a already good enough reason to include the fix.

MikeBeaton:
I think there was no test anywhere for `ReadOnly` being populated correctly(?). There was certainly no test for it being `false` for my new read-only test property (of course!). So I've just added these to the PR.

domaindrivendev:
It makes sense to ignore `set-only` members if the type is ONLY being used for a response. However, types may be used for requests too (i.e. deserialization), and in many cases are used for both. By automatically ignoring ALL `set-only` operations you're introducing unexpected behavior for those cases.

MikeBeaton:
Dear @domaindrivendev ,

Before I start, thank you for an extremely useful, and usable, project that you have written!

1. I must be missing something, somewhere, but if a property on a given type is literally set-only (everywhere) then (surely?!) it wouldn't matter where instances of the type are sent, or why, because the receiver simply cannot read from that property ever? (Therefore, of course, the only point of set-only properties is when the setter has some side effect - maybe setting another readable field. Which is why an auto-implemented set-only property, with no body on the setter, is not allowed by the compiler.) I'm sure I must be missing a use-case here, but I haven't got my head round it yet.

1. So, I guess, first thing to ask is, though I'm asking it second, is, do you agree that there is a problem here, at least? For example, if I simply add `public string Property6 { set { } }` to `Swashbuckle.AspNetCore.SwaggerGen.Test.ComplexType` then Swashbuckle starts to report this as a model with 6 members, but this is simply not a true reflection of what the API returns (which still has 5 members)

1. As it stands (I guess it should go without saying, but anyway...!) my existing PR doesn't break any existing tests.... so if you could provide a test, or just suggest a use-case in a little more detail, that lets me get my head round when and why set-only *should* be included, then I could try to code to that (with an extra test for the new use-case if needed). Would that help?

Thanks,

Mike

MikeBeaton:
Oh - I think I have got it. You're saying that it could make sense to deserialize from incoming JSON which includes a value for a set-only property. I've just seen what you mean. As a receiver of on object deserialized from that, I wouldn't be able to read the value of the set-only property, but I would be able to read its side-effects. If you can answer quickly, without trawling through everything, is there actually an existing test showing that?

Of the top of my head, it might make sense to just filter set-only properties out of output models, in that case - but then I'm guessing one problem is that a given model could be both input and output.

MikeBeaton:
Okay, in this revised version I'm no longer removing those properties, just setting `writeOnly` on them, now that I've understood (I think!) that that exists (existed already, on `OpenApiSchema`) but needed to be set. I've put a test in to determine that `readOnly` and `writeOnly` *are* getting set correctly, and then I've just tried this modified version of Swashbuckle with my own project and it works as expected now: now that the model has `writeOnly` set when applicable, the [swagger-ui](https://github.com/swagger-api/swagger-ui) (I guess it is?) is correctly determining the properties to include in its example values.

domaindrivendev:
Yep we're 100% on the same page now. Thanks for baring with me. Marking the corresponding schema as `writeOnly` is exactly what needs to happen. I wasn't even aware that was part of the OpenAPI specification so great spot and find on your part.

PR #1057, opened 2019-03-05
Improve DataTypeAttribute support

vankooch:
I've added some more mapping in ```ObjectSchemaGenerator.cs```. This resources are used for updating the mappings:

* [OpenAPI Spec](https://swagger.io/docs/specification/data-models/data-types/)
* [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input) 
* [DataTypeAttribute](https://docs.microsoft.com/en-us/dotnet/api/system.componentmodel.dataannotations.datatypeattribute?view=netcore-2.2)
* [EmailAttribute and co.](https://docs.microsoft.com/en-us/dotnet/api/system.componentmodel.dataannotations.emailaddressattribute?view=netcore-2.2)


domaindrivendev:
@vankooch these look like great additions. Could you rebase on top of latest master?

vankooch:
@domaindrivendev I've updated from upstream, should be good to go :)

domaindrivendev:
👍

PR #1071, opened 2019-03-11
failing type conversion of short enums

basman:
An enum of type short leads to a casting error. In order to cast a short enum to int, one must unbox first.
If you wish to fix it in a different way, just throw this pull request away.

https://stackoverflow.com/questions/7648871/directly-unboxing-a-boxed-int-to-short

I wanted to quickly address the issue in test Apply_SetsPropertyExample_FromPropertyExampleTags, but failed to get the test right in a first attempt. Maybe you have an idea how to construct a test case.


domaindrivendev:
FYI - give the factory is used in a couple of different contexts, I thought it should have it's own set of tests, including the scenario you identified - see 9feec76946b90c60566f3037545d13c9ed9e8347

PR #1078, opened 2019-03-15
Introduce integration testing / spec-first libs with initial support for xunit

domaindrivendev:
**Problem Statement**

With Swagger/OpenAPI, two schools of thought have evolved around how to incorporate it into API development (and documentation) workflows. With the **Code-first** approach, you implement the API first and generate the specification document from the code. **Design-first** on the other hand, advocates for designing the API contract first before writing any code. Both approaches have their advantages and disadvantages, which you can read about [here](https://swagger.io/blog/api-design/design-first-or-code-first-api-development/).

The current set of Swashbuckle tools fit firmly into the Code-first camp and so they're not particularly useful for API developers that want to adopt a Design-first approach when building API's.

**Solution**

Existing tooling for the Design-first space tends to focus on code generation. While this approach works for a lot of people, I've had little success over the years with (application) code generation techniques (OpenAPI-based or otherwise) and so I'm not compelled to go down this road. However, as somewhat of an enthusiast for BDD & TDD, it seemed very natural to leverage an OpenAPI document as an _executable specification_. That is, you describe your desired API behavior in the form of an OpenAPI document and then, with additional tooling (✋), use that specification to test your API - paths, parameters, response codes, JSON structures, the works!

This PR introduces two new packages that enable exactly that. Rather than re-inventing the wheel, it builds on the [integration testing tools already provided for ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/test/integration-tests?view=aspnetcore-2.2). While this PR includes support for `Xunit` only, the core components are decoupled from the testing framework in use and so adding support for others should be relatively trivial.


domaindrivendev:
@darrelmiller - I know you're an advocate for **Design-first**, interested to get your thoughts?

darrelmiller:
@domaindrivendev  I think that's a great idea.  There is a Java/Kotlin tool that does a similar thing https://github.com/codecentric/hikaku

Issue #1085, opened 2019-03-21
Use of TimeSpan causes KeyNotFoundException 


Jan-Magerl:
If i use the Type TimeSpan in my Rest interface i get an Exception:

you use :

ContractResolver.ResolveContract(type) is JsonPrimitiveContract;

in Json TimeSpan is PrimitiveType: https://github.com/JamesNK/Newtonsoft.Json/blob/master/Src/Newtonsoft.Json/Utilities/ConvertUtils.cs#L49

In your FactoryMethodMap TimeSpan is missing.

so this Code crashes:

            var schema = jsonPrimitiveContract.UnderlyingType.IsEnum
                ? GenerateEnumSchema(jsonPrimitiveContract)
                : FactoryMethodMap[jsonPrimitiveContract.UnderlyingType]();
https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/master/src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGen/PrimitiveSchemaGenerator.cs#L34

System.Collections.Generic.KeyNotFoundException: The given key was not present in the dictionary.
   at System.ThrowHelper.ThrowKeyNotFoundException()
   at System.Collections.Generic.Dictionary`2.get_Item(TKey key)
   at Swashbuckle.AspNetCore.SwaggerGen.PrimitiveSchemaGenerator.GenerateSchemaFor(Type type, SchemaRepository schemaRepository)
   at Swashbuckle.AspNetCore.SwaggerGen.ChainableSchemaGenerator.GenerateSchema(Type type, SchemaRepository schemaRepository)
   at Swashbuckle.AspNetCore.SwaggerGen.ChainableSchemaGenerator.GenerateSchema(Type type, SchemaRepository schemaRepository)
   at Swashbuckle.AspNetCore.SwaggerGen.ChainableSchemaGenerator.GenerateSchema(Type type, SchemaRepository schemaRepository)
   at Swashbuckle.AspNetCore.SwaggerGen.ChainableSchemaGenerator.GenerateSchema(Type type, SchemaRepository schemaRepository)
   at Swashbuckle.AspNetCore.SwaggerGen.ChainableSchemaGenerator.GenerateSchema(Type type, SchemaRepository schemaRepository)
   at Swashbuckle.AspNetCore.SwaggerGen.SchemaGenerator.GenerateSchema(Type type, SchemaRepository schemaRepository)
[...]

PR #1086, opened 2019-03-21
added TimeSpan to FactoryMethodMap

Jan-Magerl:
#1085


jobubo:
pull request belongs to #1086

fawdlstty:
It may not be a good idea to just classify it as a string

Issue #1064, opened 2019-03-07
JsonObject(ItemRequired = Required.AllowNull) not recognized


artfulsage:
Hi. I suppose there is a bug in recognizing required properties of class marked with `[JsonObject(ItemRequired = Required.AllowNull)]` while properties marked with `[JsonRequiredAttribute]` and `[JsonProperty(Required = Required.AllowNull)]` recognized well.

Repro sample:
```cs
public class Program
{
	public static void Main(string[] args)
		=> WebHost.CreateDefaultBuilder(args).UseStartup<Startup>().Build().Run();
}

public class Startup
{
	public Startup(IConfiguration configuration) => Configuration = configuration;

	public IConfiguration Configuration { get; }

	public void ConfigureServices(IServiceCollection services)
	{
		services.AddMvc().SetCompatibilityVersion(CompatibilityVersion.Version_2_2);
		services.AddSwaggerGen(c => c.SwaggerDoc("1.0", new Info()));
	}

	public void Configure(IApplicationBuilder app, IHostingEnvironment env)
		=> app.UseSwagger()
			.UseSwaggerUI(c =>
			{
				c.SwaggerEndpoint($"/swagger/1.0/swagger.json", "API");
				c.RoutePrefix = string.Empty;
			})
			.UseMvc();
}

[JsonObject(ItemRequired = Required.AllowNull)]
public class Item
{
	public int? Id { get; set; }
	public string Name { get; set; }
}

public class Item2
{
	[JsonRequired]
	public int? Id { get; set; }

	[JsonProperty(Required = Required.AllowNull)]
	public string Name { get; set; }
}

[Route("api/[controller]")]
[ApiController]
public class ItemsController : ControllerBase
{
	[HttpGet("default-item")]
	public ActionResult<Item> GetDefaultItem() => new Item();

	[HttpGet("default-item2")]
	public ActionResult<Item2> GetDefaultItem2() => new Item2();
}
```
![swagger problem](https://user-images.githubusercontent.com/35620830/53957315-15495680-40ef-11e9-8974-9a649f553d29.png)

slahabar:
I am running into issues with below. 
 [JsonProperty(Required = Required.AllowNull)] 
tags it as required field but only allows non-null values.

spfaeffli:
Currently the value for `JsonProperty.Required` only determines if the value is required - it does not allow you to indicate that a value may or may not be null.
Example:
```
[...]

PR #1119, opened 2019-04-10
added support for JsonObject.Required and JsonProperty.Required

spfaeffli:
The json `Required` value is now correctly determined for json annotated types and used to set `Required` and `Nullable` in the `OpenApiSchema`. The PR resolves the issue #1064.
To achieve the same behavior as Newtonsoft Json, I had to use some reflection. Let me know what you think.

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGen/MemberInfoExtensions.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGen/MemberInfoExtensions.cs
@@ -1,0 +1,29 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+
+namespace Swashbuckle.AspNetCore.SwaggerGen
+{
+    public static class MemberInfoExtensions
+    {
+        /// <summary>
+        /// Gets custom attribute data.
+        /// </summary>
+        /// <param name="memberInfo">Member info.</param>
+        /// <param name="attributeType">Attribute type.</param>
+        /// <param name="inherit">Search inheritance chain.</param>
+        /// <returns>Custom attribute data</returns>
+        public static CustomAttributeData GetCustomAttributeData(this MemberInfo memberInfo, Type attributeType, bool inherit)
+        {
+            var customAttrData = memberInfo.CustomAttributes.FirstOrDefault(a => a.AttributeType == attributeType);
+            if (customAttrData == null && inherit)
+            {
+                var baseType = memberInfo.DeclaringType.BaseType;
+                var baseMemberInfo = baseType?.GetMember(memberInfo.Name).FirstOrDefault();
+                customAttrData = baseMemberInfo?.GetCustomAttributeData(attributeType, inherit);
+            }
+            return customAttrData;
+        }
+    }
+}
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGen/ObjectSchemaGenerator.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGen/ObjectSchemaGenerator.cs
@@ -42,11 +42,11 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
 
                 if (Options.IgnoreObsoleteProperties && attributes.OfType<ObsoleteAttribute>().Any()) continue;
 
-                properties.Add(jsonProperty.PropertyName, GeneratePropertySchema(jsonProperty, memberInfo, attributes, schemaRepository));
+                var required = GetRequiredValue(jsonProperty, jsonObjectContract, memberInfo);
+                properties.Add(jsonProperty.PropertyName, GeneratePropertySchema(jsonProperty, required, memberInfo, attributes, schemaRepository));
 
-                if (jsonProperty.Required == Required.AllowNull || jsonProperty.Required == Required.Always || attributes.OfType<RequiredAttribute>().Any())
+                if (required == Required.AllowNull || required == Required.Always || attributes.OfType<RequiredAttribute>().Any())
                     requiredPropertyNames.Add(jsonProperty.PropertyName);
-
             }
 
             var additionalProperties = (jsonObjectContract.ExtensionDataValueType != null)
@@ -67,6 +67,7 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
 
         private OpenApiSchema GeneratePropertySchema(
             JsonProperty jsonProperty,
+            Required? required,
             MemberInfo memberInfo,
             object[] attributes,
             SchemaRepository schemaRepository) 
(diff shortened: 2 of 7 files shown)


domaindrivendev:
Thanks for putting this together. On first glance it seems to add significant complexity to the code.

I'll need a little more time to pull it down, walkthrough it and fully understand what's going on - my sense is that there should be an easier but I could be wrong

spfaeffli:
Yeah, I thought you might want to take a closer look.
The added complexity is necessary to determine what the `Required` value for a `JsonProperty` actually is. It is only necessary for `Required.Default`, since it could mean one of the following:

- The property is decorated with `[JsonProperty(Required = Required.Default)]`.
- The property is decorated with `[JsonProperty]`.
- The property is not decorated on the reflected type but somewhere in the inheritance chain.
- The property is not decorated anywhere.

If the `Required` value is not explicitly specified (no attribute nor a value in ctor.), the value `ItemRequired` of the `JsonObject` attribute is used (which is nullable, so we know if it was specified or not).

As far as I can tell, this is how Newtonsoft Json does it and I don't think there is a way to mirror this behavior without something like this. However if you find a better way to do it, that would be great - I had some doubts about this myself.

Let me know if you need something else.

spfaeffli:
Any news on when you'll be able to merge this?

It would also be helpful if you could create a new RC after the merge.
My team generates clients based on the created api specification.
This leads to some issues, since the current version does not generate the desired api specification.

domaindrivendev:
I've decided to close this PR … not because it's not a compelling use case but because I'm planning a significant refactor that will decouple the core Swashbuckle components from the `Newtonsoft` library.

spfaeffli:
@domaindrivendev Are you planning on supporting this use-case or will Newtonsoft-Attributes not have any effect on how the how the schema is generated? Will there be another way to specify required/non-required fields?

domaindrivendev:
The majority of Newtonsoft attributes _are supported today_, including `[JsonProperty(Required = Required.AllowNull)]`. In your comments above, your asking about flagging required/non-required properties fields and this is easily done using the `RequiredAttribute` as described in the [readme](https://github.com/domaindrivendev/Swashbuckle.AspNetCore#flag-required-parameters-and-schema-properties).

Is that what you want to do or am I misunderstanding the requirement?

domaindrivendev:
@spfaeffli to clarify - the "refactor" I described will still support Newtonsoft attributes but in a way that's decoupled from the Core generator. This will make it easier to support other JSON serialization libraries, most notably the native [JSON support](https://github.com/dotnet/corefx/blob/master/src/System.Text.Json/docs/SerializerProgrammingModel.md) being introduced with .NET Core 3.0

spfaeffli:
@domaindrivendev The current version of the schema generator has multiple "issues". Nullable is not correctly set for properties - take a look at the example I provided in the referenced issue #1064. Also the value of the `Required` attribute should be determined differently (as I tried to explain in my previous [post](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/pull/1119#issuecomment-482601315)). The PR also includes various unit tests that would fail without the changes in the generator.

spfaeffli:
@domaindrivendev I am aware that the current version _works_ with the attributes you mentioned - just not the same way as Newtonsoft does. The values for required and nullable in the generated schema are different than Newtonsoft handles de-/serialization. For example a property could be marked as non-nullable and required in the schema, when in reality the API (with the Newtonsoft serializer) will not require that value. This PR fixes the issue so that the schema matches what the API actually requires. The changes are solely to support Newtonsoft attributes, so it would probably make sense to decouple them from the core implementation.

domaindrivendev:
@spfaeffli I wanted to follow up on this because I've done most of the work now to decouple the Schema generator from Newtonsoft (or at least to better separate the Newtonsoft specifics). And, with that done, I was able to revisit the behavior around Newtonsoft "Required", very much inspired by your work in this PR.

As you had pointed out, reflecting the Newtonsoft behavior is fairly nuanced but, based on your solution, I _think_ I have it right. Anyway, if you get a chance, it would be great if you could pull down the latest preview from myget and test it out.

Here's the related commit - c89876fbe77f65d25b1e768361a71ba7b9ef4462

spfaeffli:
@domaindrivendev thank you for your work on this.

Today I finally found some time to run some tests with the new preview version.
For the most part it works as expected - however there is still an issue regarding inherited "Required"-values. Consider the following example:

```
[JsonObject(ItemRequired = Required.AllowNull)]
public abstract class JsonRequiredBaseType
{
    public virtual string BaseValAllowNull { get; set; }
}
public class JsonRequiredChildType : JsonRequiredBaseType
{
    public string ValAllowNull { get; set; }

    // Property is handled as Required.Default even though it is not explicitly specified.
    // Required = Required.AllowNull should be inherited from JsonObjectAttribute on base class.
    [JsonProperty]
    public string ValAllowNullIsDefault { get; set; }

    [JsonProperty(Required = Required.Default)]
    public string ValDefault { get; set; }
}
```
This will produce the following model definition:
```
"JsonRequiredChildType": {
  "required": [
    "baseValAllowNull",
    "valAllowNull"
  ],
  "type": "object",
  "properties": {
    "valAllowNull": {
      "type": "string",
      "nullable": true
    },
    "valAllowNullIsDefault": {
      "type": "string",
      "nullable": true
    },
    "valDefault": {
      "type": "string",
      "nullable": true
    },
    "baseValAllowNull": {
      "type": "string",
      "nullable": true
    }
  },
  "additionalProperties": false
}
```
As you can see, the property `ValAllowNullIsDefault` is handled as if the value for `Required` was `Required.Default`.

If you take a look again at my [implementation](https://github.com/inova-solutions/Swashbuckle.AspNetCore/blob/ecbe9ab797e0eb2096dd5d3e3ff3def947460e14/src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGen/ObjectSchemaGenerator.cs#L159) of `GetRequiredValue`, you can see that I specifically handled that scenario.

I am aware that this is an issue that most people probably won't encounter (neither will I) - but since we know about it, we might as well fix it.

Hope this helps.

domaindrivendev:
@spfaeffli - thanks for pointing this out. Just so we're on the same page, I don't actually think the inheritance from a base type is the relevant part of the repro. It's really that the property in question is decorated with a `JsonProperty` attribute but the `Required` property on that attribute has not been explicitly provided. Therefore, in this case, SB should honor the class level setting, whether that's on the declaring type directly OR a parent of the declaring type.

This stuff is all a bit nuanced so just wanted to make that clarification to ensure I'm not missing anything.

Anyway, I think this should be solved with 72c0fc391e87d850f2363bb972605b6812c7eae9.

spfaeffli:
@domaindrivendev if you didn't care about inherited attributes, you could just check the values on the provided `JsonObjectContract` object. That is why I specifically mentioned the inheritance issue - in my code I first checked if it was actually necessary to read attributes from base types. I haven't run any performance tests, so I am not really sure if that was actually worth it. You are always checking it, so you don't have to worry about that.

I haven't tested your solution but I think it should work - thanks for the update.

spfaeffli:
@domaindrivendev: There is still a minor issue regarding "nullable" properties.
The current version does not allow value types to be marked as [nullable](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/ece92f826527d532bb1e3312ba683d3ee4807c8b/src/Swashbuckle.AspNetCore.SwaggerGen/Newtonsoft/NewtonsoftApiModelResolver.cs#L165) - unless they are `Nullable<T>`. I would prefer if Swashbuckle would not restrict this. Newtonsoft provides ways to deal with null-values during serialization (ignore null values or provide your own converter to handle them). The new Json API's will probably work in a similar fashion.

PR #1124, opened 2019-04-12
Add description for usage of global tool

kant2002:
As per @domaindrivendev description, created explanation for the usage of global tool.

I take liberation and place explicit version of 5.0.0 in the CliExample, since this is required to have proper tool configuration for restoring. So if you want to release 5.0.0-rc3 then that place should be changed.


domaindrivendev:
@kant2002 I moved this to the v5 readme as I want to reserve the GitHub home page docs for the latest stable version. See 2b171c0fc6efe40f3a29a45c48d6b01221a0c214

kant2002:
That totally make sense. @domaindrivendev would you mind if just tell me change location of the documentation. That give me a bit ego boost 😄 since that way you could preserve authorship in PR.

Anyway, more interesting for me is when could I expect release to happens? I actually need that change for my CI infrastructure. If it would be in relatively short time, then I could wait without messing with private Nuget

PR #1139, opened 2019-05-02
Fix/enum underlying type casting

rsibanez89:
Previously, this enum would be converted as null values
```
[Flags]
public enum Gender : byte
{
	None = 0,
	Male = 1,
	Female = 2,
	Both = Male | Female
}
```

```
"Gender": {
	"enum": [
		null,
		null,
		null,
		null
	],
	"type": "integer",
	"format": "int32"
}
```

This is because we are missing the casting to a byte


rsibanez89:
@domaindrivendev, can you please have a look at this PR.
We are blocked by the null values being generated by the tool.

Thanks.

MikeBeaton:
@rsibanez89 I noticed that you and @epsitec are making similar changes, about bytes and shorts respectively! I looked at (and offered to do) what I think would be needed to combine them in #1161 .

rsibanez89:
@domaindrivendev can you have a look at this PR?

rsibanez89:
Ping @domaindrivendev

rsibanez89:
@domaindrivendev seems that you have been busy lately. Are you still maintaining this repo?

PR #1145, opened 2019-05-04
Fix incorrect OpenApiSchema Example for floating-point values when CultureInfo is different from "en-US"

retinio:
When CultureInfo is different from "en-US", incorrect OpenApiSchema is created.
For sample:
My cultureInfo is "uk-UA" and for DoubleProperty I get 

image


nilshedstrom:
This seems to fix the same thing as #1379 which already has been merged to master

Issue #1154, opened 2019-05-12
Supporting static swagger documents


kimbell:
We have been using Swashbuckle for some time now on ASP.NET Core. 
All our internal traffic is routed through a service proxy that is responsible for managing a path prefix for each service. The path is passed to the server using a header. 

In our service we have been using PreSerializeFilters to dynamically change the document's BasePath based on the presence of this header. This allows the Swagger UI to function both on localhost and via the service proxy. 

Not long ago, our architects started pushing for contract first and having a version of the contract checked into source control. One of the reasons is to avoid unintended changes to the contract; something that can easily happen when generated at run-time. 

The way we set things up now is:
- Add static file middleware
- place the Swagger contract on disk
- Don't register it with SwaggerGen
- Register it with SwaggerUI

This allows us to toggle between the static and run-time generated versions. 

The challenge comes in relation to the header our service proxy adds. Since it's a static file, the PreSerializeFilters are not run. We need to decide up front if it should work via service proxy or localhost. 

The best approach would be if Swashbuckle supported adding static swagger documents that get run through all the logic as the run-time generated documents, except those that depend on ApiExplorer.

domaindrivendev:
Haven’t had a chance to look at proposed solution just yet but off the top of my head - you could just create and register a custom implementation of ISwaggerProvider that simply deserializes the document from disk. That way, your PreSerializeFilters will work as the did before

kimbell:
I had a  look at the ISwaggerProvider and think I see a way of getting both static documents and run-time generated documents to work at the same time. I'll see if I can try things out during the coming week.

cecilphillip:
@kimbell let me know how that works out. I wanna try it out with https://github.com/microsoft/OpenAPI.NET.OData

kimbell:
I think I have it working for all the cases I need. This is what I had to do.

In order to make the settings available in my custom implementation, I set up the SwaggerOptions in DI. 
A lot of this code is based around the header values our service proxy uses. You may want different logic depending on what you use. 
```cs
// configure options so that they can be used by StaticSwaggerProvider
            services.Configure<SwaggerOptions>(o =>
            {
                o.RouteTemplate = "docs/swagger/{documentName}/swagger.json";
                o.SerializeAsV2 = version == OpenApiVersion.V2;
                o.PreSerializeFilters.Add((doc, request) =>
                {
[...]

PR #1155, opened 2019-05-12
proposed solution for supporting static swagger documents

kimbell:
This is my proposed solution for issue #1154

When building locally, I had problems with an embedded js file in the ReDoc projects. Had to remove it to get things to compile.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ artifacts/
 .DS_Store
 Thumbs.db
 test/WebSites/CliExample/wwwroot/api-docs/v1/*.json
+_ReSharper*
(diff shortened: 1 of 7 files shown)


EspressoBeans:
@kimbell, only curious, what build issues did you run into specifically?  

For your proposed solution, it removes the cs.proj embedded ReDoc link correct? :
`<script src="https://cdn.jsdelivr.net/npm/redoc/bundles/redoc.standalone.js"></script>`
So if I'm using ReDoc how would I add it back in?

I'm using ReDoc and need a way to specify the CDN js file, instead of Swashbuckle.AspNetCore doing it for me, and am hoping your proposed solution would help me.

kimbell:
The build issue I had was related to the embedded ReDoc file. It was not part of source control. I don't know if there is some other build system that magically places it there. The reason I mentioned it was to inform the maintainer about something that I did that would require some extra steps should the solution be approved. 

The maintainer proposed a different solution that I'm looking into now. If that works out OK, this proposal may no longer be valid. 

The changes I made were only related to server side code, it would not impact anything related to JavaScript.

PR #1161, opened 2019-05-24
Fix short/int16 casting from OpenApi

epsitec:


MikeBeaton:
Why did this build fail? It looks like a completely genuine attempt to solve a small set of (apparently!) cut-and-paste typo errors around this bit of the code. I've applied these locally and they build and test fine!

MikeBeaton:
(PS I noticed that this and #1139 are in the same area of code - it looks at first glance unlikely that both commits are fully right, since #1139 effectively does not make the changes which this commit does make?)

MikeBeaton:
@epsitec Looking at the differences between this and #1139, it looks to me as if you need to modify:

 - [`Swashbuckle.AspNetCore.SwaggerGen.PrimitiveSchemaGenerator.FactoryMethodMap`](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/c1fb708e3fe3f2c62abf4a85b91786f38b66582e/src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGen/PrimitiveSchemaGenerator.cs#L91-L113)

as well?

Also the relevant tests:
 - [`Swashbuckle.AspNetCore.SwaggerGen.Test.SchemaGeneratorTests.GenerateSchema_GeneratesPrimitiveSchema_IfPrimitiveType`](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/6f9e899cfe46e8d99a12e42e378e7c03693dd1c2/test/Swashbuckle.AspNetCore.SwaggerGen.Test/SchemaGen/SchemaGeneratorTests.cs#L44-L72)
 - [`Swashbuckle.AspNetCore.SwaggerGen.Test.SchemaGeneratorTests.GenerateSchema_GeneratesEnumSchema_IfEnumType`](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/6f9e899cfe46e8d99a12e42e378e7c03693dd1c2/test/Swashbuckle.AspNetCore.SwaggerGen.Test/SchemaGen/SchemaGeneratorTests.cs#L92-L110)

I was wondering what the origin is of the type strings allowed in `OpenApiSchema.Format`... is `"int16"` actually 'valid'? (This document doesn't seem to say: https://tools.ietf.org/html/draft-fge-json-schema-validation-00) And what type should be used for a byte (#1139)? (Maybe these choices are purely internal to Swashbuckle.AspNetCore, I'm not quite sure.)

Assuming that `"int16"` is valid and that something like `"byte"` is valid for byte (i.e. mainly assuming that @domaindrivendev didn't already consciously decide against doing this!), then it's fairly easy to (and I'd be happy to) do a PR which combines the commits from this PR with the similar commits for the byte-valued enums from #1139, and with required (I think?) matching updates to the tests and factory methods mentioned above.

epsitec:
I had to apply the `int16` fix in order for my Core 3.0 preview5 to work after migrating to this latest release of the framework, as the existing code crashed when trying to cast a `short` while unboxing. I’m not at my computer for the next few days, so I’m not in a position to test anything.

domaindrivendev:
Closing because 1) "int16" is not a standard value for the schema `Format` and 2) I don't want to target 3.0 as the _minimum_ supported version

Issue #1164, opened 2019-05-29
Default value not working for enum's


MikeBeaton:
Default value is not working for enums, e.g.:

```c#
        public ActionResult<PagedResults<string>> Get(
            [FromQuery, BindRequired] string q,
            [FromQuery]int page = 1,
            [FromQuery]int pageSize = 20,
            [FromQuery]AutocompleteMatchType matchType = AutocompleteMatchType.BeginFirst)
```

then in the generated JSON:

```json
"parameters": [
  {
	"name": "q",
	"in": "query",
	"required": true,
	"schema": {
	  "type": "string"
	}
  },
  {
	"name": "page",
	"in": "query",
	"schema": {
	  "type": "integer",
	  "format": "int32",
	  "default": 1
	}
  },
  {
	"name": "pageSize",
	"in": "query",
	"schema": {
	  "type": "integer",
	  "format": "int32",
	  "default": 20
	}
  },
  {
	"name": "matchType",
	"in": "query",
	"schema": {
	  "enum": [
		"begin",
		"any",
		"beginFirst"
	  ],
	  "type": "string"
	}
  }
]
```

Noting that there is no default value in the JSON for the enum, even though one exists.

This seems to be down to the fact that `OpenApiPrimitiveFactory.FactoryMethodMap` does not contain any way to create an `OpenApiPrimitive<T>` for an enum, therefore the call to `schema.Default = OpenApiPrimitiveFactory.CreateFrom(parameterInfo.DefaultValue);` returns and sets a `null` value within the method `SwaggerGenerator.GenerateParameter(...)` making it look from then on as if there was no default value (even though `parameterInfo.DefaultValue` with the correct default value does exist, and is passed in to that call).

I'd be interested to work on this if it's considered useful. I'm not quite sure if it is even possible to generate an `OpenApiPrimitive<T>` where `T` is an enum type? If so, then some reflection-based code added in to `OpenApiPrimitiveFactory.CreateFrom(...)` to deal with that sounds like it might be the way to go? (Noting that the default value for an enum depends on whether the values for the enum are being output as `string` or `int`, so it would be important to display the default value in the correct format as well.)

MikeBeaton:
I got a bit further with this and had my own working version with support for enum defaults on parameters and properties, and supporting tests. It was based off `c887fbb` 7 Feb 2019, but then I realised quite a lot has changed.

As of the current master, it seems like:

[...]

PR #1165, opened 2019-05-30
Correct default value for camel-cased string enums

MikeBeaton:
Addresses the remaining issue of #1164 mentioned in the comments, i.e. "Default enum value ... not correctly camel-cased for enum serialized as string, when camel-casing is set in options or `StringEnumConverter`".

This is actually complete and correct (I hope!) as long as renaming enum members via `EnumMemberAttribute` is no longer supported, so that the only issue is camel-casing.

If arbitrary mapping/renaming of enum members is to be supported then for the reasons I mentioned in the issue - which I'm sure you're aware of! - I think there is no way to do a complete solution just within `TryCreateFor`, and - as far as I can work out - refactoring would be needed to change things so that default values (at least for enums) are calculated at the time when the `OpenApiSchema` is calculated (i.e. when the required mapping info is fully available). But if that's not to be supported then, as I say, I hope that this does the trick!

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/Generator/OpenApiAnyFactory.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/Generator/OpenApiAnyFactory.cs
@@ -32,7 +32,9 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
             else if (schema.Type == "string" && value.GetType().IsEnum) {
                 if (schema.Enum != null)
                 {
-                    var loweredValue = Enum.GetName(value.GetType(), value)?.ToLowerInvariant();
+                    var name = Enum.GetName(value.GetType(), value);
+                    var mappedValue = schema.Enum.FirstOrDefault(e => string.Equals(name, ((OpenApiString)e)?.Value, StringComparison.OrdinalIgnoreCase));
+                    if (mappedValue != null) openApiAny = mappedValue;
                     var mappedValue = schema.Enum.Select(e => ((OpenApiString)e)?.Value).Where(v => v.ToLowerInvariant() == loweredValue).FirstOrDefault();
                     if (mappedValue != null) openApiAny = new OpenApiString(mappedValue);
                 }
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/Generator/OpenApiAnyFactory.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/Generator/OpenApiAnyFactory.cs
@@ -33,10 +33,8 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
                 if (schema.Enum != null)
                 {
                     var name = Enum.GetName(value.GetType(), value);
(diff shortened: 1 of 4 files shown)


src/Swashbuckle.AspNetCore.SwaggerGen/Generator/OpenApiAnyFactory.cs
```diff
@@ -28,8 +29,14 @@ public static bool TryCreateFor(OpenApiSchema schema, object value, out IOpenApi
             else if (schema.Type == "number" && schema.Format == "double" && TryCast(value, out double doubleValue))
                 openApiAny = new OpenApiDouble(doubleValue);
 
-            else if (schema.Type == "string" && value.GetType().IsEnum)
-                openApiAny = new OpenApiString(Enum.GetName(value.GetType(), value));
+            else if (schema.Type == "string" && value.GetType().IsEnum) {
+                if (schema.Enum != null)
+                {
+                    var loweredValue = Enum.GetName(value.GetType(), value)?.ToLowerInvariant();
```
bbarry: consider instead:

```suggestion
                    var name = Enum.GetName(value.GetType(), value);
```

src/Swashbuckle.AspNetCore.SwaggerGen/Generator/OpenApiAnyFactory.cs
```diff
@@ -28,8 +29,14 @@ public static bool TryCreateFor(OpenApiSchema schema, object value, out IOpenApi
...
+                    var mappedValue = schema.Enum.Select(e => ((OpenApiString)e)?.Value).Where(v => v.ToLowerInvariant() == loweredValue).FirstOrDefault();
```
bbarry:
```suggestion
                    var mappedValue = schema.Enum.FirstOrDefault(e => string.Equals(name, ((OpenApiString)e)?.Value, StringComparison.OrdinalIgnoreCase));
```

src/Swashbuckle.AspNetCore.SwaggerGen/Generator/OpenApiAnyFactory.cs
```diff
@@ -28,8 +29,14 @@ public static bool TryCreateFor(OpenApiSchema schema, object value, out IOpenApi
...
+                    if (mappedValue != null) openApiAny = new OpenApiString(mappedValue);
```
bbarry:
```suggestion
                    if (mappedValue != null) openApiAny = mappedValue;
```

src/Swashbuckle.AspNetCore.SwaggerGen/Generator/OpenApiAnyFactory.cs
MikeBeaton: Thanks!

bbarry: Anytime, hopefully https://github.com/isaacs/github/issues/1452 happens so that I could have provided a working suggestion without necessitating your further changes.

MikeBeaton:
#1843 replaces this

PR #1170, opened 2019-06-10
Ability to specify a configuration file when using the CLI tool

tnewman16:
Our application uses the Options pattern for configuration. On startup we validate our options class to avoid runtime errors. However, we're unable to use the CLI tools because the configuration file is defaulted to empty.

These changes will allow users to specify a specific host configuration when using the `dotnet swagger` tool. The value passed into the `--config` flag should be a relative path that points to a valid JSON configuration file, which will be loaded up when the host is built.


domaindrivendev:
@tnewman16 - I think your specific issue can be resolved by using the custom host configuration hooks described here:

https://github.com/domaindrivendev/Swashbuckle.AspNetCore#use-the-cli-tool-with-a-custom-host-configuration

With this approach, you can have the CLI tool share the same host configuration as your actual application, including JSON config files etc.

lockdrew:
This would be super helpful - only issue I'm trying to figure out is why I need to have a connection string env variable to generate swagger docs.

PR #1172, opened 2019-06-11
Use new Microsoft.Extensions.ApiDescription.Server package

dougbu:
- provides document download on build and aligns with Microsoft.Extensions.ApiDescription.Client p2p references
  - this feature is on by default
  - can be disabled using `<OpenApiRetrieveDocuments>false</OpenApiRetrieveDocuments>`
  - timing can be changed using `<OpenApiRetrieveDocumentsAfterBuild>false</OpenApiRetrieveDocumentsAfterBuild>`
    - then, depend on the `RetrieveOpenApiDocuments` target from somewhere else

Details
- make `IDocumentProvider` `internal` as strongly recommended
- rename `Microsoft.Extensions.ApiDescription` -> `Microsoft.Extensions.ApiDescriptions`
- align with Microsoft.Extensions.ApiDescription.Server package
- add Microsoft.Extensions.ApiDescription.Server reference to Swashbuckle.AspNetCore


dougbu:
@domaindrivendev please let me know if this needs to change at all.

FYI our 3.0 Preview 6 release should be soon.

dougbu:
3.0 Preview 6 packages are now public. This PR can be safely merged, perhaps with a release note mentioning `<OpenApiRetrieveDocuments>false</OpenApiRetrieveDocuments>` for .NET Core 2.0 and earlier web projects using the latest Swashbuckle.AspNetCore

dougbu:
Ping @domaindrivendev

dougbu:
Much appreciated @domaindrivendev❕ When do you expect your next public release -- preview or otherwise❔

domaindrivendev:
@dougbu - 5.0.0 includes significant refactors to support Swagger/OAI 3 and although these were long overdue, they also introduced a number of regression bugs that have been reported during the pre-release phase (see https://github.com/domaindrivendev/Swashbuckle.AspNetCore/milestone/7).

As you can see there's ~10 or so issues there that I'd like to address before the stable 5.0.0 release. I'm currently working on a change that will address several of them in one go but it's going slow due to other commitments. I've no cycles this week and so it will be next week before I get back to it. I'll see where I am after that and post an update on timelines end of next week.

src/Swashbuckle.AspNetCore/Swashbuckle.AspNetCore.csproj
```diff
@@ -1,23 +1,50 @@
...
     <RepositoryType>git</RepositoryType>
     <RepositoryUrl>https://github.com/domaindrivendev/Swashbuckle.AspNetCore.git</RepositoryUrl>
+
+    <!-- Package contains only the final .nuspec file with a few listed dependencies. -->
+    <GenerateNuspecDependsOn>$(GenerateNuspecDependsOn);PopulateNuspec</GenerateNuspecDependsOn>
+    <IncludeBuildOutput>false</IncludeBuildOutput>
+    <IncludeContentInPack>false</IncludeContentInPack>
+    <IncludeSource>false</IncludeSource>
+    <IncludeSymbols>false</IncludeSymbols>
+    <MicrosoftExtensionsApiDescriptionServerPackageVersion>0.3.0-preview6.19307.2</MicrosoftExtensionsApiDescriptionServerPackageVersion>
```
domaindrivendev: @dougbu Now that it's out of preview, I'm looking to upgrade the version of `ApiDescription.Server` that comes bundled with Swashbuckle. This brought me back to this PR, and now that I'm looking at it a little closer, I have a couple of questions I was hoping you could answer, so that I can understand things a little better:

1. Why does `Swashbuckle` need the dependency on the `ApiDescription.Server` package? It's not referenced anywhere in the SB code. If the goal is to introduce additional tooling, wouldn't it be more appropriate for the application to take that dependency instead of it being bundled in with SB?

2. In this PR, what was the rationale behind introducing an explicit .nuspec file and target to populate it? I preferred the simplicity (e.g. less moving parts) when it was just a single .csproj file. IIRC that was working just fine - was there something you needed for the `ApiDescription.Server` stuff that wouldn't have worked with the original approach? If so, can you let me know what that was? Thanks

dougbu: @domaindrivendev 
1. Swashbuckle needs a dependency on ApiDescription.Server to reduce the friction for those who want document generation on build. As we've discussed offline, it makes sense for this feature to be opt-in.
  Opting in should be as simple as possible and we made a choice a while ago to hide the Microsoft.Extensions.ApiDescription.Server package (which is very small) from users. It's completely unusable without Swashbuckle.AspNetCore in any case.
2. I had to use an explicit .nuspec because NuGet doesn't handle the case of a package that depends on a dev-only package without help. To avoid making this too horrendous, I pass properties from the .csproj for everything -- trying to avoid a need to edit the .nuspec in the future (unless the package content needs to change).
  In any case, making the doc generation features opt-in needs a tiny (single property) Swashbuckle.AspNetCore.props file in your package. The .nuspec makes this change trivial.

/cc @bradygaster

dougbu: (Generally, packages containing MSBuild goop are difficult to manage w/o .nuspec files.)

PR #1178, opened 2019-06-22
Removed excessive code

DamirAinullin:
value is checked for null before on line 53:

```
if (value == null || headerSpec.Schema == null) continue;
```


bbarry:
I think the removed line was supposed to read

    if (schema == null) continue;

as it is possible the schema is a reference to another document and would not be able to be resolved ([although that case happens to currently throw an exception which is not dealt with here either](https://github.com/microsoft/OpenAPI.NET/blob/master/src/Microsoft.OpenApi/Models/OpenApiDocument.cs#L282)). It is also not guaranteed by the header component dictionary that the value at that reference is not null (which I believe would be an invalid document).

PR #1187, opened 2019-07-07
Update AppVeyor definition to use .NET Core Preview 6

kant2002:


say25:
This can be closed as `master` is already using stable 3.X

say25:
This can be closed as `master` is already using stable 3.X @domaindrivendev

PR #1189, opened 2019-07-07
Expose `--serializeasv2` command line parameter

kant2002:
This parameter often nescessary to use now, since Atuorest has a lot of issues with Open API 3.0

### DIFF
--- a/README-v5.md
+++ b/README-v5.md
@@ -251,6 +251,12 @@ app.UseSwagger(c =>
 });
 ```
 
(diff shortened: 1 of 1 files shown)


domaindrivendev:
Not sure when this got added but looks like this is already there:
https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/master/src/Swashbuckle.AspNetCore.Cli/Program.cs#L35

PR #1209, opened 2019-07-27
Update README to point to v5 instructions

alefranz:
I think it's a bit unexpected to have to refer to a different README when using AspNetCore 3.0, even if I understand the reason for it.
I've added a message on the main README to point to it.

- I haven't used the previous table as it is not related to major upgrades of the library.
- The markdown looks ugly but having the tables of different size was a bit ugly. Also not using a table looked a bit off


krdmllr:
The linebreak is a bit weird
grafik

alefranz:
No longer needed as v5 README doesn't exist anymore

PR #1234, opened 2019-08-27
fixes some null refs

aersamkull:
We could not compile because of these Null-Refs. The fixes are trivial, as seen below


aersamkull:
The error occurs when using: `[DefaultValue(null)]`

src/Swashbuckle.AspNetCore.SwaggerGen/Generator/OpenApiAnyFactory.cs
```diff
@@ -25,14 +25,14 @@ public static bool TryCreateFor(OpenApiSchema schema, object value, out IOpenApi
             else if (schema.Type == "number" && schema.Format == "double" && TryCast(value, out double doubleValue))
                 openApiAny = new OpenApiDouble(doubleValue);
 
-            else if (schema.Type == "string" && value.GetType().IsEnum)
+            else if (schema.Type == "string" && value !=null && value.GetType().IsEnum)
```
EvgeniyZ: I think it's better to format it, add whitespace before null

aersamkull: Done :)

domaindrivendev:
No longer neccessary

PR #1255, opened 2019-09-14
Add feature : SwaggerPropertyAttribute

olihou:
Allow to override the documented property type (in case of JSONConverter for instance)

### DIFF
--- a/README-v5.md
+++ b/README-v5.md
@@ -715,6 +715,27 @@ services.AddSwaggerGen(c =>
     c.MapType<PhoneNumber>(() => new OpenApiSchema { Type = "string" });
 };
 ```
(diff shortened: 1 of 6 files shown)


domaindrivendev:
I don't understand the problem this solves or the sample usage provided with the readme updates. Have you looked at the existing  `SwaggerSchema` and `SwaggerSchemaFilter`  attributes - these might help with the problem you're trying to address. Additionally, this PR is now out-of-date (my bad not yours) and there's conflicts so I'm going to close for now.

If your problem is still valid, please start with an issue that clearly articulates it with uses cases and code examples - then we can start from there and figure out a suitable path to resolution. Thx

PR #1264, opened 2019-09-19
support deep nested class

lust4life:
Then this can generate nested class's description for swagger json.


lust4life:
@domaindrivendev hi, would you consider to support this for these who using deep (more then two) nested class?

domaindrivendev:
Looks like the same underlying issue was ultimately fixed by #1690.

PR #1268, opened 2019-09-20
Restore SchemaRepository.TryGetIdFor

lemoinem:
Some manipulation requires us to have access to a mapping from Type to SchemaID/Schema.
Unless there is a better way to find the SchemaID associated to a Type, it would be much helpful to restore `SchemaRepository.TryGetIdFor`.

For example, we use this to implement discriminator for Polymorphic schema.

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SchemaRepository.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SchemaRepository.cs
@@ -32,6 +32,11 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
             };
         }
 
+        public bool TryGetIdFor(Type type, out string schemaId)
+        {
+            return _reservedIds.TryGetValue(type, out schemaId);
+        }
+        
         private void ReserveIdFor(Type type, string schemaId)
         {
             if (_reservedIds.ContainsValue(schemaId))


redoz:
@lemoinem Do you find workaround for this?

lemoinem:
@redoz Staying at 5.0.0-rc2 was my only option.

redoz:
@lemoinem In my case I was able to switch to using `context.SchemaGenerator.GenerateSchema(...)` it seems to be a noop if it already exists.

lemoinem:
@redoz, Yes, but I didn't find the way to get the schema ID from the Schema

domaindrivendev:
@lemoinem @redoz - TBH, this isn't a use-case I've really considered before. Taking a step back from the solution/implementation detail, can you provide some more detail around what you're trying to do in the filter and why?

redoz:
@domaindrivendev Looking back at it, I think all of my filters are for patching up the schema for various reasons. Some were created to change behavior I disagreed with (not sure if they are bugs or by design, but at the time I just needed to make it work, so there might be better ways of accomplishing all of these things):

For instance, I have this to ensure that every operation that can return a 404 has our custom response schema. This used to to try to resolve the Schema and only generating it if doesn't exist, but as it seems GenerateSchema is idempotent I just did it this way:
```
    public class SwaggerNotFoundFilter : IOperationFilter
    {
        private readonly Type _notFoundProblemDetailsType;

        public SwaggerNotFoundFilter()
        {
            _notFoundProblemDetailsType = typeof(NotFoundProblemDetails);
        }

        public void Apply(OpenApiOperation operation, OperationFilterContext context)
        {
            if (operation.Responses.TryGetValue("404", out var openApiResponse))
            {
                var contentKeys = openApiResponse.Content.Keys.ToArray();

                var schemaReference = context.SchemaGenerator.GenerateSchema(_notFoundProblemDetailsType, context.SchemaRepository);

                foreach (var mimeType in contentKeys)
                {
                    openApiResponse.Content[mimeType] = new OpenApiMediaType
                                                        {
                                                            Schema = schemaReference
                                                        };
                }
            }
        }
    }
```

Another one is this one, something changed that ended up setting my arrays as Nullable (which I guess technically is correct as they are reference types), so I introduced a NotNullAttribute that I can apply to members of my response models to indicate that they cannot be null. Unfortunately I couldn't figure out how to go from a property (name + schema) to the real underlying ProperyInfo, it would be nice if functionality like that existed:
```
    public class SwaggerNotNullFilter : ISchemaFilter
    {
        public void Apply(OpenApiSchema schema, SchemaFilterContext context)
        {
            var actualType = context.ApiModel.Type;

            Patch(schema, actualType);
        }


        private void Patch(OpenApiSchema schema, Type type)
        {
            foreach (var (name, propertySchema) in schema.Properties)
            {
                // this isn't nice but I couldn't figure out how to go from property schema to PropertyInfo
                var propertyInfo = type.GetProperties().SingleOrDefault(pi => StringComparer.OrdinalIgnoreCase.Equals(pi.Name, name));

                if (propertyInfo is object)
                {
                    PatchProperty(propertySchema, propertyInfo);
                }
            }
        }

        private void PatchProperty(OpenApiSchema schema, PropertyInfo propertyInfo)
        {
            var notNullAttr = propertyInfo.GetCustomAttribute(typeof(NotNullAttribute));

            if (notNullAttr is object)
                schema.Nullable = false;
        }
    }
```
I also have a filter for handling Nullable<T> parameters correctly (again, this might not be necessary anymore, I haven't checked in a while):
```
    public class SwaggerNullabilityFilter : IParameterFilter
    {
        public void Apply(OpenApiParameter parameter, ParameterFilterContext context)
        {
            if (context.ParameterInfo.ParameterType.IsGenericType && context.ParameterInfo.ParameterType.GetGenericTypeDefinition() == typeof(Nullable<>))
            {
                // check if this is just a reference
                if (parameter.Schema.Reference != null)
                {
                    // get copy of schema
                    var schema = context.SchemaRepository.Schemas[parameter.Schema.Reference.Id];
                    
                    // remove schema ref
                    schema = new OpenApiSchema
                                 {
                                     Description = schema.Description,
                                     Type = schema.Type,
                                     Enum = schema.Enum,
                                     Default = schema.Default,
                                     Nullable = schema.Nullable,
                                     Required = schema.Required
                                 };
                    parameter.Schema = schema;
                }

                parameter.Schema.Nullable = true;
            }
        }
    }
```

I also had to create this filter at some point so that default values would be supported on enum parameters (same disclaimer as above, not sure if this is still needed), here I'm mostly missing a decent way of cloning a Schema as well as navigating from the OpenApi object model to the source Types.
```
public class SwaggerDefaultValueFilter : IOperationFilter
    {
        public void Apply(OpenApiOperation operation, OperationFilterContext context)
        {
            OpenApiString GetDefaultEnumValue(OpenApiSchema enumSchema, object defaultValue)
            {
                return enumSchema.Enum
                                 .OfType<OpenApiString>()
                                 .Single(enumMember => StringComparer.OrdinalIgnoreCase.Equals(enumMember.Value, defaultValue.ToString()));
            }

            var apiDescription = context.ApiDescription;
            var apiVersion = apiDescription.GetApiVersion();
            var model = apiDescription.ActionDescriptor.GetApiVersionModel(ApiVersionMapping.Explicit | ApiVersionMapping.Implicit);

            operation.Deprecated = model.DeprecatedApiVersions.Contains(apiVersion);

            if (operation.Parameters == null)
            {
                return;
            }

            foreach (var parameter in operation.Parameters)
            {
                var description = apiDescription.ParameterDescriptions
                    .First(p => p.Name == parameter.Name);

                if (parameter.Description == null)
                {
                    parameter.Description = description.ModelMetadata?.Description;
                }

                // enable support for default values for enum
                if (!(description.DefaultValue is null) && description.Type.IsEnum)
                {
                    //... and we're enum
                    if (parameter.Schema.Reference != null)
                    {
                        // get copy of schema
                        var enumSchema = context.SchemaRepository.Schemas[parameter.Schema.Reference.Id];
                        
                        // "clone" schema and 
                        enumSchema = new OpenApiSchema
                                     {
                                         Description = enumSchema.Description,
                                         Type = enumSchema.Type,
                                         Enum = enumSchema.Enum,
                                         Nullable = enumSchema.Nullable,
                                         Required = parameter.Schema.Required
                                     };
                        parameter.Schema = enumSchema;
                    }

                    parameter.Schema.Default = GetDefaultEnumValue(parameter.Schema, description.DefaultValue);
                }

                parameter.Required |= description.IsRequired;
            }
        }
    }
```

Not sure if any of this useful for you, but you asked about my use-cases and here they are :)

lemoinem:
In our case, we want to have polymorphic schemas, with a discriminator.
We have to load the discriminator somewhere, and need the schemaID for the associated types:

```
public class SwaggerDiscriminator : ISchemaFilter, IDocumentFilter
{
	public void Apply(OpenApiSchema schema, SchemaFilterContext context)
	{
		if (schema.Reference != null || schema.Discriminator != null)
		{
			return;
		}

		DiscriminatorConfiguration discriminator = DiscriminatorConfiguration.LoadConfigurationForType(context.Type);

		schema.Discriminator = new OpenApiDiscriminator
		{
			PropertyName = discriminator.Name,
		};

		if (discriminator.Mapping != null)
		{
			schema.Extensions.Add(
				DiscriminatorMapExtension.EXTENSION_NAME,
				new DiscriminatorMapExtension(discriminator.Mapping)
			);
		}
	}

	{
		foreach (OpenApiSchema schema in swaggerDoc.Components.Schemas.Values.Where(s => s.Extensions.ContainsKey(DiscriminatorMapExtension.EXTENSION_NAME)))
		{
			DiscriminatorMapExtension mapExtension = (DiscriminatorMapExtension)schema.Extensions[DiscriminatorMapExtension.EXTENSION_NAME];

			schema.Extensions.Remove(DiscriminatorMapExtension.EXTENSION_NAME);
			schema.Discriminator.Mapping = mapExtension.Mapping.ToDictionary(
				kvp => kvp.Key,
				kvp => context.SchemaRepository.TryGetIdFor(kvp.Value, out string schemaID) ? schemaID : null
			);
		}
	}
}
```

The call to `TryGetIdFor` is in the `Apply(OpenApiSchema, DocumentFilterContext )` so we can generate the `Discriminator.Mapping`. I don't know how else this can be achieved...

CodeTherapist:
I do have exactly the same issue.
Similar to @lemoinem, I try to provide multiple schema/example pair and replace the generated schema/example. The reason for that is, the type of the parameter of the API has a different structure than the required request json payload and needs to be removed from the schema repository (there is a complex modelbinder that can read the payload and map it to the type).

Thus, I would need a way to get the the SchemaId of a particular type.

The schemaId and the type is something that is anyway public and known - what is here the implementation detail? I would appreciate, when this could be elaborated @domaindrivendev.

lemoinem:
@domaindrivendev Any news on that?

We need support for discriminator and this is blocking us to update to more recent versions. We also would prefer not to be bared from moving to 5.0 when it is released. But without this featured, I see no way of generating discriminators reliably.

Issue #18, opened 2015-08-04
SwaggerUi on Spartan IE


OzBob:
On VS2015 (14.0.23107.0) download Zip of Ahoy, run project on Win10 VM on azure,  
Open in Project Spartan (IE vNext)
http://localhost:1417/swagger/ui/index.html
does not serve any pages but does on Chrome

jderus:
I couldn't repro this problem using VS2015 RTM on Win10 non-VM.
Launched and worked in Chrome and Edge, using BasicApi as the startup project.

OzBob:
I think because I had an pre-release version of win10VM from Azure, it has a crap version of Spartan.  I'll try on a newer version and if that's a problem raise another issue.  I have no idea how to get the version of Spartan installed on the VM, but I'll get a new version of win10 and try with Edge. I'm sure it's not a Ahoy issue.

PR #1283, opened 2019-09-30
Extend OpenApiAnyFactory to support array-type schemas and multivalued defaults

jimfrederic:
Using 5.0.0-rc3, I had an Operation with a Parameter that was multivalued.  I wanted to generate a swagger doc with a default for the parameter, and the default should be an array with a single value.  However the OpenApiAnyFactory does not handle arrays.  This PR adds that support.

As a result I can annotate my parameter
```
[DefaultValue(25)] int[] distance,
```

and it will now now generate
```
{
  "name": "distance",
  "in": "query",
  "description": "An array of distances",
  "schema": {
    "type": "array",  
    "items": {
      "type": "integer",
	  "format": "int32"
    },
    "default": [
      25
    ]
  }
}
```


domaindrivendev:
Closing in favor of #18

Issue #1308, opened 2019-10-09
Wrong PackageReference in Swashbuckle.AspNetCore.Swagger.csproj ?


tebeco:
Hello.

As you can see here :
https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/ad91d72078e1e2276824b38e2483c1d8ab6e5376/src/Swashbuckle.AspNetCore.Swagger/Swashbuckle.AspNetCore.Swagger.csproj#L21-L24
It seems that this nuget (if still supposed to be `aspnetcore3` compliant), need to also target `netcoreapp3.0`

And this `PackageReference` need to be toggle featured based on `$(TargetFramework)`,
That would be `FrameworkReference` for `netcoreapp3.0` then ?
As you can see here :
https://www.nuget.org/packages/Microsoft.AspNetCore.Mvc.Core
Last release was 2.2.5, so it's probably within the Metapackage (that is now part of the `FrameworkReference`

I'm not sure it will behave nicelly on `netcoreapp3.0` + `aspnetcore3.0` (blow at runtime ?) is v5.0.0 is shipped with that assembly => `MissingMethodException`

domaindrivendev:
As it stands, v5.0.0-rc4 is fully compatible with versions 2 & 3 of ASP.Net Core - i.e. it doesn't blow up 😉 

This is because the `PackageReference` syntax used above states that Swashbuckle requires a version of ASP.NET Core ">=" 2.0.0. See [here](https://docs.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files#controlling-dependency-version) for more info on how this syntax is used to control dependency versions.

domaindrivendev:
Not sure what you mean by "this dependency does not exist anymore". Follow the same link you posted and click the "+ Show More" link at the bottom of the page and you'll see it listed.

Regarding the rest of your points, I'm not exactly sure of the point you're trying to make - very possible a lack of knowledge/context on my side. Perhaps you could provide a concise set of repro steps that demonstrate the problem or some steps to walk me through the hypothetical chain of events you're concerned about?

[...]

PR #1316, opened 2019-10-12
Cross target netcoreapp3

tebeco:
This PR intend to cross target:
* `netstandard2.0` for `AspNetCore 2.0` stack
* `netcoreapp3.0` for `AspNetCore 3.0` stack

### Important :
I attempted to make few explanations on the new `FrameworkReference` are here :  
https://gist.github.com/tebeco/a6e0d9a3885a0a36e702795219bd4fe9

#### TLDR :
As a reminder, `AspNetCore 3` is not `netstandard2.0` anymore, it's targetting `netcoreapp3.0`, this is why nugets are all cross targetting `netcoreapp3.0` (to get a full resolution on the `TargetFramework`)

Also, when targetting `netstandard2.0` it is now important to pin down dependencies or transitive to a dependency chain compliant with both `Microsoft.AspNetCore.X` AND `Microsoft.Extensions.X` (`version < 3.0.0`)  
Same applies with `netcoreapp3.0` (`version >= 3.0.0`)

### To reviewers :
Some points of attentions :
- [ ] [IdentityServerPrincipal has been removed](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/pull/1316/commits/150b99b9032105eb18ae50aeb065169d642823d8)
- [ ] Can obsolete call to `Newtonsoft.Json` be [the cause of this](https://github.com/tebeco/Swashbuckle.AspNetCore/blob/94aa61fd45847e4ca66e96090374e4939771608f/src/Swashbuckle.AspNetCore.SwaggerGen/Newtonsoft/NewtonsoftApiModelResolver.cs#L86-L91)
- [ ] Shoud i update `Microsoft.NET.Test.Sdk` to 16.3 ? Since `netcoreapp3.0` need at least Vs2019 16.3 ?

Will also do in this PR :
- [x] Update `DOTNET_VERSION` in `AppVeyor` from `3.0.100-preview8` to `3.0.100`
- [x] Add `global.json` to `3.0.100`
- [x] Reworking all `Program.cs` to have a proper bootstrap per `TargetFramework`
- [x] Reworking all `Startup.cs` to have a proper bootstrap per `TargetFramework`
- [ ] Update `XUnit` to latest
- [ ] Need to clarify [questions here](https://github.com/aspnet/AspNetCore/issues/14954)

### Notes
Since this version is a major, concider dropping `netcoreapp2.0` support as [it already exceeded `End of Life` support](https://dotnet.microsoft.com/download/dotnet-core) :
image
Do you want we to update every single reference to at least `2.1.0` ?  
Same should be applied to all `UnitTest` and `IntegrationTest` for their `TargetFramework` (i had to upgrade one in the process if you look at the PR details)
____
As you can see, `Microsoft.AspNetCore.Mvc.Versioning` has not been release for `AspNetCore 3.0` (it will probably be for their `v4.0.0`)
____
`SwaggerGen` relies on obsolete Api, would love to know how to get rid of both `DescribeAllEnumAsString` and `DescribeStringEnumsInCamelCase`.  
Last bound found is the PR was linked to the use of `Obsolete` API of `Newtsonsoft` :  
`var converter = new StringEnumConveter(bool)`  
v3.0 uses `v12.x` that now have a `NamingStrategy` instead of a `bool` ending up throwing `NullReferenceException`


tebeco:
(Sorry for all the `--force-with-lease`, trying to keep a consistent commit history on this one

domaindrivendev:
> Do you want we to update every single reference to at least 2.1.0
👍

test/Swashbuckle.AspNetCore.Annotations.Test/ModelMetadataHelper.cs
```diff
@@ -19,6 +25,7 @@ public static IModelMetadataProvider GetDefaultModelMetadataProvider()
                     new DataAnnotationsMetadataProvider(Options.Create(new MvcDataAnnotationsLocalizationOptions()), null),
                 }
             ));
+#endif
         }
     }
 }
```
domaindrivendev: ^^^ This file can just be deleted as it's deadwood, not referenced anywhere

test/WebSites/Basic/Program.cs
```diff
@@ -1,20 +1,42 @@
...
+            CreateHostBuilderAndRun(args);
+        }
+
+        public static void CreateHostBuilderAndRun(string[] args) =>
+#if NETCOREAPP3_0
+            NC3_CreateHostBuilder(args).Build().Run();
+#else
+            NC2_CreateHostBuilder().Build().Run();
+
+#endif
```
domaindrivendev: This is exactly the kind of complexity I was hoping to avoid/discourage by not multi-targeting. Is this really neccessary with this PR. I want to minimize the conditional code throughout the entire repo.

test/WebSites/Basic/Program.cs
```diff
@@ -1,20 +1,42 @@
...
                 .UseIISIntegration()
-                .UseStartup<Startup>()
-                .Build();
-
-            host.Run();
-        }
+                .UseStartup<Startup>();
+#endif
     }
 }
```
domaindrivendev: Same as above

domaindrivendev: I'm halfway through the review and just don't feel comfortable with the proliferation of cross-targeting conditionals throughout the code. As I stated before, my preference is to target the lowest dependency versions that work in all cases, something which David Fowler seems to indicate is still possible. From https://github.com/aspnet/AspNetCore/issues/14393#issuecomment-535111340:
> Not really no. You don't need to target netcoreapp3.0 unless you're trying to use new features in netcoreapp3.0 (or if you end up hitting a breaking change between versions). The library author guidance is usually to target the lowest version that satisfies where you need to run. You only end up needing to multi-target when you want to take advantage of new functionality in the newer version or if you run into something truly incompatible between the 2 versions.


> Not really no. You don't need to target netcoreapp3.0 unless you're trying to use new features in netcoreapp3.0 (or if you end up hitting a breaking change between versions). The library author guidance is usually to target the lowest version that satisfies where you need to run. You only end up needing to multi-target when you want to take advantage of new functionality in the newer version or if you run into something truly incompatible between the 2 versions.

domaindrivendev: If a project in the Swashbuckle needs to use new features in netcoreapp3.0 or is hitting breaking changes, as was the case in the projects that _already_ multi-targeted prior to this PR, then obviously multi-targeting is the only way. However, if this isn't the case, then I think I'd prefere to target netstandard2.1 and therefore minimize the need for cross-targeting conditionals

tebeco: TBH, `Swashbuclke` would really need/love to use both `UseRouting` and `UseEndpoints()` as all routes metadata are now properly handled by `UseRouting` and you would be able to use it latter on, and the middleware could then be `.MapSwagger()` in the endpoints.

I trully wonder if all the new `Endpoint Routing` stack could in fact help you, as you only want to support 1 code base, the difference between 2.x and 3.x are huge regarding the routing.

Starting January 2020, only the `2.1` will be "supported" (not 2.2 anymore), and `3.x`. The version `3.1` will be LTS.
The idea of this cross targeting will let you be able to still support LTS (`netstandard2.0`) and currently `3.0` that totally change the routing, this is why i attempts in this PR to leverage both side of the `Mvc` that works in their own way.
I do intend to take a look after this PR at :
* Swashbuckle for `UseEndpoints()`
* `System.Texty.Json` (default ?) vs external package for the `NewtonsoftApiModelResolver` (not sure how yet, but as of today `Newtonsoft` hard dependency in swagger might not let user move completly out ^^)
* Use `Static Content Asset` (new in 3.0) to serve front/static file from libs

test/WebSites/Basic/Program.cs
tebeco: I can totally rename it.
It's just a copy paste of both 2.x and 3.0 default `Program.cs`

The idea is that now i'm sure all the test you wrote runs sucessfully on both stack.
There was only 1 project for `netcreapp3.0`, so it add lots of consistency and insurrance in fact since full test runs on all sample.

On top of that it might help people understanding how/why changes can be done from 2.x to 3.0

tebeco: Do you want me to rebase that change to something like this ? 
https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/b251ef18a9f855a2532f93e7d3d723f63dad0df9/test/WebSites/CliExample/Program.cs
it's probably more accurate to default template

tebeco:
@domaindrivendev 

if i bump samples to 2.1, do you prefer new commit or `rebase interactive` ? to have the same history as today ?
not sure how comments will react to a full rebase, but their will be less changes over the codebase.

test/Swashbuckle.AspNetCore.Annotations.Test/ModelMetadataHelper.cs
tebeco: Can you elaborate on that ? that's really the part of the code i'm loosing track, i've just used the `Empty` one for now in order to makes the code bui;d (PR still in Draft).
52 test are still KO because something is this class does not generate proper fake data for 3.0 ;) not sure what yet

domaindrivendev: I'm just saying the entire file `ModelMetadataProvider` can be deleted entirely (as opposed to modified) as it's no longer in use.

test/WebSites/Basic/Program.cs
domaindrivendev: Understood that SB _could_ benefit from the newer features but rather than pre-empting the need for them, can we add those dependencies (and therefore cross-target conditionals) as necessary - i.e. in future PRs?

For the sample apps, I think I'd prefer for them to just target netcoreapp3.0. There purpose is to demonstrate different configuration options with Swashbuckle - not test coverage and not documenting upgrades. Those could be reasonable well covered by just having one sample app called `NetCore2`, which would ensure everything is backwards compatible and demo's setup for the older framework version

domaindrivendev: Regarding your bullet points above - I'm already in the process of incorporating `System.Text.Json` so please don't start a PR on that as we'll be duplicating work 😉. My plan for this to support the new Json API by default. As part of this, I'll move `NewtonsoftApiModelResolver` to a separate package that's explicitly opt-in - basically mirroring the .NET Core 3.0 defaults.

domaindrivendev:
> if i bump samples to 2.1, do you prefer new commit or rebase interactive ? to have the same history as today ?
Not sure what you mean. If we get to a point of merging this PR, I'll likely want to squash all of your commits into one.

test/Swashbuckle.AspNetCore.Annotations.Test/ModelMetadataHelper.cs
tebeco: oO right, thx, deleted

tebeco:
> Not sure what you mean. If we get to a point of merging this PR, I'll likely want to squash all of your commits into one.

As you want, i splitted them on purpose to understand the process "from -> to" :)

test/WebSites/Basic/Program.cs
tebeco: > Regarding your bullet points above - I'm already in the process of incorporating System.Text.Json so please don't start a PR 

Yup saw that this week end don't worry
also each bullet point would be separated PR but these can land latter, whereas this one intend to fully ensure that all dependencies will resolves `2.x` or `3.x` but not a mix of `2.x` + `3.x` for `aspnetcore3`.

If you're confident with `v5.0.0` this PR can land latter, i just wanted to give an idea of what i thought woul dhave been the best for `AspNetCore 3` stack.

How do you want to proceed ?

domaindrivendev: So, I think these changes _should_ make their way into the Swsahbuckle code, I just wonder if it should happened on a more incremental/as-needed basis. If nothing else, smaller PR's that each solve a very specific problem will help me understand them better.

You clearly have a deep understanding of the .NET Core infrastructure and so I'm excited to have someone with that level of knowledge help transition the project so that it's optimized for the latest stable stack.

With that said, I think we could proceed with an incremental list of changes. For example:

1. Out-of-the-box support for `System.Text.Json` (in work)
2. Upgrade unit tests AND sample apps to use the latest and greatest 3.0 configuration, and retain an additional `NetCore2` sample app to ensure and demonstrate backwards compatibility
3. Leverage the new Endpoints API
4. Leverate static content assets for the UI projects

I expect 3 & 4 will involve some level of mult-targeting to continue supporting 2.0 and 3.0 stacks, and I think these would be the forcing functions to introduce those dependencies as neccessary.

As I'm currently working no 1 above, you could tackle any one (or all) of the others. Thoughts? (and thanks for all your input so far).

test/Swashbuckle.AspNetCore.SwaggerGen.Test/Fixtures/Fakes/FakeApiDescriptionGroupCollectionProvider.cs
```diff
@@ -73,7 +78,12 @@ public ApiDescriptionGroupCollection ApiDescriptionGroups
         {
             var descriptor = new ControllerActionDescriptor();
 
+#if NETCOREAPP3_0
+            // SEE : https://github.com/aspnet/AspNetCore/issues/14454#issuecomment-535571938
+            descriptor.SetProperty(new ApiDescription());
```
pranavkm: You might need to use reflection to do this until an equivalent API is added:

```C#
var type = typeof(ControllerBase).Assembly.GetType("Microsoft.AspNetCore.Mvc.ApiDescriptionActionData", throwOnError: true);
var instance = Activator.CreateInstance(type);
actionDescriptor.Properties[type] = instance;
```

tebeco: oO i think it would have took me few days/week to realize that
naming got me ;) that will teach me to reach/check what i use ahahah

thx a lot @pranavkm

davidfowl: > You might need to use reflection to do this until an equivalent API is added:

Is this something ASP.NET Core is just missing?

tebeco: @pranavkm it fixed all test except one, i'll take a look at it today, in any case if you got another idea on what i missed

tebeco: @davidfowl i'm not really sure to be honest, if you check this commit :
https://github.com/tebeco/Swashbuckle.AspNetCore/commit/ed7888be8fd41141279583916afbc42bc9d9d6d8
This is a `Fake` for `IApiDescriptionGroupCollectionProvider`

The fix proposed by @pranavkm  seems to have fixed 46 `IntegrationTest` with the existing codebase.
Is there a better code fix ? that would be "doable" with the existing subset of API in 3.0 ?
(Also note that as of today the repo is not using either `UseRouting()` or `UseEndpoints()` (just trying to fully cross target in this PR)

The main issue is this :
https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/ad91d72078e1e2276824b38e2483c1d8ab6e5376/src/Swashbuckle.AspNetCore.Swagger/Swashbuckle.AspNetCore.Swagger.csproj#L21-L24

(i'm not sure how "sure" it could be to track `Microsoft.AspNetCore.Mvc.Core` (and other package from 2.0.0) in 3.0 app, so i attempted the full cross-targeting

test/WebSites/Basic/Program.cs
domaindrivendev: @tebeco ^^^

tebeco: (finishing to fix all the test (less than 6), then will read and rewind everything properly.

I'll be able to have a stable branch to start from all test green ;)

tebeco: the 1., if i understand correctly, you are currently working on it
the 3. could only be done after the current PR to be fair as there will be lots of noise
the 4. is totally unrelated to 3. but would require this one

and to be honnest the current PR look like the 2.
as the cross targeting prove the double full compatibility of both stack with their respective dependencies
i just need to update the sample from 2.0 to 2.1
and attempt to use the `UseRouting` is the samples (not in swagger just for integration test so it prouve the double full compat)

tebeco: is that what you had in mind ?

tebeco:
@domaindrivendev 
@pranavkm 
@davidfowl 

(Not in this PR commit as i still tries to understand what is wrong with `Swashbuckle.AspNetCore` running on `AspNetCore3` / `netcoreapp3`)

I got bored of the 6 remaing unit test due to pubinternal and attempted to fork `AspNetCore3.0` and `AspNetCore2.2` here (another branch) based on this PR :
https://github.com/tebeco/Swashbuckle.AspNetCore/tree/fork-aspnetcore-internals/test/Swashbuckle.AspNetCore.SwaggerGen.Test/Fixtures/Fakes
image
https://github.com/tebeco/Swashbuckle.AspNetCore/blob/fork-aspnetcore-internals/test/Swashbuckle.AspNetCore.SwaggerGen.Test/Fixtures/Fakes/FakeApiDescriptionGroupCollectionProvider.cs#L182-217

basicly it's just :
a simple `#define EMPTY` or `define FORK_V2` at top then the creation of the `ModelMetadataProvider`
```
#if NETCOREAPP3_0
#if EMPTY
            return new EmptyModelMetadataProvider();                                                  <===== 6 tests KO
#elif FORK_V2
 //Same code as Swashbuckle, but pointing to class prefixed by FORK_V2_XXX        <===== 5 tests KO
#elif FORK_V3
 //Same code as Swashbuckle, but pointing to class prefixed by FORK_V3_XXX        <===== 5 tests KO
#endif
#else
 //Original code of Swashbuckle                                                                                 <===== all green (normal it's original source targeting 2.x)
#endif
```

the test in diff is `GetSwagger_NamesAllParametersInCamelCase_IfSpecifiedBySettings`
If i understand correctly, it checks for `Attributes` that are "stamped" on public properties of Action Controller parameter :
```
    public class FakeController
    {
        public void AcceptsBindingAnnotatedType(BindingAnnotatedType param)
        { }
    }

    public class BindingAnnotatedType
    {
        public string StringWithNoAttributes { get; set; }

        [BindRequired]
        public string StringWithBindRequired { get; set; }

        [BindRequired]
        public int IntWithBindRequired { get; set; }

        [BindNever]
        public int IntWithBindNever { get; set; }
    }
```

The failling assert does check how many parameters are detected :
https://github.com/tebeco/Swashbuckle.AspNetCore/blob/fork-aspnetcore-internals/test/Swashbuckle.AspNetCore.SwaggerGen.Test/Generator/SwaggerGeneratorTests.cs#L355

* running in 2.1 ==> returns 3                                         OK (because reference)
* running in 3.0 + Empty ==> returns 4                           Looks KO, but `Empty` ... is `Empty`, no idea of the proper repleacement though
* running in 3.0 + FORK_V2 ==> returns 3                       OK
* running in 3.0 + FORK_V3 ==> returns 3                       OK

What's the recommandation here in order to have that behavior back ?

test/WebSites/Basic/Program.cs
tebeco: @domaindrivendev 

https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/1308#issuecomment-542552055

domaindrivendev:
Hi @tebeco - I think a lot of this has progressed with some other PR's which have now been merged. Specifically #1365 and #1377 

So, at this point, I think it's best to close this PR. If you could take a moment to browse through the current state of master and identify what's missing vs your original goals that would be very helpful. If there's still gaps please create an issue with the details and we can go from there. Thankyou so much for all the effort that you put into this.

tebeco:
With pleasure ^^
I’ll try to take a look tomorrow or on Friday

Sorry I took a stop on it waiting for input on your side about the cancellationToken I honestly had no idea what was going on

I also attempted to drop 2.0 in favor of 2.1
There was lots of packages version incompatibility if you follow up both Pr / issue

So I changed the title of the issue to let you release the 5.0.0 as, just like you side it was not a blocker as I though at the beginning.

Thx a lot for the time and the patience.

I’ll try to find time again to dive again in it, as we are re-writing our endpoint routing at work it may help me understand or may be take a look at the 2 or 3 swagger endpoint ^^

PR #1358, opened 2019-11-06
A way of exporting the swagger doc as a yaml

TimmyGilissen:
I have extended the CLI to also support an export of the swagger document to the YAML format.

If you want to deploy to the Azure API management with an arm template. It has to have the YAML format to do it.

https://azure.microsoft.com/en-us/blog/announcing-the-preview-of-openapi-specification-v3-support-in-azure-api-management/


domaindrivendev:
This looks great - but for completeness, I think it would be good to also include this capability with the middleware that exposes the Swagger./OpenApi over HTTP. This could be based on the `Accept` header that's passed - e.g. `text/yaml` vs `text/json` vs `application/json` and would require some relatively trivial updates to the following:

https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/master/src/Swashbuckle.AspNetCore.Swagger/SwaggerMiddleware.cs

I know you probably don't need this but it would make the PR a little more complete from an overall project standpoint.

TimmyGilissen:
Is this the adjusment you wanted to see in the middleware ?

TimmyGilissen:
I have reverted the commits back to the working version. 

There is more work to be done just by updating the swagger middleware by looking at the HTTP headers.

By default, the path is set to export a JSON file, if you want to be able to export a YAML file you need to update the path so the file is exported to be a YAML file. 

I think a more explicit way is a better approach. When you configure the SwaggerGen you explicitly say you want to work with a YAML setup.

This makes it more clear what you are working with. Instead of using the header notation.

domaindrivendev:
I disagree - to ensure the most predictable behavior, HTTP middleware should honor the HTTP protocol, including header-based content negotiation. With that said, I know this was a little beyond what you originally bargained for. For now, it probably makes most sense to merge your original enhancement and I'll address the middleware as a separate issue.

PR #1371, opened 2019-11-14
Consolidate XUnit to 2.4.1

say25:
Currently both XUnit 2.2.0 and 2.4.0 are pulled in. 
Simply pull in exclusively 2.4.1.


domaindrivendev:
👍

PR #1396, opened 2019-11-26
Target netstandard2.0 + netcoreapp3.0 (per MS guidelines)

### LINKED ISSUE #1301: Security issue - please update Microsoft.AspNetCore.Mvc.Core to 2.0.1
ASP.NET Core 2.0 allows an attacker to steal log-in session information such as cookies or authentication tokens via a specially crafted URL aka "ASP.NET Core Elevation Of Privilege Vulnerability".
Please update Microsoft.AspNetCore.Mvc.Core to 2.0.1 to avoid this issue.

domaindrivendev:
Target netstandard2.0 + netcoreapp3.0 for maximum compatibility as per https://docs.microsoft.com/en-us/aspnet/core/migration/22-to-30?view=aspnetcore-3.0&tabs=visual-studio. Apply to all src projects for consistency.

Also fixes #1301


domaindrivendev:
@danielcrenna @tebeco. This is what I'm considering moving forward with. The MS guidance referenced above indicates that both `netstandard2.0` and `netcoreapp3.0` should be targeted. Technically, this is only necessary in the projects that leverage compiler flags to address breaking changes between ASP.NET Core 2 and 3. However, to keep things consistent, I've decided to multi-target these frameworks in all of the "src" projects. Do you guys see any issue with this approach?

tebeco:
So the original PR opened 1 month ago right ^^ ?

domaindrivendev:
@tebeco ??? Not sure what you mean

tebeco:
https://github.com/domaindrivendev/Swashbuckle.AspNetCore/pull/1316

domaindrivendev:
If this has ended up being very similar to the original PR that you posted way back when then apologies for all the back and forth. Let's call it a journey that I needed to go through to get up to speed on all this 😉 

So, I'm assuming you're good with the approach put forward in this PR?

tebeco:
No worries
I had extrem trouble explaining thing when I write
I’m mostly in cause here

Also was out/disconnected from github for few weeks so that did not helped ^^

And the best thing is that the fact you learnt this guidance on your side makes you completely aware of the how what why
At the end it’s may be even better

danielcrenna:
@domaindrivendev looks good to me.

PR #1398, opened 2019-11-27
Support IAsyncEnumerable if STJ serializer is being used

### LINKED ISSUE #1317: Controller actions returning IAsyncEnumerable<T> are not documented correctly
When an API controller action returns `IAsyncEnumerable<SomeModel>` its result is not documented correctly in the resulting API spec.

For example, this code
```C#
public IAsyncEnumerable<SomeModel> GetSomethingAsync()
{
    return _someService.GetAllAsync();
}
```

results in this schema:
```JSON
...
      "SomeModelIAsyncEnumerable": {
        "type": "object",
        "additionalProperties": false
      },
...
```

As you can see, all information about the model are missing here. I'd expect that API actions that return `IAsyncEnumerable<T>` are documented the same way as when they are when they return `ICollection<T>`.

Version `5.0.0-rc4` with `ASP.Net Core 3.0`

domaindrivendev:
Fixes #1317

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/TypeExtensions.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/TypeExtensions.cs
@@ -43,6 +43,14 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
                 return true;
             }
 
+#if NETCOREAPP3_0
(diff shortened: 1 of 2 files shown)


tsproesser:
Newtonsoft supports this type, too. Is this going to happen any time soon? I See this issue is **pretty** old already.

PR #1405, opened 2019-12-02
Add support to PKCE for SwaggerUI & update OAuth2Integration sample #999

### LINKED ISSUE #999: Support OAuth 2 Authorization Code Flow with PKCE
Since there are now recommendations to avoid using Implicit flow and for (Authorization) code flow to be secure PKCE must be used it would be nice if the interface supported that. 

More information here: https://brockallen.com/2019/01/03/the-state-of-the-implicit-flow-in-oauth2/ 

As well as a javascript client that does support it: https://github.com/IdentityModel/oidc-client-js/releases/tag/1.6.0

lurumad:
### Description
- Added support for [PKCE](https://tools.ietf.org/html/rfc7636) with the Autorization Code flow.

### Motivation and Context
- This makes the Authorization Code flow more secure because the code can only be used by the client that provided the code_challenge on the /authorize request. It's the recommended way for [Browser-Based Applications](https://www.identityserver.com/articles/change-in-recommendations-for-browser-based-applications)

Fixes #999 

### How Has This Been Tested?

- I've update the [OAuth2IntegrationSample](https://github.com/lurumad/Swashbuckle.AspNetCore/tree/feature/pkce/test/WebSites/OAuth2Integration) flow to use an interactive confidential client (Code with PKCE).

Regards!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,18 +1,3 @@
 {
-  "requires": true,
-  "lockfileVersion": 1,
-  "dependencies": {
-    "swashbuckle": {
-      "version": "file:src/Swashbuckle.AspNetCore.SwaggerUI",
-      "requires": {
-        "swagger-ui-dist": "3.19.5"
-      },
-      "dependencies": {
-        "swagger-ui-dist": {
-          "version": "3.19.5",
-          "bundled": true
-        }
-      }
-    }
-  }
+  "lockfileVersion": 1
 }
--- a/src/Swashbuckle.AspNetCore.SwaggerUI/SwaggerUIOptions.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerUI/SwaggerUIOptions.cs
@@ -203,5 +203,10 @@ namespace Swashbuckle.AspNetCore.SwaggerUI
         /// (Authorization header with Basic base64encode(client_id + client_secret))
         /// </summary>
         public bool UseBasicAuthenticationWithAccessCodeGrant { get; set; } = false;
+
+        /// <summary>
+        /// Enabled to use PKCE with the Autorization Code flow.
+        /// </summary>
+        public bool UsePkceWithAuthorizationCodeGrant { get; set; } = false;
     }
 }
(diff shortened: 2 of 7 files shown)


lurumad:
Hi!

Any news about to merge this PR? We need this support in many applications based on security advisors.

Kind Regards!

fuzzzerd:
I'd love to see this, or a variation of it in the v5 final release.

ardumez:
Hello, I have error with your sample

> Access to fetch at 'https://login.microsoftonline.com/xxx/oauth2/v2.0/token' from origin 'https://localhost:44328' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.

I have add cors allowAnyOrigin() and add c.OAuthUsePkce();

lurumad:
Hi @ardumez 

There is no way to configure Allowed Origins in Azure AD.

https://feedback.azure.com/forums/169401-azure-active-directory/suggestions/37689847-cors-for-token-endpoint

Regards!

mihalios:
Hi @lurumad, thanks for your work on this. Do you know which version of Swashbuckle this is in?

lurumad:
Hi @mihalios 

I'm using 5.3.2 in my projects and it's available. Also I wrote a blog post about how to configure PKCE in Swashbuckle ASP.NET Core:

https://lurumad.github.io/swagger-ui-with-pkce-using-swashbuckle-asp-net-core

The repo with the example:

https://github.com/lurumad/swagger-ui-pkce

Regards!

mihalios:
Thanks @lurumad. Small note, if you have PKCE enabled and you're specifying a client id, then specifying a client secret is unnecessary (in fact, it's bad).

lurumad:
Hi @mihalios 

I know it's unncessary but it's another way more to identify the client, but can you refer to some documentation saying that it's a bad practice? 

Regards!

mihalios:
@lurumad The point of using Authorization Code Flow with PKCE is that you don't use a static secret, that gets stored, inherently insecurely, on the client. With PKCE, behind the hood, the client creates itself an ephemeral secret, the code verifier. I've been using these docs https://auth0.com/docs/flows/concepts/auth-code-pkce

lurumad:
The real secret is the code_verifier that is hashed into code_challenge, for this reason I see the secret like client_id or similar, there aren't any security reason like for example in auhtorization code flow without pkce that is a real secret and for these reason before PKCE you used to use implicit because you can't keep a secret in public clients, it isn't necessary but it's not a security flaw.

mihalios:
@lurumad I think we're on the same page! I was just looking at your tutorial (very helpful and appreciated) where you setup PKCE but still make this call:

`setup.OAuthClientSecret(appSettings.Security.Jwt.ClientSecret);`

That can be removed.

Issue #1420, opened 2019-12-13
Differences in generated OpenAPI schema between rc4 and rc5


cremor:
I've found a few differences in my OpenAPI schema between rc4 and rc5:

1. All my GET parameters are now documented as "nullable: true" (previously the nullable attribute was not present for the parameters in the OpenAPI schema).
  a. Same as for properties discussed in #1346 before, required parameters are now both required and nullable. Since required properties are never nullable after #1346 required parameters shouldn't be either.
  b. Even value type parameters like int32 are marked as nullable, but those only if they are not required. I'm honestly not sure if this is a problem or fine. Does it make sense to pass a not required query string parameter as null? How would I even explicitly pass null to a query string parameter?
2. As already reported by @danzel in #1346 all response objects (arrays directly in the response "content" of methods in the schema, objects in the "components" part of the schema) are now marked as nullable. I assume this is technically true, but still a bit weird. Where would I need to put attributes to tell Swashbuckle that my response objects will never be null?
3. All my models in "components" now don't have the "additionalProperties: false" attribute any more. According to the specification the default value of that is true, so I don't think this is correct?
4. The `ProblemDetails` schema has a difference in it's "additionalProperties". Previously it had a "type: object" inline. Now it references a separate "Object" schema. Maybe not a problem by itself, but this causes an empty "Object" schema to be shown in the Swagger UI which is weird.

Versions:
Swashbuckle.AspNetCore 5.0.0-rc5
Swashbuckle.AspNetCore.Newtonsoft 5.0.0-rc5
ASP.NET Core 2.2 running on the .NET Framework runtime

domaindrivendev:
Ugghhhh - this is no good. Most of them are unanticipated side-effects of setting Nullable for _all_ reference or nullable types, something that seemed like the right thing to do. Although, I'm confused as to why you would see value types with the Nullable flag set also, I'm not seeing this behavior.

Inline vs reference for `object` type really shouldn't make any difference - and it's "empty" either way. I switched to referenced schema's for this type for consistency so that _all_ object schema's are now referenced.

TBH, supporting this `Nullable` flag has been so much hassle and impossible to get right. As you're finding out, whether or not it makese sense to set is highly contextual (parameter vs response, parameter vs property etc.). However, a lot of the generated schema's are "referenced" and therefore could be re-used across parameters, responses, properties etc.

[...]

PR #1429, opened 2019-12-19
Only assign nullable for property schemas

domaindrivendev:
Addresses #1420 with the following changes:

1. Only assign `nullable` for property schemas, and in that case, set the flag if the underlying property type is a reference type or a `Nullable<T>`

2. By default, set `additionalProperties: false` for object schemas.

3. Return an _inline_ object schema for parameters or properties of type `System.Object`
 
4. For _complex type_ properties, rather than returning a "reference" schema that points to a shared definition (under `#/components/schemas`), return an inline schema that "extends" the shared definition:

For example, instead of:
```json
{
  "type": "object",
  "properties": {  
    "prop1": {
      "$ref": "#/components/schemas/SomeType"
    }
}
```

Return ...
```json
{
  "type": "object",
  "properties": {  
    "prop1": {
      "allOf": [
        { "$ref": "#/components/schemas/SomeType" }
      ]
  }
}
```

This way, contextual metadata (e.g. property attributes, property description, nullable etc.) can be applied to the "extended" schema (as it's not shared). Previously, contextual metadata was simply omitted for any properties that had a reference type.

See https://github.com/OAI/OpenAPI-Specification/issues/1368 for the rational behind using `allOf` for _complex type_ properties.


sabbadino:
Hi, when the next release, containing this "fix", is planned ? 
Thank you

PR #1463, opened 2020-01-09
Restore support for multiple SwaggerUI and ReDoc urls

joelving:
The changes to SwaggerUIBuilderExtensions.cs and ReDocBuilderExtensions.cs in #1386 made it impossible to register more than one endpoint from which to serve SwaggerUI and ReDoc. This PR undoes the changes and adds a test website to validate the functionality (and showcase the use case).

Fetching an IOptions<T> from the service provider would always return the same (initially default) instance of SwaggerUIOptions/ReDocOptions respectively, meaning every instance of middleware would operate on the same instance of the options object. As a result, subsequent setup actions would overwrite the options object from previous registrations.

### DIFF
--- a/Swashbuckle.AspNetCore.sln
+++ b/Swashbuckle.AspNetCore.sln
@@ -82,7 +82,9 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "NetCore21", "test\WebSites\
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Swashbuckle.AspNetCore.Newtonsoft", "src\Swashbuckle.AspNetCore.Newtonsoft\Swashbuckle.AspNetCore.Newtonsoft.csproj", "{0783E72E-3483-43AF-AE4B-2A6CFAFD3DD3}"
 EndProject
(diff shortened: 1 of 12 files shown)


domaindrivendev:
We can't just revert the code as that was introduced to address a different problem. Ideally, we would find a solution that addresses both problems. @andyalm - interested to get your thoughts? While the changes you introduced provides better support for different IOC containers (I believe Autofac was one that had issue with the previous implementation), it also prevents the ability to configure multiple instances of the SwaggerUI middleware independently.

joelving:
Alright. Am I understanding the use case of the original issue correctly, that some may prefer customizing SwaggerUIOptions/ReDocOptions via their DI container of choice? Is it thinkable to **both** supply options via DI **and** customize them via a setup action?
If the answers are yes and no, respectively, could a solution be to fetch options from DI if setup action is null and otherwise invoke the method on a brand new options object.

Something along the lines of:
```
if (setupAction == null)
{
    // Fetch options from DI container
    var options = app.ApplicationServices.GetService<IOptions<ReDocOptions>>()?.Value ?? new ReDocOptions();
    app.UseMiddleware<ReDocMiddleware>(options);
}
else
{
    // Configure an options instance here and pass directly to the middleware
    var options = new ReDocOptions();
    setupAction.Invoke(options);

    app.UseMiddleware<ReDocMiddleware>(options);
}
```

or maybe slightly terser:
```
var options = setupAction == null ? app.ApplicationServices.GetService<IOptions<ReDocOptions>>()?.Value ?? new ReDocOptions() : new ReDocOptions();
setupAction?.Invoke(options);
app.UseMiddleware<ReDocMiddleware>(options);
```

Of course, this won't work if people expect to be able to inject defaults via the DI container and customize via setup actions.

domaindrivendev:
There's defintitely a regression issue (currently being tracked in #1559) that was introduced by #1386. However, at this point I don't want to simply revert that code as it too was solving a specific problem. So, I'm going to close this as and start discussing potential solutions in #1559.

PR #1468, opened 2020-01-10
Support oneOf, allOf & discriminator for polymorphic hierarchies

domaindrivendev:


the-vampiire:
can you point me where in the docs it describes how to use oneOf / allOf? i see this support was merged into master but cant find it in the docs

AdamDiament:
> can you point me where in the docs it describes how to use oneOf / allOf? i see this support was merged into master but cant find it in the docs

The method swaggerGenOptions.GeneratePolymorphicSchemas in 5.4 1 appears to still only support oneOf - @domaindrivendev is there a way to get it to use all of?

domaindrivendev:
The feature is still in an experimental phase hence why it’s not documented yet.

With that said it should already use oneOf when describing a base class (i.e. to list the known subtypes) and allOf when describing a subclass (i.e. to specify the base type).

Can you elaborate your use case please? What is your C# class hierarchy and how do you want it to be represented in Swagger using allOf?

gpgpublickey:
Is this still experimental? @domaindrivendev could you please provide documentation ?

PR #1475, opened 2020-01-14
Additional readme updates to coincide with 5.0.0 releases

DamienBraillard:
So happy with the 5.0.0 release ! Thanks !

While looking at it, I found that the code examples were still refering the 5.0.0-rc5 version rather than the 5.0.0 version.

Here are the fixes to the readme file to fix this.


domaindrivendev:
Awesome.! Thankyou

PR #1480, opened 2020-01-15
Add document name to document and operation filter contexts.

Blackbaud-DanWard:
We generate multiple swagger documents from a single service that are pre-filtered and formatted for specific consumers. Currently we do this with a (large) `PreSerializeFilters`, but document and operation filters would be much cleaner. We can exclude paths entirely with `DocInclusionPredicate`, but we can't change their shape.

In order to know which document we're processing for, `documentName` seems reasonable in the filter context(s).

Comments/suggestions welcome.

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/IDocumentFilter.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/IDocumentFilter.cs
@@ -14,11 +14,13 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
         public DocumentFilterContext(
             IEnumerable<ApiDescription> apiDescriptions,
             ISchemaGenerator schemaGenerator,
-            SchemaRepository schemaRepository)
+            SchemaRepository schemaRepository,
+            string documentName = null)
         {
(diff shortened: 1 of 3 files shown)


src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/IDocumentFilter.cs
```diff
@@ -14,17 +14,21 @@ public class DocumentFilterContext
         public DocumentFilterContext(
             IEnumerable<ApiDescription> apiDescriptions,
             ISchemaGenerator schemaGenerator,
-            SchemaRepository schemaRepository)
+            SchemaRepository schemaRepository,
+            string documentName = null)
```
Blackbaud-DanWard: Using a default here to prevent breaking existing code, but I doubt consumers are ever constructing these types themselves. If you'd prefer an overload, that's certainly an option, too.

Blackbaud-DanWard:
@domaindrivendev is this something you'd have any interest in?
Happy to go a different direction or help with a parallel effort if there's overlap.

Blackbaud-ChristiSchneider:
This can be closed without merging as it was included in #1922

PR #1481, opened 2020-01-15
Provide capability to ignore existing SwaggerOptions/SwaggerUiOptions

tomkerkhove:
Provide capability to ignore existing SwaggerOptions/SwaggerUiOptions which would allow you to expose different Swagger endpoints & UI URLs.

## Usecase

I've added OpenAPI 3.0 support to https://github.com/tomkerkhove/promitor but still support the old Swagger 2.0 approach.

With this approach, we have:
- Swagger UI 2.0 on `/swagger/index.html` which serves these Swagger doc endpoints
  - Swagger 2.0 on `/swagger/v1/swagger.json` (serialized as v2)
  - OpenAPI 3.0 on `/api/v1/docs.json`
- OpenAPI UI 3.0 on `/api/docs/index.html` which serves these OpenAPI doc endpoints
  - OpenAPI 3.0 on `/api/v1/docs.json`

This allows me to have the new approach side-by-side with the deprecated approach.

## Issue

When we're doing this:
```csharp
app.UseOpenApi(setupAction => setupAction.RouteTemplate = "api/{documentName}/docs.json");
app.UseOpenApi(setupAction => setupAction.SerializeAsV2 = true);
```

Our second call will use the configuration from the first call and the end result would be API docs being served on `api/{documentName}/docs.json` serialized as V2 instead of seperate endpoints.

With the introduction of `ignoreExistingOptions` we could explicitly say we don't want that. (but I'm open for better names) Same issue with respect to `UseSwaggerUI`.

## More information

- I've forked the bootstrapping extensions in my PR to fix this for my scenario, this might help explain why this is important. https://github.com/tomkerkhove/promitor/pull/841
- This might be in conflict/related to https://github.com/domaindrivendev/Swashbuckle.AspNetCore/pull/1463

### DIFF
--- a/src/Swashbuckle.AspNetCore.ReDoc/ReDocBuilderExtensions.cs
+++ b/src/Swashbuckle.AspNetCore.ReDoc/ReDocBuilderExtensions.cs
@@ -11,8 +11,16 @@ namespace Microsoft.AspNetCore.Builder
             this IApplicationBuilder app,
             Action<ReDocOptions> setupAction = null)
         {
-            var options = app.ApplicationServices.GetService<IOptions<ReDocOptions>>()?.Value ?? new ReDocOptions();
-            setupAction?.Invoke(options);
+            var options = new ReDocOptions();
+            if (setupAction != null)
+            {
+                setupAction(options);
+            }
+            else
+            {
+                options = app.ApplicationServices.GetRequiredService<IOptions<ReDocOptions>>().Value;
+            }
+
             app.UseMiddleware<ReDocMiddleware>(options);
 
             return app;
--- a/src/Swashbuckle.AspNetCore.Swagger/DependencyInjection/SwaggerBuilderExtensions.cs
+++ b/src/Swashbuckle.AspNetCore.Swagger/DependencyInjection/SwaggerBuilderExtensions.cs
@@ -9,9 +9,17 @@ namespace Microsoft.AspNetCore.Builder
     {
         public static IApplicationBuilder UseSwagger(
             this IApplicationBuilder app,
-            Action<SwaggerOptions> setupAction = null)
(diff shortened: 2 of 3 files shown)


tomkerkhove:
I have not added any tests as I'm seeking input on how you'd prefer to have those, just unit tests or?

tomkerkhove:
Any feedback on this @domaindrivendev.

domaindrivendev:
For tracking, the underlying issue is #1559 

@tomkerkhove I think you're on the right track, but rather than introducing an additional optional parameter, I think it would make more sense to keep the contract intact and simply base the logic off the presence of an explicit setup action. Something along the following lines:

```csharp
public static IApplicationBuilder UseSwaggerUI(
    this IApplicationBuilder app,
    Action<SwaggerUIOptions> setupAction = null)
{
    if (setupAction == null)
    {
        var options = new SwaggerUIOptions();
        setupAction(options);
        app.UseMiddleware<SwaggerUIMiddleware>(options);
    }
    else
    {
        var options = app.ApplicationServices.GetRequiredService<IOptions<SwaggerUIOptions>>().Value;
        app.UseMiddleware<SwaggerUIMiddleware>(options);
    }

    return app;
}
```

domaindrivendev:
I you want to update the PR with this change, I'd be happy to merge it. Also, could you please do the same for the `ReDoc` middleware as well? Alternatively, if you don't have bandwidth, please let me know ASAP and I can create a new PR for it.

tomkerkhove:
I'm more than happy to change it and will do it this/next week, ok?

tomkerkhove:
Everything should be updated @domaindrivendev!

Any tips on how to add tests or not required?

PR #1490, opened 2020-01-17
Updated projects to .net core v3.1

waxtell:
- Updated conditional compiler directives to NETCOREAPP3_1
- Updated framework dependent dependencies to 3.1.x
- Set version shared property to 6.0.0-beta
- All tests pass
- Local regression tests revealed no issues


domaindrivendev:
What's your rationale for this PR? As a shared library that's used in many applications on a wide range of ASP.NET framework versions, I think it's far better (and general best practice) to target the lowest version of dependencies possible unless there's a good reason not to (e.g. if the library itself depends on new features of the dependency).

If I target 3.1, then applications running 3.0 may encounter compatability issues.

waxtell:
Totally fair...  I was merely trying to update to target the LTS version of .net core, but you're correct; twill probably cause more problems than it will cure.

Thanks for your time and consideration!

PR #1519, opened 2020-02-04
Improve error feedback regarding the use of empty enums

fernando-almeida:
I came across an unexpected issue when upgrading to the latest version of Swashbuckle.AspNetCore where I had unintentionally defined an empty enum which led to the triggering of a non-obvious error by the generator. I've included a proposed fix to make the exception meaningful.


domaindrivendev:
Master has diverged significantly and the generator will no longer crash when it encounters an empty enum

PR #1535, opened 2020-02-20
Add netcoreapp3.1 to SwaggerUI

BlythMeister:
Superseeds #1533

### DIFF
--- a/src/Swashbuckle.AspNetCore.ApiTesting/Swashbuckle.AspNetCore.ApiTesting.csproj
+++ b/src/Swashbuckle.AspNetCore.ApiTesting/Swashbuckle.AspNetCore.ApiTesting.csproj
@@ -4,7 +4,7 @@
 
   <PropertyGroup>
     <Description>Core components for Swagger/OpenAPI-driven integration testing API's built on ASP.NET Core</Description>
-    <TargetFrameworks>netstandard2.0;netcoreapp3.0</TargetFrameworks>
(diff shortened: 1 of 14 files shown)


BlythMeister:
@domaindrivendev without this, it seems impossible to add swagger to a netcore3.1 application :(
so is pretty high importance.

When adding swagger to netcore3.1 you get the netstandard2 flavour, which means it starting bringing in the nuget packages

domaindrivendev:
NetCore 3.1 is backwards compatible and therefore should have no problem working with a lib that targets netcoreapp3.0, which SB already does.

I’ve been able to get SB working with no issue on both a 3.0 and 3.1 application. In other words, I can’t repro the issue this PR is trying to solve.

Could we take a step back from solutioning and try get a common repro first? Perhaps you could create a minimal project from scratch that demonstrates the incompatibility you’re describing, and then post it to GitHub where I can pull down and hopefully see for myself.

BlythMeister:
When using paket I get all sort of dependency issues (in fact, it's unresolvable)
Unfortunately this is in a repo I can't share publicly.

With this change, I can force to only make dependency tree for netcore 3.1.
If I limit to netcore3 paket doesn't add the dependency....now again, that might be a paket thing?!? I'm not sure.

BlythMeister:
hmm i'm an idiot.  My wordaround is to tell paket only use this for netcore3.0 & netcore3.1
that way, it uses the 3.0 flavour in a 3.1 app.

Gawd i hate that `Microsoft.AspNetCore.App` is now a framework reference!

PR #1547, opened 2020-03-03
Add an option to specify the serialize format version per document

AlbertoSadoc:
Added an overloaded method that specifies the serialize version when adding a document:
```services.AddSwaggerGen(c =>
{
    c.SwaggerDoc("v1", new OpenApiInfo { Title = "Test", Version = "1" }, SerializeVersion.V2);
});
```
The previous method serializes as V3 and the `SwaggerOptions.SerializeAsV2` has been marked as obsolete.
This change allows to have different versions of the documents that serialize in different formats.

### DIFF
--- a/src/Swashbuckle.AspNetCore.Cli/Program.cs
+++ b/src/Swashbuckle.AspNetCore.Cli/Program.cs
@@ -91,9 +91,9 @@ namespace Swashbuckle.AspNetCore.Cli
                             writer = new OpenApiJsonWriter(streamWriter);
 
                         if (namedArgs.ContainsKey("--serializeasv2"))
-                            swagger.SerializeAsV2(writer);
+                            swagger.Item1.SerializeAsV2(writer);
                         else
-                            swagger.SerializeAsV3(writer);
+                            swagger.Item1.SerializeAsV3(writer);
 
                         if (outputPath != null)
                             Console.WriteLine($"Swagger JSON/YAML succesfully written to {outputPath}");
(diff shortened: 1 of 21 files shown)


domaindrivendev:
I appreciate the problem that you're trying to solve but I really want to keep the generation functionality (i.e. generating a v3 `OpenApiDocument` from your routes, controllers and models) completely separate from the functionality to expose that document as serialized JSON (i.e. in the v2 or v3 format) and I feel the approach you've taken here couples the two concerns together a little to tightly.

domaindrivendev:
I think I'd prefer to see the capability you're looking for somehow incorporated into the middleware options (i.e. `SwaggerOptions`) rather than the generator options

domaindrivendev:
In fact, now that I think about it, there's already a very clean way to solve your problem without any changes. You can simply insert two instances of the Swagger middleware:

```csharp
// Configure the generator with a single document
services.AddSwaggerGen(c =>
{
    c.SwaggerDoc("v1", new OpenApiInfo { Title = "Test API", Version = "v1" });
});
```
```csharp
// Configure two instances of the Swagger middleware
app.UseSwagger(c =>
{
    // The following route will produce the requested document in the v3 format
    c.RouteTemplate = "openapi/{documentName}/openapi.json";
});

app.UseSwagger(c =>
{
    // The following route will produce the requested document in the v3 format
    c.RouteTemplate = "swagger/{documentName}/swagger.json";
    c.SerializeAsV2 = true;
});
```

AlbertoSadoc:
Hi, thank you for the comment. I actually had to update to v5.2.1 to get this working, it didn't on v5.0.0.

It's all very neat like this. Thank you!

PR #1551, opened 2020-03-04
Add support for open generics in config.MapType

### LINKED ISSUE #560: config.MapType does not seem to work with open generics
When configuring a type override for swagger docs display, open generic types don't seem to work.

```
options.MapType(typeof(Id<>), () => new Schema {Type = "string", Description = "uuid"});
```

A parameter of type `Id<Foo>[]` is still displayed as a body of type `Array[Id[Foo] {}]` with an example value of
```
[
  {}
]
```

ybeauchamph:
Fixes #560


triforcely:
Would that be possible to add an overload so schema factory function receives "Type" as a parameter?

ybeauchamph:
Fixed in #1600

PR #1576, opened 2020-03-18
Null reference exception: MemberInfo can be null

wallymathieu:
I've added some nullreference exceptions with nameof in order to get better error messages. I've fixed the issue of when MemberInfo is null.


src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/MemberInfoExtensions.cs
```diff
@@ -2,16 +2,18 @@
...
 using Microsoft.AspNetCore.Mvc;
-
+using System;
 namespace Swashbuckle.AspNetCore.SwaggerGen
 {
     public static class MemberInfoExtensions
     {
         public static IEnumerable<object> GetInlineOrMetadataTypeAttributes(this MemberInfo memberInfo)
         {
+            if (memberInfo==null) throw new NullReferenceException(nameof(memberInfo));
```
domaindrivendev: As a general rule, I prefer to put the responsibility of null checking on the extension method caller rather than inside the method itself. This is the way it is for all other instance methods and I prefer to be consistent in the case of extension methods aswell.

wallymathieu: So instead I should return empty enumerable in order to be consistent?

src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/MemberInfoExtensions.cs
```diff
@@ -2,16 +2,18 @@
...
             var metadataTypeAttribute = memberInfo.DeclaringType.GetCustomAttributes(true)
                 .OfType<ModelMetadataTypeAttribute>()
                 .FirstOrDefault();
```
domaindrivendev: If `DeclaringType` is null, then the `ModelMetadataTypeAttribute` simply isn't applicable and so we should stick to returning the original member attributes rather than throwing an exception. I think the only code change neccessary for this is to postfix the reference to the `DeclaringType` property above with the safe navigation operator (i.e "?").

src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/SchemaGenerator.cs
```diff
@@ -249,7 +250,7 @@ private void ApplyMemberMetadata(OpenApiSchema schema, Type type, MemberInfo mem
                 schema.Reference = null;
             }
 
-            if (schema.Reference == null)
+            if (schema.Reference == null && memberInfo != null)
```
domaindrivendev: I believe this additional check is unnecessary as the enclosing method `ApplyMemberMetadata` is only called on the condition that memberInfo is not null.

src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/MemberInfoExtensions.cs
wallymathieu: Or return `Enumerable.Empty<object>()` ?

src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/SchemaGenerator.cs
```diff
@@ -195,6 +195,7 @@ private OpenApiSchema GenerateObjectSchema(SerializerMetadata serializerMetadata
 
             foreach (var serializerPropertyMetadata in serializerMetadata.Properties)
             {
+                if (serializerPropertyMetadata.MemberInfo==null) continue;
```
domaindrivendev: Rather than skipping the iteration, I think the logic can be a bit more graceful by continuing with an empty list of custom attributes when `MemberInfo` is null. So, the line below could be replaced with something along the following lines:

```csharp
var customAttributes = (serializerPropertyMetadata.MemberInfo != null)
    ? serializerPropertyMetadata.MemberInfo.GetInlineOrMetadataTypeAttributes()
    : Enumerable.Empty<object>();
```

wallymathieu: this is probably the important bit, need to verify on a webapp where memberinfo is null though

src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/MemberInfoExtensions.cs
domaindrivendev: I believe the fallback call to `memberInfo.GetCustomAttributes(true)` below will do that automatically

src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/MemberInfoExtensions.cs
```diff
@@ -2,22 +2,22 @@
...
         {
-            var metadataTypeAttribute = memberInfo.DeclaringType.GetCustomAttributes(true)
+            var metadataTypeAttribute = memberInfo?.DeclaringType.GetCustomAttributes(true)
                 .OfType<ModelMetadataTypeAttribute>()
                 .FirstOrDefault();
 
-            var metadataTypeMemberInfo = metadataTypeAttribute?.MetadataType.GetMember(memberInfo.Name)
-                .FirstOrDefault();
+            var metadataTypeMemberInfo = metadataTypeAttribute?.MetadataType.GetMember(memberInfo?.Name)
+                ?.FirstOrDefault();
```
domaindrivendev: I'm fairly certain these additions are all redundant

src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/MemberInfoExtensions.cs
```diff
@@ -2,22 +2,22 @@
...
-
+using System;
 namespace Swashbuckle.AspNetCore.SwaggerGen
 {
     public static class MemberInfoExtensions
     {
         public static IEnumerable<object> GetInlineOrMetadataTypeAttributes(this MemberInfo memberInfo)
         {
-            var metadataTypeAttribute = memberInfo.DeclaringType.GetCustomAttributes(true)
+            var metadataTypeAttribute = memberInfo?.DeclaringType.GetCustomAttributes(true)
```
domaindrivendev: Like I said, I don't want to null check the extension instance inside the method itself. Rather, I was suggesting the following:

```csharp
var metadataTypeAttribute = memberInfo.DeclaringType?.GetCustomAttributes(true)
```

src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/MemberInfoExtensions.cs
domaindrivendev: My pref is to avoid null checking the extended instance (i.e. `memberInfo`) from within the extension method. I prefer this responsibility to be on the caller as is the case with regular instance methods.

src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/MemberInfoExtensions.cs
```diff
@@ -2,22 +2,22 @@
...
                 .FirstOrDefault();
 
-            var metadataTypeMemberInfo = metadataTypeAttribute?.MetadataType.GetMember(memberInfo.Name)
-                .FirstOrDefault();
+            var metadataTypeMemberInfo = metadataTypeAttribute?.MetadataType.GetMember(memberInfo?.Name)
+                ?.FirstOrDefault();
 
             return (metadataTypeMemberInfo == null)
-                ? memberInfo.GetCustomAttributes(true)
+                ? memberInfo?.GetCustomAttributes(true)
```
domaindrivendev: My pref is to avoid null checking the extended instance (i.e. `memberInfo`) from within the extension method. I prefer this responsibility to be on the caller as is the case with regular instance methods.

src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/SchemaGenerator.cs
domaindrivendev: This is where I'd prefer the null check to happen rather than within the extension method

version.props
```diff
@@ -1,5 +1,5 @@
 <Project>
   <PropertyGroup>
-    <VersionPrefix>5.2.0</VersionPrefix>
+		<VersionPrefix>5.2.0</VersionPrefix>
```
domaindrivendev: There should ne no change to this file

src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/SchemaGenerator.cs
wallymathieu: I've moved it to schema generator, what's your thought about how it looks now?

version.props
wallymathieu: Should be fixed now

wallymathieu:
Looks like there is really only a need for one small change, should I do a rebase in order to squash the above commits ?

domaindrivendev:
Thx for the contribution @wallymathieu 👍

wallymathieu:
Thanks for the feedback! @domaindrivendev

PR #1591, opened 2020-03-24
Define enum description

FalconWu2017:
Define Enum description by DescriptionAttribute or summary.


domaindrivendev:
I don't really understand the problem this solves. Additionally, this PR is now out-of-date (my bad not yours) and there's conflicts so I'm going to close for now.

If your problem is still valid, please start with an issue that clearly articulates it with uses cases and code examples - then we can start from there and figure out a suitable path to resolution. Thx

PR #1603, opened 2020-03-27
Update README.md example for setting custom swagger path

tristanjthompson:
Thanks for your project!  When using the example in the README to set a custom swagger path for the UI, I found that it didn't work.  I found that an extra line of config was necessary to get it working.

This is a one-line PR to add in the missing line I needed to get custom paths working.

### DIFF
--- a/README.md
+++ b/README.md
@@ -252,6 +252,8 @@ app.UseSwaggerUI(c =>
 })
 ```
 
+_NOTE: If you also need to update the relative path that the UI itself is available on, you'll need to follow the instructions found in [Change Relative Path to the UI](#change-relative-path-to-the-ui)._
+
 ### Modify Swagger with Request Context ###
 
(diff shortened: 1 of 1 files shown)


domaindrivendev:
Maybe this is a flaw in the way the docs are structured but (I'm planning an overhaul if I could ever get around to it), but the setting you added adjusts the path for the UI itself as opposed to the OpenAPI/Swagger JSON endpoint. The two settings are independent of eachother. There's a separate section in the docs [here](https://github.com/domaindrivendev/Swashbuckle.AspNetCore#change-relative-path-to-the-ui) describing the setting you added.

tristanjthompson:
Ah yes, that makes sense, thank for explaining!

I totally missed that when I was trying to implement this!  I've updated the PR to remove the line of code I added and instead add a note with a link to that docs section if you think that's helpful?

PR #1615, opened 2020-04-06
Support IHost in CLI

### LINKED ISSUE #1614: Custom `IHost` Configuration (not `IWebHost`)
Repro:
1. Create ASP.NET Core 3.1 web application, just like it is suggested using `IHost`

https://docs.microsoft.com/en-us/aspnet/core/fundamentals/host/web-host?view=aspnetcore-3.1
"This article covers the Web Host, which remains available only for backward compatibility. The Generic Host is recommended for all app types."

2. Follow the Swashbuckle documentation and setup "Custom IWebHost Configuration"
https://docs.microsoft.com/en-us/aspnet/core/fundamentals/host/web-host?view=aspnetcore-3.1

Actual behavior:
Interfaces are incompatible, cannot use custom host. Usage of IHost fails with 
`Unhandled exception. System.InvalidOperationException: SwaggerWebHostFactory class detected but does not contain a public static method called CreateWebHost with return type IWebHost`

Expected:
`IHost` is supported as well as `IWebHost`.

krajek:
Since `IWebHost` is deprecated, the library should support `IHost`.
Fixes #1614.

I have checked that the fix works on one of my production apps.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1211,7 +1211,13 @@ _NOTE: At this point the tool will be added to the local manifest file and insta
 
 ### Use the CLI Tool with a Custom IWebHost Configuration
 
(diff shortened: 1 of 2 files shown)


domaindrivendev:
Thx for the contribution @krajek 👍

PR #1619, opened 2020-04-09
Add endpoint routing support

### LINKED ISSUE #1401: Use endpoint routing to set SwaggerEndpoint so we can put it behind auth
I know there are a few open issues are endpoint routing but I think they all focus on generating docs for the end points. 

My understanding is that if you use end point routing when setting up middleware you can then chain authentication to those end points. That would make it really easy to put the swagger endpoint behind authentication and not leave it open to the world.

wu-yafeng:
- closes #1401
- Add endpoint routing support

### DIFF
--- a/src/Swashbuckle.AspNetCore.Swagger/DependencyInjection/SwaggerBuilderExtensions.cs
+++ b/src/Swashbuckle.AspNetCore.Swagger/DependencyInjection/SwaggerBuilderExtensions.cs
@@ -1,7 +1,20 @@
 ﻿using System;
+using System.Globalization;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Options;
+using Microsoft.OpenApi.Writers;
 using Swashbuckle.AspNetCore.Swagger;
+
+#if NETCOREAPP3_0
+
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Routing;
+using Microsoft.AspNetCore.Routing.Patterns;
+#endif
 
 namespace Microsoft.AspNetCore.Builder
 {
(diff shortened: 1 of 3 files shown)


src/Swashbuckle.AspNetCore.Swagger/DependencyInjection/SwaggerBuilderExtensions.cs
```diff
@@ -25,5 +38,70 @@ public static class SwaggerBuilderExtensions
...
+
+            return endpoints;
+        }
+
+        public static IEndpointRouteBuilder MapSwagger(this IEndpointRouteBuilder endpoints)
+        {
+            return endpoints.MapSwagger("/swagger/{documentName}/swagger.json");
+        }
+
+        private async static Task InvokeAsync(HttpContext context)
```
JamesNK, line 59: You can reuse logic in the middleware and avoid duplication by doing this:

https://github.com/dotnet/aspnetcore/blob/658b37d2bd3b0da2c07e25b53fa5c97ce8d656d3/src/Middleware/HealthChecks/src/Builder/HealthCheckEndpointRouteBuilderExtensions.cs#L75-L77

src/Swashbuckle.AspNetCore.Swagger/DependencyInjection/SwaggerBuilderExtensions.cs
```diff
@@ -25,5 +38,70 @@ public static class SwaggerBuilderExtensions
 
             return app;
         }
+#if NETCOREAPP3_0
+        public static IEndpointRouteBuilder MapSwagger(this IEndpointRouteBuilder endpoints, string pattern)
+        {
+            if (!RoutePatternFactory.Parse(pattern).Parameters.Any(x => x.Name == "documentName"))
+            {
+                throw new ArgumentException("Pattern must contains '{documentName}' parameter", nameof(pattern));
```
JamesNK, line 46:
```suggestion
                throw new ArgumentException("Pattern must contain '{documentName}' parameter", nameof(pattern));
```

wu-yafeng:
I avoid `SwaggerOptions.RouteTemplate` through rewriting values after setupAction invoked.
Maybe adding a analysis rule for swagger in endpoint routing.

``` csharp

app.UseEndpoints(endpoints=>
{
    endpoints.MapSwagger("/swagger/{documentName}/swagger.json",options=>{
        options.RouteTemplate = "/openapi/{documentName}/openapi.json";// Should report a diagnostic here? 'Assigning 'RouteTemplate' in endpoint routing will never work,use argument 'pattern' in method 'MapSwagger(...)' instead'
    });
});

```

domaindrivendev:
To avoid exposing the redundant options property, I'd suggest using a separate options type for the endpoint setupAction, and then map and wrap before passing through to the middleware:

```csharp
public static IEndpointRouteBuilder MapSwagger(
    this IEndpointRouteBuilder endpoints,
    string pattern,
    Action<SwaggerEndpointOptions> setupAction = null)
{
    ...

    Action<SwaggerOptions> middlewareSetupAction = options =>
    {
        var endpointOptions = new SwaggerEndpointOptions();
        setupAction?.Invoke(endpointOptions);

        options.RouteTemplate = pattern;
        options.SerializeAsV2 = endpointOptions.SerializeAsV2;
        options.PreSerializeFilters.AddRange(endpointOptions.PreSerializeFilters);
    };

    var pipeline = endpoints.CreateApplicationBuilder()
        .UseSwagger(middlewareSetupAction)
        .Build();

    ...
}
```

src/Swashbuckle.AspNetCore.Swagger/DependencyInjection/SwaggerBuilderExtensions.cs
```diff
@@ -25,5 +38,37 @@ public static class SwaggerBuilderExtensions
 
             return app;
         }
+#if NETCOREAPP3_0
+        public static IEndpointRouteBuilder MapSwagger(
+            this IEndpointRouteBuilder endpoints,
+            string pattern,
```
domaindrivendev, line 44: In place of the overload method below, could we just make the `pattern` property optional and default it to "/swagger/{documentName}/swagger.json". This way, the default value will be more visible via intellisense.

Issue #1423, opened 2019-12-17
Support providing custom ISwaggerProvider registration


sverrirs:
I was trying to implement the CustomProvider scenario to implement a SwaggerGenerator that would cache the swagger.json response after first generation.

This provided to be a little bit more involved than expected and I described my workaround in https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/356#issuecomment-566449902

It seems though that this scenario can be more easily supported by changing the DI service registration to a `services.TryAddTransient(...)` in this line
https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/master/src/Swashbuckle.AspNetCore.SwaggerGen/Application/SwaggerGenServiceCollectionExtensions.cs#L29

The line would become
```
services.TryAddTransient<ISwaggerProvider, SwaggerGenerator>();
```

That way the user could register their own implementation of `ISwaggerProvider` with the container before calling the `services.AddSwaggerGen(...)` extension method and no special workaround would be needed.

domaindrivendev:
What's the benefit of adding it _before_ vs after?

sverrirs:
Hi Richard,

The fix in question is to change the SwaggerGen function to perform a `TryAddTransient` internally. That way a user can register their custom implementation with the DI container before calling the SwaggerGen function. What that gives us is greater level of customization since when the SwaggerGen tries to add its built in one the TryAdd... function would detect an already present implementation (the custom one) thus skipping the registration of its built-in (now it always overrides).

domaindrivendev:
I understand the proposed change but unless I'm missing something the only tangible difference is around the positioning of the one-liner in `Startup.cs` to register the alternative `ISwaggerProvider`. That is, without your proposed change, this MUST be after the call to `AddSwaggerGen` whereas with it the line can be before or after. Or am I missing something here?

sverrirs:
Yes, you are correct. However what is needed currently is to replace the already registered instance in the DI container. It is a more complicated function call. :)

PR #1621, opened 2020-04-13
Use TryAdd to register services

JamesNK:
Avoid error if someone has already registered the service in app startup.

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/Application/SwaggerGenServiceCollectionExtensions.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/Application/SwaggerGenServiceCollectionExtensions.cs
@@ -25,11 +25,11 @@ namespace Microsoft.Extensions.DependencyInjection
             services.AddTransient<IConfigureOptions<SchemaGeneratorOptions>, ConfigureSchemaGeneratorOptions>();
 
             // Register generator and it's dependencies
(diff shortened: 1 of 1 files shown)


domaindrivendev:
This was requested previously to address #1423. While it's a pretty straightforward change, I didn't fully understand the value-add at the time of that request. As all these services are very specific to the Swagger generator provided by Swashbuckle, it would seem somewhat non-logical to register any of them before the `SwaggerGen` component itself.

However, as this is the second request for this change, I'm assuming there's a valid use-case that I'm just not seeing. I'm happy to merge this in but would like to understand the rationale a little better - just for my own education. Is it simply to give people the flexability to order their registrations however they see fit - or is their a specific use-case that's driving the need for this?

JamesNK:
```
As all these services are very specific to the Swagger generator provided by Swashbuckle, it would seem somewhat non-logical to register any of them before the SwaggerGen component itself.
```

Service registration shouldn't require an order unless it needs to. If you look at any of the AddServices methods in dotnet/aspnetcore you'll see they all use TryAdd for this reason.

https://github.com/dotnet/aspnetcore/blob/1360d6559e800a1b82623b8b7ac5d8009c339258/src/Security/Authorization/Core/src/AuthorizationServiceCollectionExtensions.cs#L32-L37

My library will have an add services method and it will override one of your services. I want it to work whether it is called before AddSwaggerGen or after. Today AddSwaggerGen will throw if the service has already been registered. There is no reason it needs to do this.

PR #1638, opened 2020-04-24
Add explicit support for aspnetcore31

bede-marian-stanev:
With netcoreapp3.0 now EOL we started upgrading to 3.1 and noticed that 3.x features are not supported if you install the package(s) in a netcoreapp3.1 - only the netstandard features are available.

This one adds explicit support for 3.1. 
SInce 3.0 EOL is not that far in the past I decided to keep the netcoreapp3.0 target so that people have more time to migrate/upgrade.

Also fixed the nuget deps for netcoreapp3.x in the Swashbuckle.AspNetCore nuspec.

Would be awesome if those changes make it into the next version.


domaindrivendev:
I need to understand exactly what you’re trying to solve a little better. Could you give an example of the SB features that are not supported for a 3.1 app?

bede-marian-stanev:
Basically everything that is wrapped in ```#if NETCOREAPP3_0 ... #endif```. This only applies if your app targets `netcoreapp3.0`.
`endpoints.MapSwagger(...)` is the easiest one to reproduce - it is not available if you target 3.1.

Another issue is that if you install the package in a netcoreapp3.1 it pulls all the 2.x netstadard deps.

domaindrivendev:
> Basically everything that is wrapped in #if NETCOREAPP3_0 ... #endif. This only applies if your app targets netcoreapp3.0.

This statement is incorrect. The functionality within that block will actually be available to any application that targets 3.0 _or above_. By design, the Swashbuckle build process targets _exactly two_ frameworks but the binaries produced can be referenced by applications targeting a much broader range of framework versions.

For example, if your application references 3.0 _or above_, it will use the `netcoreapp3.0` Swashbuckle binaries, and will be 100% compatible without any "loss of features". The same goes for the dependencies that get pulled in - an application targeting 3.0 _or above_ will pull in the transient dependencies from Swashbuckle's `netcoreapp3.0` package and not from it's `netstandard2.0` package as you've suggested above.

Regarding the `endpoints.MapSwagger` functionality, I suspect you're not seeing this in your `3.1` app because that feature was only merged recently and hasn't been published to Nuget yet.

bede-marian-stanev:
Indeed. I must have messed up big time during my tests.
Thanks for taking the time.

Issue #1670, opened 2020-05-15
Convert XML document comment to Markdown


foriequal0:
Hi! I wrote a converter that converts XML document comments to Markdown.
It was originally written for NSwag, but I want to graft it to here too.

This is the code:
https://github.com/foriequal0/Namotion.Reflection/tree/markdown/src/Namotion.Reflection/Markdown

And this is the tests:
https://github.com/foriequal0/Namotion.Reflection/blob/markdown/src/Namotion.Reflection.Tests/XmlDocsExtensionsMarkdownTests.cs

I want to hear your opinion and contribution guides before grafting it to here

domaindrivendev:
By design, Swashbuckle supports a very small subset of the standard XML Comments tags - specifically `<summary/>, <remarks/>, <param/> and <example/>`. The goal in doing this was to provide a convenient way for API authors to provide multiline, _yet simple_ textual descriptions for API operations, parameters, payloads and responses. I don't believe piggybacking onto XML Comments, which are actually intended for an internal audience (as opposed to exernal) is the best approach but it sure is convenient, and so I've continued to support it, but as I said for _simple_ descriptions only.

The code in that area is already complex and at this point I don't want to increase that complexity any further by trying to support the many other tags used for XML-comment based documentation. As this feature is a good fit for the filter-style extensibility options provided by Swashbuckle (in fact, this is exactly how the current XML comments support is implemented), I think it may be a good idea for someone in the community to create a separate "add-on" project (e.g. `Swashbuckle.RichXmlComments`) that incorporates XML-comments with a much richer level of support for the various tags available. Once created, I'd be more than happy to include it as a community link in the readme.

PR #1674, opened 2020-05-20
Convert Xml to Markdown for description

foriequal0:
Related: https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/1670

I've tried to stick to some rules to match Markdown rendered results would look similar to Jetbrains Rider's documentation rendering.

1. Preserve the given text format (indentation, newlines) for top-level texts if possible when it is converted to Markdown so you can write top-level comments in Markdown if you want.

Test fixture `MixedXmlAndMarkdown` would be a best example, and some other fixtures might show you how top-level texts and XML nodes interact.

However, not to break the intention of the node, it can inject some whitespaces.

e.g.
```xml
Some adjacent inlines
  <c>inline</c><c>inline</c>
```
is converted to
```markdown
Some adjacent inlines
  `inline` `inline`
```
Note that the indentation is preserved and the whitespace is inserted.
Markdown render will render it like this:
image

2. But normalize the text inside of XML nodes so it won't break the intention of the parent node.

e.g.
```xml
<para>
some paragraph can contain

multiple newlines

      and indentation.
</para>
```
If we preserve the format while converting it like this:
```markdown
some paragraph can contain

multiple newlines

      two newlines and indentation.
```
It'll become two paragraphs and one code block. It wasn't an intention of a `<para>` node.
image

I normalized it like this:
```markdown
some paragraph can contain
multiple newlines
two newlines and indentation.
```
and it becomes a single paragraph.
image

I know the code looks quite hacky and is worrisome to maintain. Then it would be great to include the first two commits to enable postprocessing the XML.

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenOptionsExtensions.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenOptionsExtensions.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Xml;
 using System.Xml.XPath;
 using Microsoft.OpenApi.Models;
 using Microsoft.AspNetCore.Mvc.ApiExplorer;
@@ -382,7 +383,7 @@ namespace Microsoft.Extensions.DependencyInjection
             string filePath,
             bool includeControllerXmlComments = false)
         {
-            swaggerGenOptions.IncludeXmlComments(() => new XPathDocument(filePath), includeControllerXmlComments);
+            swaggerGenOptions.IncludeXmlComments(() => new XPathDocument(filePath, XmlSpace.Preserve), includeControllerXmlComments);
         }
     }
 }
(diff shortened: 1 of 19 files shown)


domaindrivendev:
By design, Swashbuckle supports a very small subset of the standard XML Comments tags - specifically `<summary/>, <remarks/>, <param/> and <example/>`. The goal in doing this was to provide a convenient way for API authors to provide multiline, _yet simple_ textual descriptions for API operations, parameters, payloads and responses. I don't believe piggybacking onto XML Comments, which are actually intended for an internal audience (as opposed to exernal) is the best approach but it sure is convenient, and so I've continued to support it, but as I said for _simple_ descriptions.

The code in that area is already complex and at this point I don't want to increase that complexity any further by trying to support the many other tags used for XML-comment based documentation. As this feature is a good fit for the filter-style extensibility options provided by Swashbuckle (in fact, this is exactly how the current XML comments support is implemented), I think it may be a good idea for someone in the community to create a separate "add-on" project (e.g. `Swashbuckle.RichXmlComments`) that incorporates XML-comments with a much richer level of support for the various tags available. Once created, I'd be more than happy to include it as a community link in the readme.

PR #1629, opened 2020-04-19
Add examples to XML params

mattfrear:
Fixes #1628.
I ensure the example type is the correct type and not always a string.
Includes unit tests and updated Readme.

PR #1678, opened 2020-05-25
Do not duplicate parameter's description when there is an example, and inline the example.

foriequal0:
Previous PRs: #1552, #1629

```csharp
<param name="foo" example="baz">foo parameter</param>
public void Get(string foo) {}
```
Before:
```json
          {
            "name": "foo",
            "in": "path",
            "description": "foo parameter",
            "schema": {
              "type": "string",
              "description": "foo parameter",
              "example": "baz"
            }
          }
```

After removing duplication:
```json
          {
            "name": "foo",
            "in": "path",
            "description": "foo parameter",
            "schema": {
              "type": "string",
              "example": "baz"
            }
          }
```

After moving examples under its parameters:
```json
          {
            "name": "foo",
            "in": "path",
            "description": "foo parameter",
            "example": "baz",
            "schema": {
              "type": "string"
            }
          }
```

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsParameterFilter.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsParameterFilter.cs
@@ -1,5 +1,9 @@
-﻿using System.Reflection;
+﻿using System;
+using System.ComponentModel;
+using System.Globalization;
+using System.Reflection;
 using System.Xml.XPath;
+using Microsoft.OpenApi.Any;
 using Microsoft.OpenApi.Models;
 
 namespace Swashbuckle.AspNetCore.SwaggerGen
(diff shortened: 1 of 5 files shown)


src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsParameterFilter.cs
```diff
@@ -52,7 +56,32 @@ private void ApplyParamTags(OpenApiParameter parameter, ParameterInfo parameterI
...
+
+                var example = paramNode.GetAttribute("example", string.Empty);
+                if (!string.IsNullOrEmpty(example))
+                {
+                    parameter.Example = ConvertToOpenApiType(parameterInfo.ParameterType, parameter.Schema, example);
+                }
+            }
+        }
+
+        private static IOpenApiAny ConvertToOpenApiType(Type memberType, OpenApiSchema schema, string stringValue)
```
mattfrear, line 68: Maybe you shouldn't duplicate this method and move it somewhere so it can be used here and in XmlCommentsSchemaFilter.

domaindrivendev:
I believe the underlying bug has since been resolved by some refactors to the Xml comments filters and so this PR is no longer neccessary. I'm not sure if these changes are present in the lastest stable release but they will be in the latest preview release available on [myget.org](https://myget.org/feed/domaindrivendev/package/nuget/Swashbuckle.AspNetCore).

@foriequal0 if you get a chance, please try out the preview package and let me know if the duplication is in fact resolved. If it's not we'll need to create a new issue and subsequent PR. Thx

PR #1690, opened 2020-06-02
#1689 Fix documentation for multiple nested classes

### LINKED ISSUE #1689: BUG: XML documentation not included when class has deep nesting
Using version 5.4.1

When given class:
```
namespace MyProduct.Namespace
{
	public class BaseClass
	{
		[DataContract]
		public class FirstNest
		{
			/// <summary>
			/// This description is visible
			/// </summary>
			[DataMember]
			public SecondNest SomeName { get; set; }

			[DataContract]
			public class SecondNest
			{
				/// <summary>
				/// This description is NOT included
				/// </summary>
				[DataMember]
				public string MyProperty { get; set; }
			}
		}
	}
}
```

Then in the XML it generates as:
```
<member name="P:MyProduct.Namespace.BaseClass.FirstNest.SecondNest.MyProperty ">
  <summary>
        This description is NOT included
  </summary>
</member>
```
But it is search by the key `P:MyProduct.Namespace.FirstNest.SecondNest.MyProperty`.

I believe that the culprit is the `XmlCommentsNodeNameHelper.QualifiedNameFor()` which checks only one level of nesting.

siimv:
Fixes #1689

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsNodeNameHelper.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsNodeNameHelper.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
 using System.Text;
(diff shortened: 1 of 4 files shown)


domaindrivendev:
👍

PR #1720, opened 2020-06-24
Apply value type custom mappings to nullable version

### LINKED ISSUE #1648: After upgrading to 5.4.1 when explicitly mapping a struct type the mapping it not used for the nullable version
In 4.x if you explicitly specified a mapping for a struct type it would also be used for the nullable version of the struct. In 5.4.1 you have to explicitly map the nullable version as well. Is this change intentional?

In our API we use NodaTime.Serialization.JsonNet.Extensions.ConfigureForNodaTime to add support for (de)serializing NodaTime types (like Instant) in DTOs. For example Instant is serialized to a String like `2013-07-26T16:45:20.1234567Z`. Because of this we need to explicitly map these types in Swashbuckle so that it knows about this as well:

```
swaggerGenOptions.MapType(
    typeof(Instant),
    () => new OpenApiSchema {
        Type = "string"
    });
```

And here's an example DTO:
```
public class TestDto {
        public Instant instant;
        public Instant? nullableInstant;
}
```

In 4.x they would both be mapped as a string in swagger.json but in 5.x only the non-nullable one is:
```
"instant": {
  "type": "string"
},
"nullableInstant": {
  "nullable": true
}
```

Adding a mapping for `Instant?` works as expected.

Was this change intentional?

thomaslevesque:
Fixes #1648

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/SchemaGenerator.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/SchemaGenerator.cs
@@ -86,11 +86,22 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
 
         private bool TryGetCustomMapping(Type type, out Func<OpenApiSchema> mapping)
         {
+            // Look for a mapping specific to this type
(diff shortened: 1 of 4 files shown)


domaindrivendev:
I'm going to stick with the mapping-per-type semantics for now.

PR #1728, opened 2020-06-30
Configurable use of ref instead of oneOf for polymorphic types

xDDDx:
Added configuration to use `ref` intead of `oneOf`, for better support of code generation tools, that can not handle `oneOf` correct


domaindrivendev:
@xDDDx thanks for putting this together. As it happens, I'm in the middle of a refactor to the schema generator that will address your issue along with some other issues related to polymorphic / inheritance based schema structures. So, with respect, I'm going to decline this PR but will hopefully have a similar solution in place soon with the larger refactor. Thanks

PR #1755, opened 2020-07-24
Use netcore3.1

mikeblakeuk:


domaindrivendev:
What problem are you trying to solve here? Are you having compatability issues with the current version of Swashbuckle? If so, please provide details.

PR #1766, opened 2020-08-03
Improve polymorphism & inheritance behavior incl. more flexible config

### LINKED ISSUE #1564: Keep inheritance information
Right now I'm using this library for generating my models (in my case actually interfaces) including its inherited properties. But can someone tell my why the properties get mapped as direct properties of my interface and won't get actually "inherited"? Or is inheritance just not supported by OpenAPI? I'm not sure about this.

### LINKED ISSUE #1643: Classic inheritance going bad?
Well first thanks for your time to have a look at the following issue, much appreciated. I ran into an issue with inheritance and I'm not sure if I'm doing something wrong or that its a bug, I expect the later since I cannot find examples how to do it differently. 

## VERSION: ##
5.3.3 / .net core 2.2

## STEPS TO REPRODUCE: ##
You can download the exameple solution here: http://kieftsoftwareengineering.nl/files/WebApplication1.zip
But for the complete case I have the following code:

AnimalController:

```csharp
[Route("api/[controller]")]
[ApiController]
public class AnimalsController : ControllerBase
{
  ...
  // POST api/values
  [HttpPost("getOne")]
  public ActionResult Post([FromBody] Request request)
  {
      if(request.Animal is Cat)
      {
         return Ok(new Cat { Name = "cat", CatSpecificProperty = "somethingspecificAboutthiscat"});
      }

     return Ok(new Dog { Name = "dog", DogSpecificProperty = "somethingspecificAboutthiscat" });
   }
 }
```
Request object:
```csharp
public class Request
{
    public Animal Animal { get; set; }
}
```

Dog object:
```csharp
public class Dog : Animal
{
  public string DogSpecificProperty { get; set; }
}
```

Cat object:
```csharp
public class Cat : Animal
{
  public string CatSpecificProperty { get; set; }
}
```

And my startup looks like this:

```csharp
 public class Startup
 {
    ...
    {
      ...
      services.AddSwaggerGen(c =>
      {
        c.SwaggerDoc("v1", new OpenApiInfo { Title = "My API", Version = "v1" });
        c.GeneratePolymorphicSchemas();
      });

      services.AddSwaggerGenNewtonsoftSupport();
     }

     ...
}
```

## EXPECTED RESULT: ##
I  expect that a request object contains an animal object which can be either a dog or a cat. So I would expect the following json 👍 

```json
"Request": {
  "type": "object",
  "properties": {
  "animal": {
   "oneOf": [
    {
       "$ref": "#/components/schemas/Animal"
    },
    {
       "$ref": "#/components/schemas/Cat"
     },
     {
        "$ref": "#/components/schemas/Dog"
      }
   ],
   "nullable": true
  }
 },
 "additionalProperties": false
}
```
And the generated client looks like this:
```csharp
    public partial class Animal 
    {
    }

    public partial class Cat : Animal
    {
        public string CatSpecificProperty { get; set; }
    }

    public partial class Dog : Animal
    {
        public string DogSpecificProperty { get; set; }
    }
    
    public partial class Request 
    {
        public Animal Animal { get; set; }
    }
}
```

## ACTUAL RESULT: ##
However I receive a client that requires a cat object in the request. 
```json
"Request": {
  "type": "object",
  "properties": {
  "animal": {
   "oneOf": [
    {
       "$ref": "#/components/schemas/Cat"
     },
     {
        "$ref": "#/components/schemas/Dog"
      }
   ],
   "nullable": true
  }
 },
 "additionalProperties": false
}
```

```csharp
    public partial class Animal 
    {
    }

    public partial class Cat : Animal
    {
        public string CatSpecificProperty { get; set; }
    }

    public partial class Dog : Animal
    {
        public string DogSpecificProperty { get; set; }
    }
    
    public partial class Request 
    {
        public Cat Animal { get; set; }
    }
}
```

## ADDITIONAL DETAILS ##
I have used nswag studio to generate the client but I get similar results with other generators as well. (For example https://github.com/swagger-api/swagger-codegen) 

p.s.: I created the issue in the wrong project, recreated it here

### LINKED ISSUE #1662: discriminatorSelector in GeneratePolymorphicSchemas is not resolved through data contract resolver
Thanks for contributing to Swashbuckle.AspNetCore! As per the [contributing guidelines](CONTRIBUTING.md), please adhere to the following rules of thumb before submitting your issue:

Version 5.0.0 (apply to 5.4.x)
 
Currently, `GeneratePolymorphicSchemas()` will set the discriminator by default to `$type`.

This might not be the actual implementation at the client, and the user can change that but only by providing a function that returns the discriminator (`discriminatorSelector`).
It also add the `$type` property to the schema, even if it's not really part of my schema and there's nothing to do here except a custom document filter

This very simple and naive solution which does not allow much room for customization...

- Since it's a global handler and each model might have a different discriminator there's no mechanizem to describe it (e.g. something like an attribute on the base model which defines the discriminator)
- No all inheriting models have a discriminator, I would like to express inheritance but without a discriminator, which is added with no option to cancel. (e.g. definition on the attribute above that will define that the base class does not describe a discriminator, only inheritance)

The above issues can be solved local, by using a document filter which fixes everything based on the user's logic... However, I do think they should be part of the library.

The thing that is not possible to do (or I didn't find how to...) is to pass the discriminator key through the data contract resolver, so the property key is transformed.

For example, if the discriminator property in my model is **Type**, I want to pass it and when using newtonsoft or other JSON serializer, it will be passed as **type**...

---

IMHO, the solution here is to provide a`Context` parameter to the function we provide to `GeneratePolymorphicSchemas` so they have better control over the process.

In addition, take the discriminator selector and transform it via the data contract resolver so it express the right key going out as an API.

Finally, being able to control if the model has a discriminator or not, it might not have... right now it forces it by creating a discriminator and adding the property... Maybe if the discriminator returned is null, it will not add it....

---

domaindrivendev:
Fixes #1564 
Fixes #1643 
Fixes #1662

__Overview__
With the existing support for polymorphism/inheritance (enabled with the `GeneratePolymorphicSchemas` setting), Swashbuckle will implicitly use the `oneOf` keyword in base type schemas (see "Polymorphism" in the [Swagger docs](https://swagger.io/docs/specification/data-models/inheritance-and-polymorphism/)) AND the `allOf` keyword in sub type schemas (see "Model Composition" in the [Swagger docs](https://swagger.io/docs/specification/data-models/inheritance-and-polymorphism/)).

However, it turns out there's many use cases where these two keywords might be used independently. For example, `oneOf` is particularly useful for documenting the various acceptable structures for payloads that are bound to an abstract/base class. But `allOf` on the other hand is more useful for client code generators when you want to maintain an inheritance structure in the generated client models.

This PR splits the setting into two separate settings that can be enabled independently or together:

```csharp
services.AddSwagger(c =>
{
    ...
    c.UseOneOfForPolymorphism();
    c.UseAllOfForInheritance();
}
```

Additionnally, the existing implementation of "Polymorphic" schemas (i.e. schemas that include the `oneOf` keyword) didn't include the abstract/base class itself in the list which is technically incorrect. This PR also addresses this issue by including the abstract/base class in the list.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,4 +12,5 @@ artifacts/
 Thumbs.db
 test/WebSites/CliExample/wwwroot/api-docs/v1/*.json
 test/WebSites/CliExampleWithFactory/wwwroot/api-docs/v1/*.json
+test/WebSites/NswagClientExample/NSwagClient/
 *ncrunch*
--- a/Swashbuckle.AspNetCore.sln
+++ b/Swashbuckle.AspNetCore.sln
@@ -89,6 +89,8 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "CliExampleWithFactory", "te
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Swashbuckle.AspNetCore.TestSupport", "test\Swashbuckle.AspNetCore.TestSupport\Swashbuckle.AspNetCore.TestSupport.csproj", "{6E3C0128-931F-4438-AEDD-984E0E2B2C7B}"
 EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "NSwagClientExample", "test\WebSites\NswagClientExample\NSwagClientExample.csproj", "{9840E751-5845-431C-943B-C75CAE596A45}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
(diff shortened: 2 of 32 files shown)


maarten-kieft:
When can we expect the release for this one? Works as a charm :)

kgreed:
I am hanging out for this release 👍

domaindrivendev:
@kgreed `v5.6.0` now pushed to Nuget. Pull it down and let me know how it goes

Issue #1779, opened 2020-08-16
App running at subpath but httpContext.Request.PathBase is not taken into account


abdusco:
When running an app from a subpath, swagger ui middleware tries to redirect to index.html at root instead of subpath.

Here I have an app running at `/subapp`
```c#
public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
{
    app.Map("/subapp", app =>
    {
        if (env.IsDevelopment())
        {
            app.UseDeveloperExceptionPage();
        }
        
        app.UseRouting();
        app.UseSwagger(options => options.RouteTemplate = "docs/{documentName}.json");
        app.UseSwaggerUI(options =>
        {
            options.SwaggerEndpoint("docs/api.json", "api");
            options.RoutePrefix = "";
        });

        app.UseEndpoints(endpoints => { endpoints.MapControllers(); });
    });
}
```

when I go to `localhost:5000/subapp` swagger tries to redirects me to `index.html` 

https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/939ad54164af5820f93a926083d322899c09049f/src/Swashbuckle.AspNetCore.SwaggerUI/SwaggerUIMiddleware.cs#L57-L61

But the url resolves to nothing because swagger only responds to `/subapp/index.html`.

The solution is to prefix `relativeRedirectPath` with `httpContext.Request.PathBase`

domaindrivendev:
Reverted the merge as it introduced a regression. Have a broader plan to resolve this issue but needs to wait for next _major_ release

PR #1780, opened 2020-08-16
Fix redirect to index.html for apps running at a subpath

abdusco:
Fixes https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/1779


abdusco:
I've applied the same fix to ReDoc middleware too, but I haven't added a test specific to ReDoc.

domaindrivendev:
@abdusco - I've reverted these changes (see https://github.com/domaindrivendev/Swashbuckle.AspNetCore/commit/31c561dc0381412e817a78ce9bc1ff86ade4e6f8) as they're breaking other scenarios that were working previously.

I've have a plan for addressing your issue in a different way, but it will need to wait for the next _major_ release. I've re-opened the referenced issue for tracking

abdusco:
@domaindrivendev no problem. You know the code base better than anyone, I'm okay with any fix you choose to implement.

PR #1800, opened 2020-09-06
Adding support for yaml in ui

Rudde:
This pull request let users use yaml in addition to json.

Simply replace json with yaml

here

`app.UseSwaggerUI(x =>
            {
                x.SwaggerEndpoint("/swagger/v1/swagger.yaml", "Zeipt Dashboard API");
            });`

And you're set. Requests to "/swagger/v1/swagger.json" will still return the json file, and requests to "/swagger/v1/swagger.yaml" will return the yaml file if you continue to use json.

I've made this change because a client I have use yaml for their API documentation and referring them to my Swashbuckle json file every time I made an API change was not satisfactory and the yaml file was to be preferred.

### DIFF
--- a/src/Swashbuckle.AspNetCore.Swagger/SwaggerMiddleware.cs
+++ b/src/Swashbuckle.AspNetCore.Swagger/SwaggerMiddleware.cs
@@ -47,7 +47,14 @@ namespace Swashbuckle.AspNetCore.Swagger
                     filter(swagger, httpContext.Request);
                 }
 
-                await RespondWithSwaggerJson(httpContext.Response, swagger);
+                if (Path.GetExtension(httpContext.Request.Path.Value) == "yaml")
+                {
+                    await RespondWithSwaggerYaml(httpContext.Response, swagger);
+                }
+                else
+                {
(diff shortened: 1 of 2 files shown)


src/Swashbuckle.AspNetCore.Swagger/SwaggerMiddleware.cs
```diff
@@ -47,7 +47,14 @@ public async Task Invoke(HttpContext httpContext, ISwaggerProvider swaggerProvid
                     filter(swagger, httpContext.Request);
                 }
 
-                await RespondWithSwaggerJson(httpContext.Response, swagger);
+                if (Path.GetExtension(httpContext.Request.Path.Value) == "yaml")
```
domaindrivendev, line 50: Should probably support either .yml OR .yaml file extensions.

domaindrivendev: Meh - maybe ignore that suggestion, given the following line from the Swagger docs, I think leaving it as is makes sense:

> It is RECOMMENDED that the root OpenAPI document be named: openapi.json or openapi.yaml.

Issue #1298, opened 2019-10-07
Automatic handling of reverse proxy headers


Leon99:
NSwag introduced automatic handling of X-Forwarded-* headers a few months ago (https://github.com/RicoSuter/NSwag/pull/2196). Is there an equivalent in Swashbuckle?

domaindrivendev:
Are you just trying to get Swashbuckle to work in a reverse-proxy environment, or is there some specific reason why you need the documents `server` property populated based on reverse proxy headers?

If it's the former, then the recommended way to support reverse proxy environments is to use relative URLs in your Swagger configuration. See #1253 for details.

Leon99:
I'm trying to get Swashbuckle to work in a reverse-proxy environment. I don't have a problem with referencing the generated .json, but I don't want to glue the app to a specific hosting model, so I'd like to avoid specifying the path via `.UsePathBase` and have the generator taking it from X-Forwarded-Prefix.

freeranger:
There is a further issue in reverse proxy scenarios you may want to consider - authentication.
We have clients going to, say https://foo.myco.com/api/swagger and https://bar.myco.com/api/swagger which resolves to the same actual instance of a service behind the reverse proxy.
They need to get an OAuth2 token to use the "Try it Out" feature.
The authentication service url will be https://foo.myco.com/oauth and https://bar.myco.com/oauth respectively, so the swagger doc needs to reflect that.

If you were looking to handle reverse proxy environments automatically then it would be great to support this aspect too.

Our current solution is a PreSerialize filter to alter the TokenUrl for the OAuth2Scheme on every request.

Also bear in mind that Microsoft use X-Original-Host rather than the "standard" X-Forwarded-Host" for their Application Gateway RP so perhaps support the ability to override the header deemed to be the host header?

domaindrivendev:
@Leon99 - if you use the _relative_ URL approach as described here https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/1253#issuecomment-530850121, you're entire Swagger setup should be completely agnostic of the hosting model and there should be no need to leverage `UsePathBase`. Did you try the approach described?

Leon99:
[...]

PR #1801, opened 2020-09-07
Infer and assign server metadata base on common reverse proxy request…

domaindrivendev:
Wraps up #1298 by automatically honoring the following, relatively common, reverse proxy request headers:

- X-Forwarded-Host
- X-Forwarded-For
- X-Forwarded-Proto
- X-Forwarded-Port
- X-Forwarded-Prefix

### DIFF
--- a/src/Swashbuckle.AspNetCore.Swagger/SwaggerMiddleware.cs
+++ b/src/Swashbuckle.AspNetCore.Swagger/SwaggerMiddleware.cs
@@ -1,10 +1,12 @@
-﻿using System.Globalization;
+﻿using System;
+using System.Globalization;
 using System.IO;
 using System.Text;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Routing;
 using Microsoft.AspNetCore.Routing.Template;
+using Microsoft.Extensions.Primitives;
 using Microsoft.OpenApi.Models;
 using Microsoft.OpenApi.Writers;
 
@@ -33,13 +35,12 @@ namespace Swashbuckle.AspNetCore.Swagger
                 return;
             }
 
-            var basePath = !string.IsNullOrEmpty(httpContext.Request.PathBase)
-                ? httpContext.Request.PathBase.ToString()
-                : null;
-
             try
             {
-                var swagger = swaggerProvider.GetSwagger(documentName, null, basePath);
+                var swagger = swaggerProvider.GetSwagger(
+                    documentName: documentName,
+                    host: GetHostOrNullFromRequest(httpContext.Request),
+                    basePath: GetBasePathOrNullFromRequest(httpContext.Request));
 
                 // One last opportunity to modify the Swagger Document - this time with request context
                 foreach (var filter in _options.PreSerializeFilters)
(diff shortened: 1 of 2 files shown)


tinohager:
When will the next release be published on nuget?

domaindrivendev:
@tinohager v5.6.0 now pushed to Nuget. Pull it down and let me know how it goes

tinohager:
@domaindrivendev 

Hi, i have test the new version, but i have a few points that I find strange.

`X-Forwarded-Host:` Forward the Host header of the original client, is not required only for logging
`X-Forwarded-For:` Forward the IpAddress of the original client, is not required for swagger only for logging
`X-Forwarded-Proto:` Forward the protocol of the original client, is not required for swagger only for logging
`X-Forwarded-Port:` Forward the port of the original client, is not required for swagger only for logging

You can find further information here:
https://stackoverflow.com/questions/19084340/real-life-usage-of-the-x-forwarded-host-header

`X-Forwarded-Prefix` this solution partly works for me.

In my case we have a microservice and a reverse proxy in front of it. The microservice listen on `http://microservice1/swagger/index.html` in front of my reverse proxy the address is translated to `http://mydomain.com/microservice1/swagger/index.html`

In the `swagger.json` now i have a server url with the correct path but the UI loads the config object from the wrong path
`JSON.parse('{"urls":[{"url":"/swagger/v1/swagger.json","name":"My API V1"}]` instead of `JSON.parse('{"urls":[{"url":"/microservice1/swagger/v1/swagger.json","name":"My API V1"}]`. It seems that the `X-Forwarded-Prefix` is ignored in the UI.

domaindrivendev:
@tinohager I'm not sure what your point is wrt to the following headers. Could you please elaborate?

- `X-Forwarded-Host`
- `X-Forwarded-Proto`
- `X-Forwarded-Port`

For `X-Forwarded-For`, I believe Swashbuckle is interpreting this incorrectly as pointed out by @NatMarchand, and in fact SB should just be ignoring this header. This is an easy fix that I can push out ASAP.

Finally, for `X-Forwarded-Prefix`, you're correct in stating that it's ignored by the UI. In fact, for the UI _you_ explicitly provide the URL where the generated Swagger is found with the following config:

```csharp
app.UseSwaggerUI(c =>
{
    c.SwaggerEndpoint("/microservice1/swagger/v1/swagger.json", "My API V1");
});
```

If you want a configuration that works in both proxy and non-proxy environments, you can simply make it _relative_ to the page at which the UI is served by removing the leading slash and prefix as follows:

```csharp
app.UseSwaggerUI(c =>
{
    c.SwaggerEndpoint("v1/swagger.json", "My API V1");
});
```

This way, if the UI was accessed directly at `http://microservice1/swagger/index.html` then the Swagger JSON will be retrieved from: `http://microservice1/swagger/v1/swagger.json`

Or, if the UI was accessed via the proxy at `http://mydomain.com/microservice1/swagger/index.html`, then the Swagger JSON will be retrieved from: `http://mydomain.com/microservice1/swagger/v1/swagger.json`.

tinohager:
From my point of view this block can be discarded `GetHostOrNullFromRequest`

PR #1804, opened 2020-09-07
Correct order of attribute inheritance when using model metadata and duplicate attributes

IanKemp:
If an attribute (or subclass) is defined on both a model and its metadata class, Swashbuckle will use the latter's definition. This differs from ASP.NET Core which takes the model as the source of truth. This commit makes Swashbuckle behave the same as ASP.NET Core.

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/MemberInfoExtensions.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/MemberInfoExtensions.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
 using Microsoft.AspNetCore.Mvc;
@@ -9,8 +10,7 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
     {
         public static IEnumerable<object> GetInlineAndMetadataAttributes(this MemberInfo memberInfo)
         {
-            var attributes = memberInfo.GetCustomAttributes(true)
-                .ToList();
+            var attributes = memberInfo.GetCustomAttributes(true).ToList();
 
             var metadataTypeAttribute = memberInfo.DeclaringType.GetCustomAttributes(true)
                 .OfType<ModelMetadataTypeAttribute>()
@@ -21,7 +21,16 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
 
             if (metadataMemberInfo != null)
             {
-                attributes.AddRange(metadataMemberInfo.GetCustomAttributes(true));
+                // if the same attribute (or a derived version) exists on both the model class and its metadata
+                // class, asp.net core uses the one on the model class - we need to conform with that behaviour
+                foreach (var metadataAttribute in metadataMemberInfo.GetCustomAttributes(true))
+                {
+                    if (!attributes.Any(a => a.GetType() == metadataAttribute.GetType()
+                                             || metadataAttribute.GetType().IsSubclassOf(a.GetType())))
+                    {
+                        attributes.Add(metadataAttribute);
+                    }
+                }
             }
 
             return attributes;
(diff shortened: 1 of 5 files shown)


src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/MemberInfoExtensions.cs
```diff
@@ -21,7 +21,16 @@ public static IEnumerable<object> GetInlineAndMetadataAttributes(this MemberInfo
...
+                // class, asp.net core uses the one on the model class - we need to conform with that behaviour
+                foreach (var metadataAttribute in metadataMemberInfo.GetCustomAttributes(true))
+                {
+                    if (!attributes.Any(a => a.GetType() == metadataAttribute.GetType()
+                                             || metadataAttribute.GetType().IsSubclassOf(a.GetType())))
+                    {
+                        attributes.Add(metadataAttribute);
+                    }
+                }
             }
```
domaindrivendev, line 34: I think it would be a simpler implementation to just union both sets of attributes but order so that the inline attributes (i.e. the ones on the model itself) are at the end, and will therefore take precedence in the downstream processing.

IanKemp: That is indeed the simplest implementation, but relies on enumerating the returned attribute collection in the correct order to get correct results.

The implementation I've chosen yields a collection that contains only the _effective_ attributes that must be emitted to the generated `swagger.json`. That collection can be enumerated in any order, inspected in the debugger, etc. and you will immediately know what you are getting. It's a more complex and slower algorithm, but also more robust, and I don't believe the added complexity is a massive concern, since it's encapsulated in this single method.

IanKemp: Given my above statement, perhaps it would be appropriate to rename that method to `GetEffectiveAttributes` or similar.

domaindrivendev: Added complexity is always a concern for me 😉 

So, I dug a little deeper into the ASP.NET Core code to see exactly how validation works when you have the same validation attribute duplicated on both the original type and metadata type. And, it looks like _both_ attributes are used to validate a corresponding model instance. I was able to confirm this behavior with the following test controller:

```csharp
[Route("[controller]")]
[ApiController]
public class TestController
{
    [HttpPost]
    public IActionResult Create([FromBody]TestDto test)
    {
        throw new System.Exception();
    }
}

[ModelMetadataType(typeof(TestDtoMetadataType))]
public class TestDto
{
    [Range(1, 10)]
    public int Value { get; set; } 
}

public class TestDtoMetadataType
{
    [Range(1, 20)]
    public int Value { get; set; } 
}
```

If I submit a JSON value of `{ "Value": 0 }` to this operation, the built-in model validation returns the following:

```json
{
  "type": "https://tools.ietf.org/html/rfc7231#section-6.5.1",
  "title": "One or more validation errors occurred.",
  "status": 400,
  "traceId": "|430cd1b5-41d9d40783f43663.",
  "errors": {
    "Value": [
      "The field Value must be between 1 and 10.",
      "The field Value must be between 1 and 20."
    ]
  }
}
```

As you can see, both attributes are being applied. This actually makes it very difficult for Swashbuckle to assign appropriate values for the generated schema fields (in this case `minimum`, and `maximum`) because it would need to combine values from mulitple attributes, taking the underlying semantics of each attribute type into account. I'm not willing to take on this complexity to support such a questionable use-case (`the ModelMetadataType` isn't meant to be used in this manner). 

So, the reality is neither take precedence. With this being the case, I actually think it makes more sense to leave the code exactly as it is now in master. My rationale being ... if a developer puts the same attribute in both places, it's a reasonable assumption that their intention was actually to use the `ModelMetadataType` version exclusively, given they went to the trouble of creating it etc.

PR #1808, opened 2020-09-09
Description Attribute filter used for OpenAPI documentation

Rudde:
This is a filter much like XmlCommentsSchemaFilter but instead of using the summary from the XML documentation it used the text defined in [Description("text")] attribute. 

I understand that the purpose of having filters like this is that we can write our own filters to fit our needs, but I feel like this is so ingrained in C# that it is at least as native as supporting XML docs. I was very confused when I figured out it was not supported but XML docs was. 

I even think it would make sense to consider this default behavior. 

Behavior when combined with .IncludeXmlComments() is that whichever one if decleared last will overwrite whatever the one in front of it if both are initialized. If both are used but both aren't.

I made this as a draft as it might not be wanted in the main project, and to to get an opinion on if this should or should not be default behavior as I think it should.


domaindrivendev:
I'm not opposed to adding support for this attribute but my preference is to implement it in the `SchemaGenerator itself`, alongside existing support for `DefaultValueAttribute` and `ObsoleteAttribute`, rather than as a filter.

PR #1809, opened 2020-09-10
Added possibility to define client-side interceptors

TobbenTM:
This will allow us to define interceptors without having to replace the
entire index.html for simple use cases such as adding XSRF tokens to all
requests. The functions are parsed and passed to the SwaggerUI
configuration without using eval().

### DIFF
--- a/README.md
+++ b/README.md
@@ -1108,6 +1108,8 @@ app.UseSwaggerUI(c =>
     c.ShowCommonExtensions();
     c.EnableValidator();
     c.SupportedSubmitMethods(SubmitMethod.Get, SubmitMethod.Head);
+    c.UseRequestInterceptor("(request) => { return request; }");
+    c.UseResponseInterceptor("(response) => { return response; }");
 });
 ```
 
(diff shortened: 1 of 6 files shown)


DaiKnapz:
Hi there, would there be anything that needs to be done to help get this PR in. It would be a really useful feature for our current project as I have had no luck so far trying to get XSRF protection working with swagger. So far I've been stuck trying to get it to work with injected javascript as described [here](https://aspnetboilerplate.com/Pages/Documents/Swagger-UI-Integration), but this implementation looks far simpler.

TobbenTM:
> 
> 
> Hi there, would there be anything that needs to be done to help get this PR in. It would be a really useful feature for our current project as I have had no luck so far trying to get XSRF protection working with swagger. So far I've been stuck trying to get it to work with injected javascript as described [here](https://aspnetboilerplate.com/Pages/Documents/Swagger-UI-Integration), but this implementation looks far simpler.

I'm not the one who can answer what needs to be done to be merged, but as far as hacking it to work until that time, we've used the following dirty code in our injected javascript:

```javascript
      var originalRequestInterceptor = window.ui.getConfigs().requestInterceptor;
      var csrfInterceptor = function (req) {
        req.headers["X-XSRF-TOKEN"] = csrfCookie;
        return req;
      };
      window.ui.getConfigs().requestInterceptor = function (req) {
        req = csrfInterceptor(req);
        return originalRequestInterceptor(req);
      }
```

Obviously, injecting a proper interceptor when building the UI which this PR would allow is much better, the above might solve your issue temporarily. It is not using any public APIs though, so it might break. It should also preserve existing interceptors, so that any other Swagger UI extensions should continue to work.

PR #1817, opened 2020-09-16
Fixes #1661

### LINKED ISSUE #1661: Make IDocumentFilter async compatible
Hi, in our application we want to support feature flag (https://docs.microsoft.com/en-us/azure/azure-app-configuration/use-feature-flags-dotnet-core) and some endpoint will be activate/deactivate based on the maturity of the feature or the target environment (Dev, SIT, UAT, Prod, ...), we want to remove the endpoint from the Swagger UI when the feature endpoint is not active. I use the IDocumentFilter to do so, but feature flag is Task-based, so i had to use GetAwaiter().GetResult() to make it work, but it would be better to have a async version of the IDocumentFilter like IActionFilter/IAsyncFilter of aspnetcore (https://docs.microsoft.com/en-us/aspnet/core/mvc/controllers/filters?view=aspnetcore-3.1#implementation). Here the code that i use as an example : 
```csharp
public class FeatureGateDocumentFilter : IDocumentFilter
{
    private readonly IFeatureManager _featureManager;
    public FeatureGateDocumentFilter(IFeatureManager featureManager)
    {
        _featureManager = featureManager;
    }

    {
        foreach(var apiDescription in context.ApiDescriptions)
        {
            if (apiDescription.ActionDescriptor is ControllerActionDescriptor action)
            {
                var controllerFeatureNames = action.MethodInfo
                    .DeclaringType
                    ?.GetCustomAttributes<FeatureGateAttribute>()
                    .SelectMany(att => att.Features)
                    .ToList() ?? new List<string>();

                var actionFeatureNames = action.MethodInfo
                    .GetCustomAttributes<FeatureGateAttribute>()
                    .SelectMany(att => att.Features)
                    .ToList();

                bool showOperation = true;
                if (actionFeatureNames.Any())
                {
                    showOperation = actionFeatureNames.Any(featureName => _featureManager.IsEnabledAsync(featureName).GetAwaiter().GetResult());
                }
                else if(controllerFeatureNames.Any())
                {
                    showOperation = controllerFeatureNames.Any(featureName => _featureManager.IsEnabledAsync(featureName).GetAwaiter().GetResult());
                }

                if (!showOperation)
                {
                    if(Enum.TryParse<OperationType>(apiDescription.HttpMethod, true, out var httpVerb))
                    {
                        swaggerDoc.Paths[$"/{apiDescription.RelativePath}"].Operations.Remove(httpVerb);
                    }
                    else
                    {
                        swaggerDoc.Paths[$"/{apiDescription.RelativePath}"].Operations.Clear();
                    }
                }
            }
        }
    }
}
```

aersam:


src/Swashbuckle.AspNetCore.Swagger/ISwaggerProvider.cs
```diff
@@ -13,6 +14,14 @@ public interface ISwaggerProvider
...
     }
 
+    public interface IAsyncSwaggerProvider
+    {
+        Task<OpenApiDocument> GetSwaggerAsync(
+            string documentName,
+            string host = null,
+            string basePath = null);
+    }
+
```
domaindrivendev, line 24: Independently of this change, I was wanting to change the `GetSwagger` signature to the following as it aligns more closely with Swagger/OpenAPI terminology.

```csharp
OpenApiDocument GetSwagger(string documentName, string serverUrl = null);
```

Now, seeing as we're introducing a _new_ interface anyway, I think it might be a good opportunity to introduce that preferred signature. Does that make sense?

domaindrivendev: Furthermore, and on a similar note of opportunism, I wonder if we should call it `IOpenApiProvider` instead of `IAsyncSwaggerProvider`?

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/IDocumentFilter.cs
```diff
@@ -9,6 +10,11 @@ public interface IDocumentFilter
         void Apply(OpenApiDocument swaggerDoc, DocumentFilterContext context);
     }
 
+    public interface IAsyncDocumentFilter : IDocumentFilter
+    {
+        Task ApplyAsync(OpenApiDocument swaggerDoc, DocumentFilterContext context);
+    }
+
```
domaindrivendev, line 17: Similar to my point above, I'm wondering about the name for the new interfaces. Prefixing with `Async` feels a little unconventional because really the standard is to postfix the async methods as opposed to the interface itself. Of course it can't be the same as the original name but perhaps we could move to something like `IOpenApiDocumentFilter`, `IOpenApiOperationFilter` etc.?

aersamkull: Yes it's a bit strange, but it's what MSFT does: for example see [IAsyncEnumerator](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.iasyncenumerator-1?view=dotnet-plat-ext-3.1) with the MoveNextAsync Method.
But of course I can call it something like IOpenApiDocumentFilter (I like this one)

src/Swashbuckle.AspNetCore.Swagger/ISwaggerProvider.cs
aersamkull: Well, of course we could change the interface, but that sounds breaking. Is it a common use case to use ISwaggerProvider directly? I never did that, therefore doesn't hurt me :)
Would you change the CLI as well to accept --serverurl rather then --host and --basepath?
Of course just renaming IAsyncSwaggerProvider to IOpenApiDocumentProvider would unproblematic

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/IDocumentFilter.cs
aersamkull: Maybe we could mark IDocumentFilter with [Obsolete] then

niemyjski:
Any updates on this?

damilpanchal:
Any updates on this yet ? This seems like an important PR still open

martincostello:

aersam:
I'll close it as I dont work with .Net anymore and I wont find time to do this

Issue #1555, opened 2020-03-06
IFormFile couldn't show as file upload button directly


Aoi-hosizora:
dotnetcore 3.0
Swashbuckle.AspNetCore : 5.0.0-rc5
Swashbuckle.AspNetCore.SwaggerUI : 5.0.0-rc5
Feature Requests

I used `[FromForm(Name = "file"] IFormFile imageFile` to upload a image (single file), but the swagger generated this as a object. 

I have search some issues and I have a solution following. But I don't want to copy this code anywhere, so I hope that Swashbuckle could support IFormFile render to upload file directly.

```c#
[AttributeUsage(AttributeTargets.Method)]
public class FileUploadAttribute : System.Attribute { }

// options.OperationFilter<FormFileOperationFilter>();

public class FormFileOperationFilter : IOperationFilter {

    private static readonly string[] FileParameters = { "contentType", "contentDisposition", "headers", "length", "name", "fileName" };
    private const string FormData = "multipart/form-data";
    private const string Json = "application/json";

    public void Apply(OpenApiOperation operation, OperationFilterContext context) {
        var attr = context.MethodInfo.GetCustomAttributes(true).OfType<FileUploadAttribute>().Any();
        if (!attr) {
            return;
        }
        var content = operation.RequestBody.Content;

        var jsonProperties = GetProperties(content, Json);
        foreach (var parameter in jsonProperties.Where(p => FileParameters.Contains(p.Key))) {
            jsonProperties.Remove(parameter);
        }
        if (jsonProperties.Count == 0) {
            content.Remove(Json);
        }

        var formDataProperties = GetProperties(content, FormData);
        formDataProperties.Add("file", new OpenApiSchema { Type = "file" });
    }

    private static IDictionary<string, OpenApiSchema> GetProperties(IDictionary<string, OpenApiMediaType> contents, string mediaType) {
        if (!contents.ContainsKey(mediaType)) {
            contents.Add(mediaType, new OpenApiMediaType {
                Schema = new OpenApiSchema {
                    Type = "object",
                    Properties = new Dictionary<string, OpenApiSchema>()
                }
            });
        }
        return contents[mediaType].Schema.Properties;
    }
}
```

![image](https://user-images.githubusercontent.com/31433480/76095029-8bccad00-5ffe-11ea-98b1-4bbaf714e0b4.png)

[...]

Issue #1183, opened 2019-07-01
File upload support is currently partially/incorrectly documented


MikeBeaton:
Hi,

Would a PR adding (very brief) documentation for the file upload support be welcomed?

At the moment:

 - The [README.md](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/07adbdbf833510bfb050b98da3a037c2f911722b/README.md) still points the user to [Swashbuckle.AspNetCore.Filters](https://github.com/mattfrear/Swashbuckle.AspNetCore.Filters) for file upload support, but in fact (on looking through the commit histories and issues of both projects - but unfortunately not the current README.md of either) file upload support has been in this project since 4.0, and hence has been removed from Swashbuckle.AspNetCore.Filters.

It would simplify tracking down how to use file uploads if that small piece of text in the README was removed.

- `IFormFile` is indeed now supported by this project, with a clean file upload button, but this does not work if the `IFormFile` is marked `[FromForm]`, which is counterintuitive (to me, at least!)

With `IformFile file` parameter:

![image](https://user-images.githubusercontent.com/11946605/60445893-14957c80-9c18-11e9-84be-ab0a1377faf0.png)

With `[FromForm]IFormFile file` parameter:

![image](https://user-images.githubusercontent.com/11946605/60445940-342ca500-9c18-11e9-9cd8-a032225c0e6a.png)

It would be helpful if the requirement not to tag `IFormFile` with `[FromForm]` was documented, since I guess (?!) that this is something that maybe not just I am going to try by mistake. (At the same time, having this info, very briefly, would serve the additional purpose of clarifying what the user has to do to get a file upload up and running, right in the README.)

Does that sound sensible? Thanks!

domaindrivendev:
@MikeBeaton makes sense. If you want to go ahead and create a PR for the docs update I'll be happy to merge. Thx

dernop:
Is this really just a documentation issue and not a bug? I was stuck on this problem for two days now and i was only to solve it by "trial-and-error", when i eventually removed the [FromForm] attribute from the IFormFile parameter. 
Since the ASP app works fine WITH this attribute and i thus consider it "legal" to do, i rather consider this a bug. It should also work with this attribute...

domaindrivendev:
[...]

PR #1835, opened 2020-09-24
Important, straightforward README update: IFormFile support not in .Filters repo + IFormFile requires no FromForm

MikeBeaton:
This updates README.md to remove the statement (wrong since v4) that `IFormFile` support requires using the separate .Filters repo, and also to briefly explain up-front that `IFormFile` must be used without `FromForm`.

@domaindrivendev although  #1183 is now closed (I think maybe for other reasons, such as that the discussion under the issue had gone back on to the repeated debate about `IFormFile` and `FromForm`, and that after the delay in your being able to reply, I also hadn't been able to get back to this for quite some time either), but the minor changes to the documentation which I suggested there (and which you kindly agreed sounded sensible, back in Jan.) still remain needed.

I have included some of the explanation about `IFormFile` and `FromForm` from your own info about this issue in #1555.

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,7 +170,7 @@ These packages are provided by the open-source community.
 
 |Package|Description|
 |---------|-----------|
-|[Swashbuckle.AspNetCore.Filters](https://github.com/mattfrear/Swashbuckle.AspNetCore.Filters)| Some useful Swashbuckle filters which add additional documentation, e.g. request and response examples, a file upload button, etc. See its Readme for more details |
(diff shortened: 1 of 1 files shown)


MikeBeaton:
@domaindrivendev  - sorry, I know you're v. busy, but these two small, but significant change to the README would have saved me considerable time hunting down why my `IFromFile` params were not displaying properly in the swagger ui when I first started with your excellent library, and it's clear from all the issues that lots of people assume (wrongly) that it's correct, or at least harmless, to decorate `IFormFile` from `FromForm`.

domaindrivendev:
Much appreciated @MikeBeaton and thx for your patience

Issue #1836, opened 2020-09-24
Exception within method GetHostOrNullFromRequest in SwaggerMiddleware


mhoeper:
Hi, 

I get the below exception when running Swagger 5.6.0 on our servers. Probably the load balancer set a bad or empty X-Forwarded-Host header. Can you please surround the method with try/catch and in case of exception, return null? Or, another way would be to optionally set the host in startup for swagger. 

`System.UriFormatException at Swashbuckle.AspNetCore.Swagger.SwaggerMiddleware.GetHostOrNullFromRequest`

simluk:
Hi there,

We also facing same issue and the issue is that our internal proxy adds following headers:
`
X-Forwarded-Host : <proxy-one>, <proxy-two>
X-Forwarded-Server : <server-one>, <server-two>
`
The swashbuckle code on the other hand executes following:
https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/1a2b824ebff66b68e00cbe6d3027a2d070476186/src/Swashbuckle.AspNetCore.Swagger/SwaggerMiddleware.cs#L72

Takes [0] element but does not split it with comma. I believe this has to be fixed to support comma separated host/server list. Let me know if I should raise PR for this because it is blocking update to new version.

Thanks,
Simonas

simluk:
@domaindrivendev - when can we expect to have v6.0.0 released?

mhoeper:
@simluk I forked the SwaggerMiddleware.cs and just return null in GetHostOrNullFromRequest method instead. Maybe this is a mitigation that works for you, too, until the fix is released.

domaindrivendev:
@simluk. It will likely be a week or two but you can pull down preview packages built from latest master from myget.org in the meantime.

See https://myget.org/feed/domaindrivendev/package/nuget/Swashbuckle.AspNetCore

simluk:
@domaindrivendev when do you think the v6.0.0 is going to be released? We cannot use nuget sources other than nuget.org, so it would be great to have the fix in there.

PavelA85:
Hi,
any update on this?

Melchy:
Hi,
any update on this?

pada57:
hello
same issue here, any update on when this will be fix ?

ggleo08:
aha , same mistake,
![image](https://user-images.githubusercontent.com/32858631/104573378-c7f58300-568f-11eb-8deb-0609bb0c138f.png)

PR #1838, opened 2020-09-26
Adding support for comma separated X-Forward-* headers

simluk:
Bugfix for following [issue](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/1836).


domaindrivendev:
No longer applicable as I've removed support for the `X-Forwarded-*` headers altogether. My current view on the this is that the application as a whole should be configured to work with proxy servers and load balancers as described [here](https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/proxy-load-balancer?view=aspnetcore-3.1).

PR #1843, opened 2020-09-30
Allow string enum defaults to respect all enum member name transforms

MikeBeaton:
**Summary**

This is a fix for #1164 .

It fixes string enum default values being incorrect in two different situations:

- Camel case settings change the enum member name
- The `EnumMember` attribute changes the enum member name in `Newtonsoft.Json` (noting that, if and when implemented, [`JsonStringEnumMember`](https://github.com/dotnet/runtime/issues/31081#issuecomment-635521221) in `System.Text.Json` would have the same problem)

EDIT: I've removed the overly long description here of how I thought through to the most lightweight way to do this - hopefully the code itself can speak for itself. Of course I remain more than happy to discuss if needed, explain or justify (if I can!) and modify if it helps!


MikeBeaton:
@domaindrivendev Now re-organised: keeping the extra required info in a private dictionary in `SchemaRepository` and no longer changing any existing public API surface.

domaindrivendev:
@MikeBeaton in general, I think the current approach that Swashbuckle (not your PR) uses to derive `default` (and `example`) values is somewhat flawed. Rather than reverse engineering an appropriate `IOpenApiAny` instance based on type and serializer metadata, I think it should be invoking the actual serializer configured by the application (STJ, Newtonsoft etc.) to get a raw JSON representation of the C# default (or example), and then constructing the `IOpenApiAny` instance from that JSON instead. In fact, this deserialization from JSON value to `IOpenAPIAny` is _exactly_ what the OpenAPI.NET library does when deserializing a Swagger / OpenAPI JSON document into an `OpenApiDocument`, so I'm hoping there might be something there that can be leveraged.

IMO, this would be the ideal approach but it would require a significant restructuring of the code to facilitate in a clean way. I'm already planning a fairly large re-design of the `SchemaGenerator/ISerializerDataContractResolver` abstractinos and will likely incorporate this "preferred" `default` and `example` handling into that body of work. However, it may be a number of months off ...

So, in the meantime, I'm on board with getting something out that solves the immediate problem you're facing WRT to string enums.

MikeBeaton:
@domaindrivendev 

Yes, I agree with what you're saying about the ideal approach.

In terms of fixes for now, thanks for being willing to get something out. How is this existing code (i.e. this PR &/or the merged PR) looking?

I have just back-ported a test and fix which was only in the merged PR into this PR where it belongs (tests 
62481ad + fix to make those tests pass ad84307), but NB there are tests (test cases) in the merged PR which confirm that both proposed sets of changes (defaults with $ref/allOf + string enum default renaming) work in combination with each other which (by definition) can't be in either PR separately.

domaindrivendev:
Doh - that's my bad that the master branch is failing - will try get that resolved ASAP.

Re this PR, I provided two additional inline points that you haven't addressed yet - did you take a look at them?

domaindrivendev:
No it was my bad - never submitted review, you should see them now

src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/ISerializerDataContractResolver.cs
```diff
@@ -11,13 +11,14 @@ public interface ISerializerDataContractResolver
...
-        public static DataContract ForPrimitive(Type underlyingType, DataType dataType, string dataFormat, IEnumerable<object> enumValues = null)
+        public static DataContract ForPrimitive(Type underlyingType, DataType dataType, string dataFormat, IEnumerable<object> enumValues = null, Func<object, string> enumConverter = null)
         {
             return new DataContract(
                 underlyingType: underlyingType,
                 dataType: dataType,
                 dataFormat: dataFormat,
-                enumValues: enumValues);
+                enumValues: enumValues,
+                enumConverter: enumConverter);
```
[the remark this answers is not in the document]
MikeBeaton: Okay, I've made this change

src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/SchemaGenerator.cs
```diff
@@ -145,9 +145,14 @@ private OpenApiSchema GenerateSchemaForType(Type type, SchemaRepository schemaRe
...
         }
 
-        private OpenApiSchema GeneratePrimitiveSchema(DataContract dataContract)
+        private OpenApiSchema GeneratePrimitiveSchema(DataContract dataContract, SchemaRepository schemaRepository)
         {
-            var schema = new OpenApiSchema
+            if (dataContract.EnumConverter != null)
+            {
+                schemaRepository.RegisterEnumConverter(dataContract.UnderlyingType, dataContract.EnumConverter);
+            }
```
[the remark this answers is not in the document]
MikeBeaton: I was previously thinking of `SchemaRepository` as something that held all seen schemas. So that it might make sense to store additional information about each of them, e.g. about how to create a value for each schema type in the repository (potentially replacing the existing `OpenApiAnyFactory`).

But anyway, that's not how it works at the moment, since it only stores schemas which are required for `$ref` lookups.

So I agree that adding an enum converter dictionary, one which ended up being for schemas which might not even themselves be in the repository, looks tacked on!

So I've re-done this the way you've suggested, in the latest version of this PR.

test/Swashbuckle.AspNetCore.Newtonsoft.Test/SchemaGenerator/NewtonsoftSchemaGeneratorTests.cs
```diff
@@ -88,6 +90,18 @@ public void GenerateSchema_GeneratesFileSchema_IfFormFileOrFileResultType(Type t
...
+        [Fact]
+        public void GenerateSchema_DedupsEnumValues_IfEnumTypeWithDuplicateValues()
+        {
+            var schemaRepository = new SchemaRepository();
+
+            var referenceSchema = Subject().GenerateSchema(typeof(HttpStatusCode), schemaRepository);
+
+            Assert.NotNull(referenceSchema.Reference);
+            var schema = schemaRepository.Schemas[referenceSchema.Reference.Id];
+            Assert.Equal(schema.Enum.Cast<OpenApiInteger>().Select(v => v.Value).Distinct().Count(), schema.Enum.Count);
```
MikeBeaton, line 102: Because we're making a _dictionary_ from enum member to converted value, we are forced to do the deduplication earlier, in the two separate contract resolvers. Therefore this test is now needed here to check that it's being done right, and won't cause a duplicate key exception.

src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/SchemaGenerator.cs
```diff
@@ -155,8 +159,7 @@ private OpenApiSchema GeneratePrimitiveSchema(DataContract dataContract)
 
             if (dataContract.EnumValues != null)
             {
-                schema.Enum = dataContract.EnumValues
-                    .Distinct()
+                schema.Enum = dataContract.EnumValues.Values
```
MikeBeaton, line 162: Removed `Distinct()` here, as it's theoretically done already (see comment below). However, with pathalogical enum member name remapping, the values here could be distinct even if though the keys can't be. There isn't a test case that makes this happen, currently, and not really sure if it needs addressing?

Actually, on reflection, maybe this is exactly the situation that the previous enum duplicates test case was really intended to simulate? In which case I guess it does need addressing, perhaps with a quick test which actually uses enum member renaming to get to this situation(?), and by restoring `Distinct()` here.

MikeBeaton (commented):
I'm not quite sure if you see my comments already, or if I have to click 'Submit review', so I'm adding this comment and clicking it.

test/Swashbuckle.AspNetCore.Newtonsoft.Test/SchemaGenerator/NewtonsoftSchemaGeneratorTests.cs
domaindrivendev: 👍

domaindrivendev:
@MikeBeaton this looks good now - thx for making those updates. There's still one or two stylistic changes & test organization changes that I'd like to make, but rather than more back and forth on the PR, I think it would be quickest for me to apply those with a follow-up commit.

Thanks a lot for the contribution!

Issue #1850, opened 2020-10-02
Default values not present when using UseAllOfToExtendReferenceSchemas


MikeBeaton:
Background: I'm raising these issues because I need to be able to use string enums in a live project's API, with enum name conversion (specifically, camel casing) *and* with parameter defaults.

I've been running my own, patched versions of the code - first v4, now v5 - to do this successfully, but I'd love to be able to offer this back and have it merged ... if possible!

- Many thanks for fixing string enum camel casing in v5 (which, as per #1164, now works great for enum values but not yet for enum defaults)
- I've developed what I think could be a lightweight enough solution to correctly converting enum defaults in #1843 - if so, it would mean a working fully solution to the above, using `UseInlineDefinitionsForEnums`
- `UseInlineDefinitionsForEnums` is required in the above solution because defaults aren't legal in OpenAPI when using $ref (e.g. [here](https://swagger.io/docs/specification/using-ref/) and [here](https://github.com/OAI/OpenAPI-Specification/issues/1843)), but it occurred to me that (I think...?!) defaults would be legal in the structure produced when using `UseAllOfToExtendReferenceSchemas`. However they aren't currently supported there: they remain `null`. I think implementing this is do-able, also in a lightweight manner, and compatible with #1843

I'd like to finish that and offer it in another PR so you can see what I'm talking about! If it works (and if you agree that it's acceptably non-invasive, and also legal according to the relevant standards) then it would be a different solution to allowing string enum defaults, using `UseAllOfToExtendReferenceSchemas` instead of `UseInlineDefinitionsForEnums`. It would need to be combined with #1843 to be a fully working solution (i.e. to also respect name conversions), in which case there would be two fully working, valid solutions to the above problem, available to anyone needing them - which I think would be good!

domaindrivendev:
Hi @MikeBeaton. Thanks for all the work you're putting into this - it's much appreciated, and apologies for my lack of responsiveness in getting back to you. Unfortunately, I can only feasibly give so many hours to this project. But this is an area that needs to be addressed, so if you bare with me I will get back to you as quickly as I can.

For this particular issue, i.e. not setting defaults in the context of `UseAllOfToExtendReferenceSchemas`, I need some time to dig a little deeper because the SchemaGenerator was _designed_ to handle this case, and if it's not I'd prefer to fix that design rather than temoprarily changing the schema state to work around it.

To help me with this, perhaps you could post a code sample that repro's the issue?

[...]

PR #1851, opened 2020-10-02
Make defaults work with `UseAllOfToExtendReferenceSchemas`

MikeBeaton:
This addresses #1850. i.e. it allows `OpenApiSchema.Default` values to work when using `UseAllOfToExtendReferenceSchemas`, which I think is good/needed/helpful for the reasons mentioned there.

Most of the changed lines here are [the tests](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/pull/1851/commits/778fb986d633072ed22e6cf3aa9f40bc5041e055), the [change itself](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/pull/1851/commits/014af8c3e64edf7d131f0e38fc93a2edf3090ec2) is simple - while generating a schema in which `Reference` is being replaced with `AllOf`, temporarily set the schema type to the referenced type, so that attribute processing can work properly, then clear it again when done.

Note that the code path for setting defaults was always running in this case anyway, but just didn't have quite enough info to actually set the value.

This produces JSON like:

```js
{
 "name": "matchType",
 "in": "query",
 "schema": {
   "allOf": [
     {
       "$ref": "#/components/schemas/AutocompleteMatchType"
     }
   ],
   "description": "Allowed match types",
   "default": "BeginFirst"
  }
}
```

`"default"` would have always been null, even though specified by the API, without this suggested change.

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/OpenApiSchemaExtensions.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/OpenApiSchemaExtensions.cs
@@ -9,7 +9,9 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
 {
     public static class OpenApiSchemaExtensions
     {
-        public static void ApplyCustomAttributes(this OpenApiSchema schema, IEnumerable<object> customAttributes)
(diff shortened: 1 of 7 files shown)


MikeBeaton:
Note: This needs to be merged with #1843 in order to _both_ support defaults with refs (i.e. this PR), _and_ support renaming string enum defaults when correct to do so (#1843).

MikeBeaton:
(Force-pushed forwards and back again to make this PR rebuild and show as passing with now-fixed master branch - but haven't actually updated it - thanks!)

PR #1852, opened 2020-10-03
Merged defaults fixes: defaults work with $ref/allOf and with string enum member renaming

MikeBeaton:
This PR merges #1843 and #1851.

It allows default values to work generally with `UseAllOfToExtendReferenceSchemas`, and allows string enum defaults to respect all active enum member name transforms (e.g. camel casing, attribute-based enum member renaming) in all relevant (i.e. non-$ref) cases, namely when using either of `UseInlineDefinitionsForEnums` or `UseAllOfToExtendReferenceSchemas`.

Using `UseInlineDefinitionsForEnums`, v5 already allows

```js
{
  "name": "matchType",
  "in": "query",
  "schema": {
    "enum": [
      "begin",
      "any",
      "beginFirst"
    ],
    "type": "string",
    "description": "Allowed match types",
    "default": "BeginFirst"
  }
}
```

e.g. when referencing an enum with values `Begin`, `Any` and `BeginFirst` in C#, with camel casing turned on: the string enum default is present as of v5, which is great, but is only fully correct if there is no enum member name remapping, e.g. in the above it has not been camel cased, unlike the enum members values which have.

The intention of #1843 is to change this to

```js
{
  "name": "matchType",
  "in": "query",
  "schema": {
    "enum": [
      "begin",
      "any",
      "beginFirst"
    ],
    "type": "string",
    "description": "Allowed match types",
    "default": "beginFirst"
  }
}
```

where the camel case setting is now respected, and in general with `UseInlineDefinitionsForEnums` string enum default values now fully support all enum member name remapping.

#1851 additionally allows defaults to work when using `UseAllOfToExtendReferenceSchemas`. When merged with #1843, we can now alternatively choose to get:

```js
{
 "name": "matchType",
 "in": "query",
 "schema": {
   "allOf": [
     {
       "$ref": "#/components/schemas/AutocompleteMatchType"
     }
   ],
   "description": "Allowed match types",
   "default": "beginFirst"
  }
}
```

i.e. a string enum default value in an allOf/$ref schema, which correctly respects all enum member name mapping.


MikeBeaton:
While making this merge I realised that `PropertyInfo` is another source of default values, and required an additional one line change to also now support correct enum member re-mapping, plus tests for this added too: 5af62d2 & 10a6b5f .

MikeBeaton:
I've just force pushed from 5af62d2 to a7ebfdb and back again because AppVeyor failed to pull from GitHub the first time, for whatever reason, and listed the build as failed! :-/ Pushing it back and forwards again has made it retry and succeed.

MikeBeaton:
I've update this merge of both separate PRs again to confirm it all works.

Also NB that as before there are tests here (i.e. only in this merged PR, because they have to be) which confirm that both PRs work together nicely.

I am noticing (as I'm sure you will!) that `SchemaRepository` and `DataContract` are getting passed around together a lot - which is probably saying something! But not necessarily anything that needs to be addressed right now.

PR #1884, opened 2020-10-30
Refactored schema mapping and generation API

jcracknell:
I've taken a stab at implementing an improved interface for schema mappings that supports recursion and re-entry into the schema generation process. This permits mapping of generic types with arbitrary type parameters as well as other complex type hierarchies.

This is a substantial refactoring of the schema generation code, however I have attempted to preserve all existing behaviors (or at least all of those covered by tests).

Central to this process is the introduction of SchemaMapping as an abstraction providing both a SchemaDisposition (Reference or Inline) and a method to generate the requested schema. The existing SchemaGenerator implementation already reflects the fact that the disposition must be known prior to schema  generation in order to handle recursion properly, so we might as well formalize this:

https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/9b17843afe8a997052db6541ed5283621ea0143b/src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/SchemaGenerator.cs#L103-L108

You then define ISchemaMappingProvider:

    interface ISchemaMappingProvider {
      bool TryGetMapping(Type type, out SchemaMapping schemaMapping);
    }

as a way to check if the type is supported and resolve the disposition in one shot (as typically you would need to use the same reflective process for both). This closely parallels the existing behavior of SchemaGenerator.GenerateSchemaForType.

It's then intuitive to make the SchemaGenerator a sort of "fallback schema mapping" responsible for deriving schemas for arbitrary types, and transfer schema management responsibilities (namely registration and filtering) to the SchemaRepository, which becomes the primary schema management interface.

Obviously this represents a fairly substantial changeset, however there is also a lot of value here. I will note that:

  - This implements breaking changes to every filter interface, however I like to think that filter users are probably capable of redirecting the appropriate calls to the SchemaRepository.

  - Filters are required to be idempotent and can run more than once on a schema - mostly because there is no way to determine whether or not a new schema is created for properties/parameters.

There are still a few rough edges here as the SchemaRepository is now dependant on the configured SchemaGeneratorOptions, however I wanted to be particularly conservative around breaking changes to the configuration and wireup process so that it can be refactored as part of a major update.

Some other thoughts that I decided not to tackle at the moment:

  - I'd like ISchemaMappingContext to provide a something like `bool TryGetExample(Type type, out IOpenApiAny example)` - possibly via the SchemaRepository, but haven't investigated it in any great detail.

  - UseAllOfToExtendReferenceSchemas desperately needs to go, as the default behavior is fundamentally broken. I would replace it with ISchemaInliningStrategy with implementations {No,AllOf}SchemaInliningStrategy.

  - I would consider adding some kind of abstraction over member/property metadata that would permit you to unify application of ParameterInfo/MemberInfo/DataProperties, and also provide a single pipeline for reasoning about e.g. nullability.

  - If you really want to prevent running filters multiple times, you'd need to construct a collection of mutations and then check if it is empty. Even then there are no guarantees with inline schemas, so you're best off just requiring idempotent filters.

  - Consider adding I{Member,Parameter,Type}SchemaFilter interfaces and deprecate ISchemaFilter.

  - The current polymorphic schema generation behavior is really sketchy and brittle, however you could now ship it as an optional ISchemaMappingProvider so you're not married to it forever.


domaindrivendev:
This is a complete rewrite of the schema generation process and to be honest there is not a chance in hell of me accepting such a significant change to the codebase.

I suggest we take a step waaaaaayyyyyyy back, up out of the implementation weeds, and have you create an issue that articulates the exact problem you're trying to solve. I understand you're encountering limitations with the `MapType` setting but I don't understand what you're trying to use it for in the first place. That particular setting is intended for very simple use cases, and there's other extensibility options for some more complex cases. If you could explain with a basic API project example exactly what you're wanting to do, I think that will help us get to a solution, and hopefully a far more modest refactor than what you've presented here.

jcracknell:
I was afraid of that.

The current type mapping interface allows you to register a mapping for an open generic type, however in practice it's useless to do so: as there is no way to re-enter the schema generation process, you are required to define a single schema for all possible type arguments:

```cs
class Container<T> {
    public T Thing { get; set; }
}

options.MapType<Container<>>(() => new OpenApiSchema {
    Type = "object",
    Properties = {
        // What do I do here?
        ["Thing"] = throw new InvalidOperationException()
    }
});
```

The same problem applies to a mapping for any kind of complex type hierarchy unless you want to write out the schema for the entire hierarchy at every possible entry point for the hierarchy.

Per the design of the current SchemaGenerator (see above), the mapping process also has to produce a combination of what I am calling a "disposition" and a lazy schema factory. Currently all custom mappings produce inline schemas, so if your mapped type contains a reference to itself at any point in its hierarchy, you get a stack overflow trying to generate the schema.

As noted in my other issue, you can kind of get this working using an ISchemaFilter as it exposes the SchemaGenerator and SchemaRepository, however this requires a *lot* of hackery - to the extent that I'm worried a future update could permanently break my mappings.

I can "fix" this for my purposes relatively easily by copying, patching and replacing the SchemaGenerator via DI  (mostly because I want all of my mappings to produce reference schemas and I don't care about autogenerated polymorphic schemas), however I would prefer to arrive at a solution for everyone, as this seems like a desirable feature.

It's possible I got too invested in trying prevent repeatedly resolving factories for recursive reference schemas. When I have some time I'll try and see if I can get this working without splitting the SchemaGenerator, which should dramatically reduce the amount of churn.

jcracknell:
I was afraid of that.

The current type mapping interface allows you to register a mapping for an open generic type, however in practice it's useless to do so: as there is no way to re-enter the schema generation process, you are required to define a single schema for all possible type arguments:

```cs
class Container<T> {
    public T Thing { get; set; }
}

options.MapType<Container<>>(() => new OpenApiSchema {
    Type = "object",
    Properties = {
        // What do I do here?
        ["Thing"] = throw new InvalidOperationException()
    }
});
```

The same problem applies to a mapping for any kind of complex type hierarchy unless you want to write out the schema for the entire hierarchy at every possible entry point for the hierarchy.

Per the design of the current SchemaGenerator (see above), the mapping process also has to produce a combination of what I am calling a "disposition" and a lazy schema factory. Currently all custom mappings produce inline schemas, so if your mapped type contains a reference to itself at any point in its hierarchy, you get a stack overflow trying to generate the schema.

As noted in my other issue, you can kind of get this working using an ISchemaFilter as it exposes the SchemaGenerator and SchemaRepository, however this requires a *lot* of hackery, to the extent that I'm worried a future update could permanently break my mappings.

I can "fix" this for my purposes relatively easily by copying, patching and replacing the SchemaGenerator via DI  (mostly because I want all of my mappings to produce reference schemas and I don't care about autogenerated polymorphic schemas), however I would prefer to arrive at a solution for everyone, as this seems like a desirable feature.

It's possible I got too invested in trying prevent repeatedly resolving factories for recursive reference schemas. When I have some time I'll try and see if I can get this working without splitting the SchemaGenerator, which should dramatically reduce the amount of churn.

domaindrivendev:
As mentioned, the `MapType` setting is intended for very simple use cases. For the level of control you're looking for, I would recommend a good old "Decorator" pattern, where you wrap the out-of-the-box generator with your own custom version. This is clean, will work without any changes, and gives you the exact level of control you need. Here's how you would wire it up:

```csharp
// CustomerSchemaGenerator.cs
public class CustomSchemaGenerator : ISchemaGenerator
{
    private readonly SchemaGenerator _defaultSchemaGenerator;

    public CustomSchemaGenerator(
        SchemaGeneratorOptions generatorOptions,
        ISerializerDataContractResolver dataContractResolver)
    {
        _defaultSchemaGenerator = new SchemaGenerator(generatorOptions, dataContractResolver);
    }

    public OpenApiSchema GenerateSchema(
        Type type,
        SchemaRepository schemaRepository,
        MemberInfo memberInfo = null,
        ParameterInfo parameterInfo = null)
    {
        if (type.IsConstructedGenericType && type.GetGenericTypeDefinition() == typeof(TestDto<>))
        {
            // Custom mapping goes here
        }
        else
        {
            // Revert to default behavior
            return _defaultSchemaGenerator.GenerateSchema(type, schemaRepository, memberInfo, parameterInfo);
        }
    }
}
```

```csharp
// Startup.cs
services.AddSwaggerGen(c =>
{
    c.SwaggerDoc("v1", new OpenApiInfo { Title = "Test API", Version = "v1" });
});

// Replace the default generator with your custom implementation
services.Replace(ServiceDescriptor.Transient<ISchemaGenerator, CustomSchemaGenerator>());
```

jcracknell:
Thanks for your response - the problem with that approach is that once you pass control to the built-in SchemaGenerator, you never get it back (e.g. for object properties) because it is not aware of your wrapper. If you try to extend the generator, the methods you need to override are either private or non-virtual.

I don't think adding of a context object with type information and re-entry support to the mapping functionality is a huge undertaking, as you can retain support for the existing `Func<OpenApiSchema>` style mappings by just ignoring the context object. Unless you are dramatically opposed to the idea I'd still like to try and put together a less ambitious pull request for you to review.

domaindrivendev:
OK - now I understand the issue a little better thanks. Re creating another PR, I would hold off as I'm currently exploring a re-design of this area in general to better support things like "per-media type" schema's and nullable reference types. And, now that I understand your use-case, I'll take that and possibly some of the design approaches you've suggested into consideration. On that, I hope to have something in place in the next couple of weeks.

domaindrivendev:
In the meantime though, it would be very helpful for me if you could create a minimal project that highlights exactly what you're trying to do on the API side that is driving the need for this level of customization, and post it to github. Having a minimal reference project like this is the most effective way for me to truly understand your underlying use case, and ensure it's accounted for in the new design.

jcracknell:
I don't know that a whole project is necessary; the the heart of the issue is supporting mappings for generic types with arbitrary (unknown) type parameters. This requires the ability to map reference schemas, as well as a mapping context providing:

  - information about the constructed/type-parametrized type being mapped, and
  - a way to re-enter the schema mapping/generation process (because you don't know the schema for a type argument, or often even just other types).

It just happens that re-entry is also very useful for mapping complex, non-generic types as well.

Current tests for generic types are pretty contrived because they don't deal with the type parameters at all:
https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/9b17843afe8a997052db6541ed5283621ea0143b/test/Swashbuckle.AspNetCore.SwaggerGen.Test/SchemaGenerator/SystemTextJsonSchemaGeneratorTests.cs#L333-L335

A simple mapping would look something like the following:

```cs
options.MapType(typeof(Container<>), SchemaDisposition.Reference, context => new OpenApiSchema {
    Type = "object",
    Properties = {
        // Generic argument type provided by property type
        ["thing"] = context.GetMemberSchema(context.Type.GetProperty("Thing"))
    }
});
```

Some immediately applicable examples that I am trying to implement include paged search results for various entities:

```cs
public class PagedResponse<A> {
  public int PageSize { get; set; }
  public int PageNumber { get; set; }
  public int PageSize { get; set; }
  public int PageCount { get; set; }
  public int MatchCount { get; set; }
  public List<A> Results { get; set; }
}
```

And trees of composable search criteria as a highly-recursive example:

```cs
public abstract class SearchCriteria<R> {
  public class Conjunction : SearchCriteria<R> {
    public List<SearchCriteria<R>> And { get; set; }
  }

  public class Disjunction : SearchCriteria<R> {
    public List<SearchCriteria<R>> Or { get; set; }
  }

  public class Negation : SearchCriteria<R> {
    public SearchCriteria<R> Not { get; set; }
  }

  public class Requirement : SearchCriteria<R> {
    public R Req { get; set; }
  }
}

objects.MapType(typeof(SearchCriteria<>), SchemaDisposition.Reference, context => new OpenApiSchema {
  OneOf = {
    new OpenApiSchema {
      Type = "object",
      Properties = {
        ["$and"] = new OpenApiSchema {
          Type = "array",
          // You could just generate a reference here, but you might as well make the recursion
          // explicit because the context has to be able to figure this out in case the invocation
          // with this specific type occurs INSIDE of this call.
          Items = context.GetTypeSchema(context.Type)
        }
      }
    },
    new OpenApiSchema {
      Type = "object",
      Properties = {
        ["$or"] = new OpenApiSchema {
          Type = "array",
          Items = context.GetTypeSchema(context.Type)
        }
      }
    },
    new OpenApiSchema {
      Type = "object",
      Properties = {
        ["$not"] = context.GetTypeSchema(context.Type)
      }
    },
    new OpenApiSchema {
      Type = "object",
      Properties = {
        // Generate schema for the type argument, which is unknown
        ["$req"] = context.GetTypeSchema(context.Type.GetGenericArguments()[0])
      }
    }
  }
});
```

This is getting complex, but it's also when you would create and register an `ISchemaMappingProvider` or whatever instead.

PR #1887, opened 2020-11-03
#1886 Include properties defined in base interface in SchemaGenerator

### LINKED ISSUE #1886: Missing properties defined in base interface in SchemaGenerator
If the method in ASP.NET Core WebAPI controller returns Interface that is inherited from base interface, properties defined in base interface are missing.

### Example

ASP.Net Core WebAPI application with controller as below:

```csharp
    [ApiController]
    [Route("[controller]")]
    public class SampleController : ControllerBase
    {
        private readonly ExtendedUser extendedUser = new ExtendedUser
        {
            FirstName = "Johnson",
            LastName = "John",
            MiddleName = "Jay"
        };

        /// <summary>
        /// Sample method
        /// </summary>
        /// <returns></returns>
        [HttpGet]
        public ActionResult<IExtendedUser> GetExtendedUser() => extendedUser;
    }

    public interface IUser
    {
        public string FirstName { get; set; }
        public string LastName { get; set; }
    }

    public interface IExtendedUser : IUser
    {
        string MiddleName { get; set; }
    }

    public class ExtendedUser : IExtendedUser
    {
        public string FirstName { get; set; }
        public string LastName { get; set; }
        public string MiddleName { get; set; }
    }
```

The real GET request will return JSON with all properties:

```json
{"firstName":"Johnson","lastName":"John","middleName":"Jay"}
```

But schema for `IExtendedUser` will contain only `middleName` property.

```json
  "components": {
    "schemas": {
      "IExtendedUser": {
        "type": "object",
        "properties": {
          "middleName": {
            "type": "string",
            "nullable": true
          }
        },
        "additionalProperties": false
      }
    }
  }
```

Properties `firstName` and `lastName` are missing, though are returned in HTTP response.

navferty:
fixes #1886

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/SystemTextJsonBehavior.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/SystemTextJsonBehavior.cs
@@ -122,7 +122,7 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
         {
             extensionDataType = null;
 
-            var applicableProperties = objectType.GetProperties(BindingFlags.Instance | BindingFlags.Public | BindingFlags.NonPublic)
+            var applicableProperties = GetProperties(objectType)
                 .Where(property =>
                 {
                     return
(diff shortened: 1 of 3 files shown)


domaindrivendev:
Looks good. The only feedback I would have is to add the tests to the existing `SystemTextJsonSchemaGeneratorTests` class rather than introducing a new test class just for these tests. With that said, this is a simple change that I can add very quickly so rather than going through the back and forth, I'm going to merge your PR and make the test change in a quick follow-up commit

PR #1903, opened 2020-11-17
Ignore case when serving /index.html in redoc

### LINKED ISSUE #1902: Redoc Index.html url is case sentive, even when redirected
Using version 5.6.3, a site with the route prefix "api-doc" will redirect a request to the the URL http://localhost/api-doc to http://localhost/api-doc/index.html, which serves the redoc ui.

Thus a request to http://localhost/Api-Doc will redirect to http://localhost/Api-Doc/index.html, which will return a 404, since the check for index.html is not made case insensitive.

I will submit a PR with test shortly.

grstearns:
Fixes #1902, making the check for {routeprefix}/index.html case insensitive like /{routeprefix}.

I also noticed the regex escape in the redirect url check. Should that also be applied here?


domaindrivendev:
👍

Issue #1854, opened 2020-10-03
Default oauth2RedirectUrl problem with query parameter


walik92:
Swashbuckle.AspNetCore Version: 5.6.3
ASP.NET Core Version: 3.1

I have two versions of API and at the top of page in Swagger I can change definition v1 or v2. When I'm changing to v2, swagger set query param in browser `?urls.primaryName=v2`.
When I click Authorize button, swagger generate redirect_uri `http://localhost:5000/swagger/oauth2-redirect.html?urls.primaryName=v2`, but my authorization server has in configuration redirect_uri `http://localhost:5000/swagger/oauth2-redirect.html` and I get error Invalid redirect_uri.

I found code which generate redirect_uri in index.html 
`if (!configObject.hasOwnProperty("oauth2RedirectUrl"))
            configObject.oauth2RedirectUrl = window.location.href.replace("index.html", "oauth2-redirect.html").split('#')[0];`

I had to customize index.html and a change is very easy and small
`.split('#')->.split('?')`

Is it bug? What do you think?

ulrichb:
Same here. Why is the `oauth2RedirectUrl` overwritten at all? The default value (see https://github.com/swagger-api/swagger-ui/blob/7a0c1ea556a4b393bace21f80a5c65acaed7a49f/src/core/index.js#L39) already uses the current `window.location` and does _not_ pass the query string.

ulrichb:
@domaindrivendev  The relevant commits are
* 893aa299547207fc74455399ec4292bd8aa8902f, and
* aa682fa11e5cea2a6aa27f0c15e11c7cb42c3c8d, and
* 2dbb80765a1c47402dcbe05268d46a4a3a6dab9f

GGJason:
Facing the same problem here.

GGJason:
After deeply thinking, I thought that this is not a bug.
Because the query parameter is basically used to send information to the backend or the application to perform more jobs. So the parameter need to be kept for more actions in some scenario.
Maybe it could be an option can be set in Startup.cs.

GGJason:
I add a parameter and some script to implement the idea I wrote previously. I create a pull request #1911 .

ulrichb:
> After deeply thinking, I thought that this is not a bug.

[...]

PR #1911, opened 2020-11-27
Add Skip Query Parameter Option In Swagger UI Configuration

GGJason:
To make the package more flexible in response of Issue #1854 
 I add a SwaggerUIOptions ConfigObject property name 'IgnoreOAuth2RedirectUrlQueryParameters' with default false.
``` C#
app.UseSwaggerUI(options =>
{
    options.ConfigObject.IgnoreOAuth2RedirectUrlQueryParameters = true;
}
```
If the property is true, in index.html the script will just keep the url before query '?'.
``` JavaScript
if (configObject.ignoreOauth2RedirectUrlQueryParameters)
    configObject.oauth2RedirectUrl = configObject.oauth2RedirectUrl.split('?')[0];
```


domaindrivendev:
See my comments in related issue

PR #1922, opened 2020-12-08
Include document name in filter context

Blackbaud-ChristiSchneider:
Includes PR #1480 which was never reviewed plus the rest of the filter types that were left out of that review.

Add the document name to each filter context so the filter can be applied conditionally on each document. 

For schemas, rather than including it on the `SchemaFilterContext`, I included it on the `SchemaRepository` which I can change if you don't like that. The main reason I did that was because this is already being passed around everywhere for the schema generator and is included in the `SchemaFilterContext`. The `SchemaRepository` is also passed into the other filter `Apply` functions but it seems a bit dirty to use the schema repository for the other filter types.

I provided a default value of `null` for the publicly visible constructors and functions but can change it to require the parameter instead.

I wrote integration tests for this but can change how they were implemented. I was not really sure if I should include them in one of the existing test services or make a new one.

This will enable the following StackOverflow posts:
- [How can a filter be applied to a single swagger document with Swashbuckle](https://stackoverflow.com/questions/65203234/how-can-a-filter-be-applied-to-a-single-swagger-document-with-swashbuckle)
- [How to add a Schema Filter to just one Swashbuckle api document version](https://stackoverflow.com/questions/59093955/how-to-add-a-schema-filter-to-just-one-swashbuckle-api-document-version)

In addition to the above scenarios where this would be helpful, this would also eliminate the need for some preserialization filters. My service has 3 preserialization filters that all do something based on which document was requested, which you can only get from the HttpRequest context at this time because the change in this PR isn't available. If this change was put in, I could get rid of those 3 preserialization filters which would improve performance as well as make it so that I can get the document outside the context of a request.


Blackbaud-ChristiSchneider:
@domaindrivendev thoughts on this contribution?

domaindrivendev:
It's definitely not ideal to piggyback this onto the `SchemaRepository` object but it seems the only alternative would be to modify the `ISchemaGenerator` interface, which is something I'd like to do for this and a number of other reasons but given the backlog of PRs/issues is best left on the backburner.

I don't feel it's any _less_ dirty doing to this for schema filters as opposed to the other filter types and so I propose we accept the broader design issues at play, and use `SchemaRepository` to pass the value around for all filter types. However, I would make the property `internal` and expose the value via a `public` property on the various context objects. For example:

```csharp
public DocumentName => SchemaRepository.DocumentName;
```

This way, we uphold the public surface area that we'd really like for all filters, and encapsulate the "dirty" (and hopefully temporary) implementation going on under the hood.

domaindrivendev:
Regarding the intergration tests, I feel the logic here is far too trivial to warrant integration tests. I'm all for automated tests, but I also believe every test comes at an ongoing maintenance cost. I'd be OK weith no tests for this change, or alternatively I would look at incorporating this as an additional assertion in the existing filter option tests:

- https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/master/test/Swashbuckle.AspNetCore.SwaggerGen.Test/SwaggerGenerator/SwaggerGeneratorTests.cs#L731
- https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/master/test/Swashbuckle.AspNetCore.SwaggerGen.Test/SwaggerGenerator/SwaggerGeneratorTests.cs#L766
- https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/master/test/Swashbuckle.AspNetCore.SwaggerGen.Test/SwaggerGenerator/SwaggerGeneratorTests.cs#L801
- https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/master/test/Swashbuckle.AspNetCore.SwaggerGen.Test/SwaggerGenerator/SwaggerGeneratorTests.cs#L829

Blackbaud-ChristiSchneider:
Excellent, I will make these changes. Thank you!

Blackbaud-ChristiSchneider:
Okay, I've updated per your comments. I don't know why the csproj file has ID differences, probably just Visual Studio being Visual Studio, but if you wouldn't mind opening it on your machine and build/test that would be great. 

I did tests in the places you suggested and also added a new test for ISchemaFilter. I'm not completely sure that it's in the right test file but it is very similar to the other filter tests so I thought it made sense. The other file that has a test that uses ISchemaFilter is around json generation, which doesn't seem like the right place.

test/Swashbuckle.AspNetCore.SwaggerGen.Test/SwaggerGenerator/SwaggerGeneratorTests.cs
```diff
@@ -845,7 +851,53 @@ public void GetSwagger_SupportsOption_DocumentFilters()
...
+                        new VendorExtensionsSchemaFilter()
+                    }
+                }, new SystemTextJsonBehavior(new JsonSerializerOptions()))
+            );
+
+            var document = subject.GetSwagger("v1");
+
+            Assert.Equal(2, document.Components.Schemas["XmlAnnotatedType"].Extensions.Count);
+            Assert.Equal("bar", ((OpenApiString)document.Components.Schemas["XmlAnnotatedType"].Extensions["X-foo"]).Value);
+            Assert.Equal("v1", ((OpenApiString)document.Components.Schemas["XmlAnnotatedType"].Extensions["X-docName"]).Value);
```
domaindrivendev, line 900: This should go in the `SystemTextJsonSchemaGeneratorTests` class. But that's OK, I can move it to thr right place rather than more back n forth.

domaindrivendev: Looks good thx 👍

Issue #1829, opened 2020-09-22
Inconsistent Dependency Injection for SwaggerUIOptions and SwaggerOptions Compared to SwaggerGenOptions


Vaccano:
Dependency Injection for `SwaggerUIOptions` and `SwaggerOptions` is inconsistent with `SwaggerGenOptions`.

When setting up `SwaggerGenOptions` using `IConfigureNamedOptions`, it works as I would expect.  Meaning that the `Configure` method of the class implementing `IConfigureNamedOptions` is called, then the lambda from `AddSwaggerGen` is called with the `SwaggerGenOptions` object that was the modified in the `IConfigureNamedOptions.Configure` method..

But for `SwaggerUIOptions` and `SwaggerOptions` it is either one or the other.  If you provide a lambda, then the `IConfigureNamedOptions` is not called.

This makes it hard to create libraries that expand the settings.  It means that any such library has to take ownership of ALL the settings.  Or expose non-standard callbacks, etc.

Here is a walk through on how to repro this issue:

1. Create a basic API and add Swashbuckle.AspNetCore 5.5.1.
2. Create a class like this:

        public class ExtraSwaggerUiOptions : IConfigureNamedOptions<SwaggerUIOptions>
        {
            public void Configure(SwaggerUIOptions options)
            {
                    options.DocumentTitle = "Configure was Called";
            }

            public void Configure(string name, SwaggerUIOptions options)
            {
                     options.DocumentTitle = "Configure was Called";
            }
        }
3. Register it in `ConfigureServices` like this:

     services.AddTransient<IConfigureOptions<SwaggerUIOptions>, ExtraSwaggerUiOptions >();

4. In `Configure` update it to have the following code:

            app.UseSwaggerUI(options =>
            {
                options.SwaggerEndpoint("/swagger/v1/swagger.json", "OpenAPI Authentication Demo");
            });

Result: The DocumentTitle is not modified.
Expected: Both the DocumentTitle and the SwaggerEndpoint would be modified. (With the `Configure` method being called first.)

As I said above,  The "Expected" result is how `SwaggerGenOptions` works.  This issue is to try to get `SwaggerUIOptions` and `SwaggerOptions` to align with `SwaggerGenOptions`

Vaccano:
I have worked around this issue by creating an extension method on `SwaggerUIOptions` that requires the developer to pre-inject the needed information.  Not very nice, but it could be worse.  

I added this comment:

[...]

PR #1955, opened 2021-01-13
Consistent dependency injection of SwaggerOptions and SwaggerUIOption…

nuttytree:
…s when using a setup action on the UseSwagger and UseSwaggerUI methods.

### DIFF
--- a/src/Swashbuckle.AspNetCore.Swagger/DependencyInjection/SwaggerBuilderExtensions.cs
+++ b/src/Swashbuckle.AspNetCore.Swagger/DependencyInjection/SwaggerBuilderExtensions.cs
@@ -24,15 +24,8 @@ namespace Microsoft.AspNetCore.Builder
             this IApplicationBuilder app,
             Action<SwaggerOptions> setupAction = null)
         {
-            SwaggerOptions options = new SwaggerOptions();
-            if (setupAction != null)
-            {
-                setupAction(options);
-            }
-            else
-            {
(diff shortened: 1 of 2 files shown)


nuttytree:
This PR resolves issue #1829

domaindrivendev:
@nuttytree - we need to revisit this PR. While it introduces some additional consistency, it removes the flexability to configure multiple instances of the middleware with different configurations. For example, the following configuration to emit v2 and v3 versions of the Swagger/OpenAPI JSON side by side is now broken:

```csharp
app.UseSwagger(c =>
{
    c.RouteTemplate = "swagger/{documentName}/swagger.json";
    c.SerializeAsV2 = true;
});

app.UseSwagger(c =>
{
    c.RouteTemplate = "/swagger/{documentName}/openapi.json";
});
```

nuttytree:
I just got done with really rough day at work trying to track down a very problematic bug.  I will take a look tomorrow and see if I can come up with a solution.

domaindrivendev:
@nuttytree no worries - I went ahead and pushed the following change - b6c6919c8fc19394f767d6159bb6286036e2e843, which feels like a decent compromise. Basically, if you use the overload that accepts a setup action `Action<SwaggerOptions>` then your using the DI-injected options. Alternatively, if you want to configure multiple instances of the middleware with independent options, then you use the overload that accepts a `SwaggerOptions` instance.

PR #1967, opened 2021-01-21
Target net461

aliabusaleh:
Hi!, 
I was trying to use Swashbuckle.AspNetCore with .Net framework 4.6.1 (as my project built using it) and I made these changes to get Swashbuckle.AspNetCore dll files and use them inside net461.

### DIFF
--- a/src/Swashbuckle.AspNetCore.Annotations/AnnotationsDocumentFilter.cs
+++ b/src/Swashbuckle.AspNetCore.Annotations/AnnotationsDocumentFilter.cs
@@ -1,9 +1,9 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Linq;
-using Microsoft.AspNetCore.Mvc.Controllers;
+﻿using Microsoft.AspNetCore.Mvc.Controllers;
 using Microsoft.OpenApi.Models;
 using Swashbuckle.AspNetCore.SwaggerGen;
+using System;
+using System.Collections.Generic;
+using System.Linq;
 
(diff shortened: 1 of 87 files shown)


domaindrivendev:
A few comments ...

Firstly, please don't include subjective stylistic changes (e.g. ordering of using statements etc.) in files that are unrelated to the PR - it makes it more difficult to follow the PR and it's a little presumpitive to think you're preferences would be preferred over what's already there.

Secondly, I'm a little confused by this PR in general. You're adding `net461` as a target but removing all of the other ones - meaning the library would only be available to the small subset of ASP.NET Core apps that target `net461` - or am I missing something?

aliabusaleh:
Exactly, my addition was to include the built-in libraries in ASP.NET core explicitly into the files, which make it working with net461.
for the comment about it's targeting the only net461, it's the point of these changes, as many users who have net461 project can't use Swashbuckle.AspNetCore as it's not supported net461 directly, so if these changes was put inside a branch for targeting net461 now anyone who has net461 project will be able to use it without any conflict by pulling only the target net461.
many thanks

domaindrivendev:
I didn't read through the full PR, but did you get a sense of how difficult it would be (I'm assuming a number of compiler directives) to target net461 _in addition to_ the existing frameworks with the same source code?

aliabusaleh:
yes, my idea is to make these changes into a new branch, if someone needs to target net461, it will be available on a branch without merging it into the master :D
thanks for your time and understanding

PR #1968, opened 2021-01-25
Support for Nullable Reference Types

### LINKED ISSUE #1686: Support for Nullable Reference Types
Are there any plans to add support for Nullable Reference Types? It would be nice, and with NRT we would no longer need to annotate required properties with the `[Required]` attribute.

wu-yafeng:
- Add options for suppress generate value `Nullable` by nullable-reference-type
- Add unit tests for fixed/pure NRT

Resolved #1686


wu-yafeng:
Created another pr.

Issue #1974, opened 2021-01-27
ISchemaFilter.Apply not called for reference schemas


maurei:
I'm writing a Swashbuckle integration for [JsonApiDotnetCore](www.jsonapi.net) (JADNC). Using a `IDocumentFilter` I iterate through `OpenApiDocument.Components.Schema` and manipulate every root schema as well as every nested property schema in `OpenApiDocument.Properties` according to the needs of a developer that uses JADNC. 

In order to reliably match every property schema to the actual model property it represents, I want to perform lookup of the `MemberInfo` associated with the property schema. These `MemberInfo`s are not exposed in the document filter by default, so I inject them in my document filter using a custom cache in which these `MemberInfo`s are stored during the execution of my `ISchemaFilter`. The mechanism is as follows:

```c#
public class JsonApiSchemaFilter : ISchemaFilter
{
    private readonly ISchemaMetadataCache _schemaMetadataCache;

    public JsonApiSchemaFilter(ISchemaMetadataCache schemaMetadataCache)
    {
        _schemaMetadataCache = schemaMetadataCache;
    }

    public void Apply(OpenApiSchema schema, SchemaFilterContext context)
    {
        // SchemaFilterContext.MemberInfo associated to the schema is what I need in my DocumentFilter
        _schemaMetadataCache.Set(schema, context)
    }
}

public class JsonApiDocumentFilter : IDocumentFilter
{
    private readonly ISchemaMetadataCache _schemaMetadataCache;

    public JsonApiDocumentFilter(ISchemaMetadataCache schemaMetadataCache)
    {
        _schemaMetadataCache = schemaMetadataCache;
    }

    public void Apply(OpenApiDocument swaggerDoc, DocumentFilterContext context)
    {
        foreach (var (rootKey, rootSchema) in swaggerDoc.Components.Schemas)
        {
            foreach (var (propertyKey, propertySchema) in rootSchema)
            {
                SchemaFilterContext metadata = _schemaMetadataCache.Get(propertySchema);
                DoAllSortsOfCustomStuff(rootSchema, key, propertySchema, metadata.MemberInfo) 
            }
        }
    }
}
```

[...]

PR #1976, opened 2021-01-27
Run schema filters for reference schemas

maurei:
Opening up this PR to investigate the impact of this change. Feedback is welcome


domaindrivendev:
Can you provide more info around the problem you’re trying to solve?

maurei:
My bad: this one is for #1974

Issue #1970, opened 2021-01-26
Cache outputted OpenApiDocument


maurei:
It seems to me that some easy performance improvements can be made and redundancy can be removed in the default `ISwaggerProvider` implementation by caching the outputted `OpenApiDocument` by `documentName` rather than re-rendering each document upon a new request.

I guess in principle it would be possible to add new actions to the application after startup, and that the strategy of re-rendering the document would allow for picking up these changes. But this seems far fetched to me and as far as I could tell by the testing codebase, this use-case isn't thoroughly covered. I would expect rendering every `OpenApiDocument` once per application lifecycle is sufficient. Is there a reason I'm missing that explains why the output isn't being cached?

domaindrivendev:
IMO, in-memory caching is far too intrusive for a library to provide out-of-the-box, particularly in the days of cloud computing when increasing the memory footprint of an application comes at an ongoing cost. So, I've made the explicit decision to leave it up to application developers to implement caching should they feel the need. There's a number of different, relatively trivial approaches for doing this, but admitedly none of them are particularly well documented at the moment.

For now, I'll point you to the following comment in a related story, describing a "decorator" approach where you register a simple wrapper around the built-in `ISwaggerProvider` implementation that includes the in-memory, read-through caching:

https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/1565#issuecomment-599459249

Another approach would be to leverage the [CLI tool](https://github.com/domaindrivendev/Swashbuckle.AspNetCore#swashbuckleaspnetcorecli) to generate a static Swagger/OpenAPI JSON file at runtime and then serve it at runtime via ASP.NET Core's static file middleware (e.g. by saving it in `wwwroot/swagger/v1/swagger.json`). See the following for an example of this approach:

https://github.com/domaindrivendev/Swashbuckle.AspNetCore/tree/master/test/WebSites/CliExample

PR #2008, opened 2021-02-11
Add option that allows to cache OpenApiDocument for identical requests until applying PreSerializeFilters

unchase:
This will help to load the large OpenAPI documents much faster for the second and subsequent identical requests.


unchase:
@domaindrivendev ready for review.
Could you take a look? Thx

unchase:
Are there any reasons why this PR is not being reviewed?

domaindrivendev:
As I've mentioned in a number of related issues, I'm reluctant to introduce memory caching out of the box, because I believe it's intrusive to do so, and is a responsibility that is better left up to to the application itself. Admittedly the docs around this need work, but in the meantime I've suggested two relatively trivial approaches for layering in caching in the following ticket:

#1970

Issue #1996, opened 2021-02-08
Multiple UseSwaggerUI support removed?


effyteva:
Hi,

We're trying to migrate to the new 6.0.2 release,
We're currently using UseSwaggerUI twice, while each UI renders different API methods (one is used for the most common methods, and the other for all methods regardless of their usage).
The methods are excluded using the DocInclusionPredicate, based on the DocumentName.

After migrating to the new 6.0.2 release, only the second UseSwaggerUI call is actually working, and the first one seems to be completely ignored.
Is that intended behavior or a bug? Is there any other way to achieve multiple endpoints with different methods?

Thank you,
Effy

domaindrivendev:
Hi @effyteva. In the latest version, if you're configuring the `Swagger` or `SwaggerUI` middleware with the overload that accepts a setup action (e.g. `Action<SwaggerUIOptions>`) then you're implicitly using the _singleton instance_ of options from you're app's DI container (see the [ASP.NET Core Options pattern](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/options?view=aspnetcore-5.0) for more info on this).

So, for your usecase you'll need to use the overload that accepts a specific instance of options instead of the setup action:

```csharp
app.UseSwaggerUI(new SwaggerUIOptions
{
    RoutePrefix = "foo-docs",
    ConfigObject = new ConfigObject
    {
        Urls = new[] { new UrlDescriptor { Url = "/swagger/foo/swagger.json", Name = "Foo Docs"} }
    }
});

app.UseSwaggerUI(new SwaggerUIOptions
{
    RoutePrefix = "bar-docs",
    ConfigObject = new ConfigObject
    {
        Urls = new[] { new UrlDescriptor { Url = "/swagger/bar/swagger.json", Name = "Bar Docs"} }
    }
});
```

effyteva:
Thank you @domaindrivendev 
I appreciate your quick response, this worked perfect, and we've managed to upgrade to the newest release once again.

Effy

halloersmal:
That helped me out! Thank you so much for the ConfigObject usage Example! I worked for days to have two different Swagger UIs for different endpoints!

PR #2016, opened 2021-02-15
Handle multiple document versions in SwaggerUI and Redoc

SukharevAndrey:
Last changes to Swashbuckle returned back old options' mutation bug that prevents generating SwaggerUI and Redoc pages for separate swagger document versions.
Authors suggest to use a different overload of UseSwaggerUI and UseRedoc here: https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/1996
But I still think that singleton options mutation is misleading and users don't expect that. So in this PR I restored the old behaviour and also added new tests to ensure it will not break again.

### DIFF
--- a/src/Swashbuckle.AspNetCore.ReDoc/ReDocBuilderExtensions.cs
+++ b/src/Swashbuckle.AspNetCore.ReDoc/ReDocBuilderExtensions.cs
@@ -22,9 +22,12 @@ namespace Microsoft.AspNetCore.Builder
             this IApplicationBuilder app,
             Action<ReDocOptions> setupAction = null)
         {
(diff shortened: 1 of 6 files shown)


domaindrivendev:
This whole area has been a frustrating game of whackamole - hence the back and forth. Looking at some of the existing ASP.NET middleware, I've noticed most of them provide an overload for passing an explicit option instance and if this isn't provided, then the injected options instance is used. But none of them seem to support a hybrid approach - i.e. where you can use the injected options but also enrich them in the context of a given middleware.

So, the current API is a little odd in general, and ideally I'd like to simply remove the "hybrid" overlaod. But in lieue of that, yours is a very nice solution to support it without having the unexpected mutation behavior.

PR #2026, opened 2021-02-21
Add swagger annotation attributes for headers and multipart/form-data.

kikaragyozov:
The idea of this addition is that `SwaggerHeaderAttribute` allows you to define headers in the generated swagger documentation, while `SwaggerMultiPartFormData` allows defining multipart/form-data boundaries, for when a user needs to consume the data without the use of any model binding provided by the framework. _(`IFormFile`, `IFormCollection` or a more complex object containing these as properties.)_ Usually a user would do this if they require working directly with the underlying stream, used primarily for large files.

Please let me know of anything I should be aware of, as this is my first time contributing to this repository.

If this seems like a reasonable addition, I can also include some unit tests.

**Things to note:** Swashbuckle.AspNetCore.Filters includes a type `AddHeaderOperationFilter` that adds headers for **every** endpoint in the generated document. `SwaggerHeaderAttribute` can instead be applied for specific endpoints and/or controllers.


SpiritBob:
@domaindrivendev Could you let me know your thoughts about this?

martincostello:
Thanks for contributing - if you'd like to continue with this pull request, please rebase against the default branch to pick up our new CI. We'd also require tests.

PR #2042, opened 2021-03-02
Add status code descriptions

### LINKED ISSUE #1762: All 2xx response types are rendered with description "Success".
Version: 5.5.1

I decorated an Function with
`[ProducesResponseType(typeof(string), StatusCodes.Status201Created)]
[ProducesResponseType(typeof(string), StatusCodes.Status204NoContent)]
[ProducesResponseType(typeof(string), StatusCodes.Status400BadRequest)]`

When Rendered on the ui endpoint all of the Response types are rendered as
201   Success
204   Success
400   Bad Request

I would have expected a render with
201   Created
204   No Content
400  Bad Request
by default and at least a way to change the description, but I did not found it.

MikeFH:
Partially fixes #1762


src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
```diff
@@ -492,23 +492,31 @@ private OpenApiMediaType CreateResponseMediaType(ModelMetadata modelMetadata, Sc
...
+            new KeyValuePair<string, string>("404", "Not Found"),
+            new KeyValuePair<string, string>("405", "Method Not Allowed"),
+            new KeyValuePair<string, string>("406", "Not Acceptable"),
+            new KeyValuePair<string, string>("408", "Request Timeout"),
+            new KeyValuePair<string, string>("409", "Conflict"),
+            new KeyValuePair<string, string>("429", "Too Many Requests"),
+            new KeyValuePair<string, string>("4\\d{2}", "Client Error"),
+
+            new KeyValuePair<string, string>("5\\d{2}", "Server Error"),
+            new KeyValuePair<string, string>("default", "Error")
```
domaindrivendev, line 519: @MikeFH - I'm happy to include the specific descriptions for more of the status codes as you've done here. But, I'd prefer to stick with the more simplified `Dictionary<,>` initializers. I get the desire to make it readonly it's private to the class and IMO highly unlikely that anyting will try to mutate it. So, I'd opt for the reduced boilerplate/increased readability instead.

domaindrivendev: If you're happy to revert this and then resolve the conflicts, I'll go ahead and merge 👍

MikeFH: Ok, no problem. The thing I was worried about is that we rely on the order of the rules to find the right description and I don't think the generic dictionary guarantees the order of insertion is the order of enumeration but there's probably a low chance of it becoming an issue.

domaindrivendev: @MikeFH actually you bring up a really good point that hadn't occurred to me. I don't think the mapping logic should rely the order in which keys are enumerated. Thinking about alternatives, I wonder if we could scrap the static mapping altogether and leverage the `System.Net.HttpStatusCode` enum instead. Something along the following lines:

```csharp
 var description = int.TryParse(statusCode, out int intValue)
      ? Enum.GetName(typeof(HttpStatusCode), intValue)
      : "Error";
```

Maybe some additional logic to split out the enum names into more natural text. For example, "BadRequest" becomes "Bad Request" etc. Thoughts?

MikeFH: I looked a bit into it and I think it could be doable borrowing, for example, this method from Humanizer https://github.com/Humanizr/Humanizer/blob/main/src/Humanizer/StringHumanizeExtensions.cs#L26. But I noticed "429 Too many request" is missing from the HttpStatusCode enum. Maybe this could become a problem in the future if we want to support codes not in the enum.
In the end I wonder if we would not be better off hard coding all the codes from the enum plus others we need in the dictionary and simply get the description based on the key, removing the regex evaluation.

domaindrivendev:
For now, let's go with your original implementation which ensures enumeration of the dictionary in the correct order. I'll revert your last commit as part of the merge.

domaindrivendev:
@MikeFH - could you revert your last commit, then I'll go ahead and merge the PR

MikeFH:
Done ! I merged master into my branch to resolve a conflict.
However, do not merge this PR if you feel there is a better solution, I wouldn't mind.

PR #2054, opened 2021-03-11
Fixed missing parenthesis that could cause a property to be added as required multiple times.

Eneuman:
This PR fixes a "AND"/"OR" logic problem that could cause a property to be added as required multiple times.


domaindrivendev:
Good catch - thx

Issue #1487, opened 2020-01-16
[5.0.0] Respect Nullable context


mrmartan:
When C# 8 nullable context is enabled and ASP.NET Core 3 `MvcOptions.SuppressImplicitRequiredAttributeForNonNullableReferenceTypes == false` (the default) the generated API spec does not match C# semantics (e.g. `string` model field is marked as nullable whci is not true with C# nullable context enabled) and ASP.NET model validation behavior.

```xml
<Project Sdk="Microsoft.NET.Sdk.Web">

<PropertyGroup>
  <TargetFramework>netcoreapp3.0</TargetFramework>
  <Nullable>enable</Nullable>
</PropertyGroup>
```
```csharp
services
    .AddControllers(options =>
    {
        options.SuppressImplicitRequiredAttributeForNonNullableReferenceTypes = false;
    });
```
```csharp
namespace TestApp
{
    [ApiController]
    public class HomeController : ControllerBase
    {
        [HttpPost]
        [Route("welcome")]
        [ProducesResponseType(StatusCodes.Status200OK, Type = typeof(TestModel))]
        public async Task<ActionResult<TestModel>> WelcomeJsonModel(TestModel model)
        {
            return model;
        }
    }

    public class TestModel
    {
        public string TestString { get; set; } = null!;

        [Required]
        public int TestInt {  get; set; }

        public AnotherTestModel? NullableSubmodel { get; set; }
    }

    public class AnotherTestModel
    {
        public int TestInt2 { get; set; }
    }
}
```
![image](https://user-images.githubusercontent.com/2376580/72530092-e81d1700-386e-11ea-8715-4a25021e1057.png)

![image](https://user-images.githubusercontent.com/2376580/72530260-49dd8100-386f-11ea-9557-8decb6bb7b89.png)

ketovdk:
Not sure it is about required. It just does not detect that these types are not nullable. I tried to do something like 
`
			public class RequiredSchemaFilter: ISchemaFilter
			{
				public void Apply(OpenApiSchema schema, SchemaFilterContext context)
				{
					schema.Required ??= new HashSet<string>();
					schema.Required.AddRange(schema.Properties.Where(x => !x.Value.Nullable).Select(x => x.Key));`
But it detects not nullable types as nullable

foriequal0:
I use this.
```
    class NullableAttributeSchemaFilter : ISchemaFilter
    {
        private readonly MvcOptions mvcOptions;

[...]

PR #2056, opened 2021-03-12
Fixed problem with finding nullable attribute in nested types because of roslyn optimizations

### LINKED ISSUE #2017: 6.0.5 subclass non-nullable strings shown as nullable
Hi,

Great that the nullable reference types support is coming along! I tested it and found an issue that was a bit confusing.
When SupportNonNullableReferenceTypes is enabled:

```
public class TopLevelClass{
    public string TopNonNullableString { get; set; } = null!;
    public string? TopNullableString {get; set; }


    public class SubClass{
        public string SubNonNullableString { get; set; } = null!;
        public string? SubNullableString {get; set; }
    }
}
```

For the `TopLevelClass` the properties' nullable context is shown correctly, but for `SubClass` both are shown as nullable.
Now if I move the `SubClass` to be on the top level, the nullable context is shown correctly.

I'm on version 6.0.5

Eneuman:
This change only affects the logic in finding compiler annotated null attributes when using the "SupportNonNullableReferenceTypes" option.

- [x] Adjusted code to handle roslyn compiler optimizations for nested types.
- [x] Fixed bug that cause a crash if annotation nullable attribute flag returned was byte[].
- [x] Added tests for compiler optimization senarios.
- [x] Added default values to test class to remove warnings.

More information about roslyn nullable metadata can be found here https://github.com/dotnet/roslyn/blob/main/docs/features/nullable-metadata.md

This PR should fix #2017 and probobly #1487

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/MemberInfoExtensions.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/MemberInfoExtensions.cs
@@ -39,7 +39,10 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
                 ? ((FieldInfo)memberInfo).FieldType
                 : ((PropertyInfo)memberInfo).PropertyType;
 
-            if (memberType.IsValueType) return false;
+            if (memberType.IsValueType)
+            {
+                return false;
+            }
 
             var nullableAttribute = memberInfo.GetNullableAttribute();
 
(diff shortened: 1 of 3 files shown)


src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/MemberInfoExtensions.cs
```diff
@@ -39,7 +39,10 @@ public static bool IsNonNullableReferenceType(this MemberInfo memberInfo)
                 ? ((FieldInfo)memberInfo).FieldType
                 : ((PropertyInfo)memberInfo).PropertyType;
 
-            if (memberType.IsValueType) return false;
+            if (memberType.IsValueType)
+            {
+                return false;
+            }
```
domaindrivendev, line 45: Could you please revert this change as it's stylistic not functional - I prefer the previous style

Eneuman: Code change reverted

PR #2077, opened 2021-04-01
Fix ui url without index.html

### LINKED ISSUE #2076: OAuth2 redirect breaks in UI hosted by URL without index.html in path
We use URL rewriting to host Swagger UI by root URL "/" instead of "/index.html" to fix load balancer pings, because the load balancer can't follow 301-s to "/index.html" and it's a problem for us to ping another URL there.

```cs
app.UseRewriter(new RewriteOptions().AddRewrite("^$", "/index.html", skipRemainingRules: true));
```

This breaks OAuth2 redirect URL, because in Swashbuckle custom UI [index.html](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/master/src/Swashbuckle.AspNetCore.SwaggerUI/index.html) file the code expects UI is always hosted by URL ending with "/index.html":

```js
// If oauth2RedirectUrl isn't specified, use the built-in default
if (!configObject.hasOwnProperty("oauth2RedirectUrl"))
    configObject.oauth2RedirectUrl = window.location.href.replace("index.html", "oauth2-redirect.html").split('#')[0];
```

Steps to reproduce:

1. Install latest `Swashbuckle.AspNetCore` version
2. Configure Swagger and Swagger UI with OAuth2 in Startup
3. Add server-side URL rewrite from "/" to "/index.html"
4. Open Swagger UI and try to Authorize

**Expected**: after successful sign-in, user is redirected back to Swagger UI with authenticated state

**Actual**: after successful sign-in, user is redirected back to Swagger UI without authentication

I'm going to push a PR with the fix shortly.

Current workaround:

1. Create custom "index.html" file based on Swashbuckle [index.html](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/master/src/Swashbuckle.AspNetCore.SwaggerUI/index.html)
2. Make a fix in your file
   ```js
   configObject.oauth2RedirectUrl = (window.location.href.replace("index.html", "") + "oauth2-redirect.html").split('#')[0];
   ```
3. Load it in Swagger UI middleware options:

   ```cs
   var assembly = GetType().Assembly;
   
   app.UseSwaggerUI(c =>
   {
       c.IndexStream = () => assembly.GetManifestResourceStream("index.html");
   });
   ```

whyleee:
Fixes #2076

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerUI/index.html
+++ b/src/Swashbuckle.AspNetCore.SwaggerUI/index.html
@@ -79,7 +79,7 @@
             // Workaround for https://github.com/swagger-api/swagger-ui/issues/5945
             configObject.urls.forEach(function (item) {
                 if (item.url.startsWith("http") || item.url.startsWith("/")) return;
-                item.url = window.location.href.replace("index.html", item.url).split('#')[0];
+                item.url = (window.location.href.replace("index.html", "") + item.url).split('#')[0];
             });
 
             // If validatorUrl is not explicitly provided, disable the feature by setting to null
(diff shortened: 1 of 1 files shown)


domaindrivendev:
I believe this is already fixed by #2128

Issue #2091, opened 2021-04-15
If the JSON string contains bool value, using OpenApiAnyFactory.CreateFromJson() will get null in this field.


djjsdtc:
I was creating an example of my model using OpenApiAnyFactory.CreateFromJson(string). When the JSON contains bool value (true/false), I found the output example's field is filled by "null".

Example:
Expected: { "TestValue": true }
Actual result: { "TestValue": null }

Example 2:
```csharp
class Program
{
    static void Main(string[] args)
    {
        var testObj = new Test {TestValue = true};
        var openApiObj = OpenApiAnyFactory.CreateFromJson(JsonSerializer.Serialize(testObj)) as OpenApiObject;
        Console.WriteLine(openApiObj["TestValue"] == null);
        // Should be false, but actual output is true
    }
}

class Test
{
    public bool TestValue { get; set; }
}
```

I looked into the code of OpenApiAnyFactory.CreateFromJson(string). I found when converting bool values in an object, an exception is thrown and catched, like this:
```
System.Text.Json.JsonException: 'T' is an invalid start of a value. Path: $ | LineNumber: 0 | BytePositionInLine: 0.
 ---> System.Text.Json.JsonReaderException: 'T' is an invalid start of a value. LineNumber: 0 | BytePositionInLine: 0.
   at System.Text.Json.ThrowHelper.ThrowJsonReaderException(Utf8JsonReader& json, ExceptionResource resource, Byte nextByte, ReadOnlySpan`1 bytes)
   at System.Text.Json.Utf8JsonReader.ConsumeValue(Byte marker)
   at System.Text.Json.Utf8JsonReader.ReadFirstToken(Byte first)
   at System.Text.Json.Utf8JsonReader.ReadSingleSegment()
   at System.Text.Json.Utf8JsonReader.Read()
   at System.Text.Json.Serialization.JsonConverter`1.ReadCore(Utf8JsonReader& reader, JsonSerializerOptions options, ReadStack& state)
   --- End of inner exception stack trace ---
   at System.Text.Json.ThrowHelper.ReThrowWithPath(ReadStack& state, JsonReaderException ex)
   at System.Text.Json.Serialization.JsonConverter`1.ReadCore(Utf8JsonReader& reader, JsonSerializerOptions options, ReadStack& state)
[...]

PR #2092, opened 2021-04-15
JSON string with bool value conversion issue

djjsdtc:
See issue #2091

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/OpenApiAnyFactory.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/OpenApiAnyFactory.cs
@@ -49,9 +49,22 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
 
             foreach (var item in jsonElement.EnumerateArray())
             {
-                var json = item.ValueKind == JsonValueKind.String
-                    ? $"\"{item}\""
-                    : item.ToString();
+                string json;
+                switch (item.ValueKind)
+                {
+                    case JsonValueKind.String:
+                        json = $"\"{item}\"";
+                        break;
+                    case JsonValueKind.True:
+                        json = "true";
+                        break;
+                    case JsonValueKind.False:
+                        json = "false";
+                        break;
+                    default:
+                        json = item.ToString();
+                        break;
+                }
 
                 openApiArray.Add(CreateFromJson(json));
             }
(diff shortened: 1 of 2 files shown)


domaindrivendev:
Good catch - this is definitely a bug, and there could be other similar ones due to the use of `ToString()`. Regarding the fix, I think we can reduce the amount of duplicated logic, and get away from `ToString()` altogether by taking a slightly different approach ...

I think we can extract a new _private_ method `CreateFromJsonElement(JsonElement)` out of `CreateFromJson(string)`. For example, the latter would be reduced to the following:

```csharp
public static IOpenApiAny CreateFromJson(string json)
{
    try
    {
        var jsonElement = JsonSerializer.Deserialize<JsonElement>(json);

        return CreateFromJsonElement(jsonElement);
    }
    catch { }

    return null;
}
```

And, all of the `JsonElement` to `IOpenApiAny` mapping code would be moved into `CreateFromJsonElement(JsonElement)`. Then finally, the code in the object and array specific mapping methods could just call the new method. For example, `CreateOpenApiArray` could be simplified to the following:

```csharp
private static IOpenApiAny CreateOpenApiArray(JsonElement jsonElement)
{
    var openApiArray = new OpenApiArray();

    foreach (var item in jsonElement.EnumerateArray())
    {
        openApiArray.Add(CreateFromJsonElement(item));
    }

    return openApiArray;
}
```

I believe this will solve the issue and make the code a lot more resilient. Additionally, this PR should be accompanied by at least one test. You could probably just add an additional case to the following:
https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/master/test/Swashbuckle.AspNetCore.SwaggerGen.Test/SchemaGenerator/JsonSerializerSchemaGeneratorTests.cs#L267

djjsdtc:
> Good catch - this is definitely a bug, and there could be other similar ones due to the use of `ToString()`. Regarding the fix, I think we can reduce the amount of duplicated logic, and get away from `ToString()` altogether by taking a slightly different approach ...
> 
> I think we can extract a new _private_ method `CreateFromJsonElement(JsonElement)` out of `CreateFromJson(string)`. So, the latter would be reduced to the following:
> 
> ```cs
> public static IOpenApiAny CreateFromJson(string json)
> {
>     try
>     {
>         var jsonElement = JsonSerializer.Deserialize<JsonElement>(json);
> 
>         return CreateFromJsonElement(jsonElement);
>     }
>     catch { }
> 
>     return null;
> }
> ```
> 
> All of the `JsonElement` to `IOpenApiAny` mapping code would be moved into `CreateFromJsonElement(JsonElement)`. Then finally, the code in the object and array specific mapping methods could just call the new method. For example, `CreateOpenApiArray` could be simplified to the following:
> 
> ```cs
> private static IOpenApiAny CreateOpenApiArray(JsonElement jsonElement)
> {
>     var openApiArray = new OpenApiArray();
> 
>     foreach (var item in jsonElement.EnumerateArray())
>     {
>         openApiArray.Add(CreateFromJsonElement(item));
>     }
> 
>     return openApiArray;
> }
> ```
> 
> I believe this will solve the issue and make the code a lot more resilient. Additionally, this PR should be accompanied by at least one test. You could probably just add an additional case to the following:
> https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/master/test/Swashbuckle.AspNetCore.SwaggerGen.Test/SchemaGenerator/JsonSerializerSchemaGeneratorTests.cs#L267

Thanks for suggestion. I have moved the code outside to another method, and added unit tests for this method.

domaindrivendev:
This is great - thankyou 👍

Issue #2078, opened 2021-04-06
Allow CORS on swagger.json endpoint only


cyrildurand:
I can't find a way to enable CORS on the swagger.json endpoint only whithout applying CORS for whole API. 

I read the documentation here :  https://docs.microsoft.com/en-us/aspnet/core/security/cors?view=aspnetcore-3.1#enable-cors-with-attributes and I understand how I can enable cors for the whole API including the swagger.json endpoint but my external application only need to access my swagger.json definition.

It would be a nice feature if we can specify the CORS policy for the swagger.json endpoint : 

something like 

     app.UseSwagger(c => {
         c.EnableCors("MyCorsPolicy"); 
     });

or maybe 

     app.UseSwagger(c => {
         c.AddMetadataEndpoint(new EnableCorsAttribute("MyCorsPolicy")) 
     });

This question has also been posted on stackoverflow here : https://stackoverflow.com/questions/60232282/add-cors-policy-for-swagger-endpoints-only

wu-yafeng:
I create a pr for this.

before merged, the solution is creating a extensions method.
``` csharp

public static IEndpointConventionBuilder MapSwagger(
            this IEndpointRouteBuilder endpoints,
            string pattern = "/swagger/{documentName}/swagger.json",
            Action<SwaggerEndpointOptions> setupAction = null)
        {
            if (!RoutePatternFactory.Parse(pattern).Parameters.Any(x => x.Name == "documentName"))
            {
                throw new ArgumentException("Pattern must contain '{documentName}' parameter", nameof(pattern));
            }

            Action<SwaggerOptions> endpointSetupAction = options =>
            {
                var endpointOptions = new SwaggerEndpointOptions();

                setupAction?.Invoke(endpointOptions);

                options.RouteTemplate = pattern;
                options.SerializeAsV2 = endpointOptions.SerializeAsV2;
                options.PreSerializeFilters.AddRange(endpointOptions.PreSerializeFilters);
            };

            var pipeline = endpoints.CreateApplicationBuilder()
                .UseSwagger(endpointSetupAction)
                .Build();

            return endpoints.MapGet(pattern, pipeline);
        }
```

Update your `Startup.cs`

``` csharp
app.UseEnpoint(endpoint=>{
    endpoint.MapSwagger().RequireCors();
});
```

cyrildurand:
Thanks for your reply and pull request.

I finally took time to test your extension method and everything work.

[...]

PR #2101, opened 2021-04-22
Allow add endpoint metadata to Swagger Enpoint.

wu-yafeng:
Allowed add endpoint metadata to SwaggerEndpoint.

Such as:

``` csharp

app.UseEnpoint(endpoint=>{
    endpoint.MapSwagger().RequireCors().RequireAuthorize().RequireHost("host");
});

```

Address #2078

### DIFF
--- a/src/Swashbuckle.AspNetCore.Swagger/DependencyInjection/SwaggerBuilderExtensions.cs
+++ b/src/Swashbuckle.AspNetCore.Swagger/DependencyInjection/SwaggerBuilderExtensions.cs
@@ -40,7 +40,7 @@ namespace Microsoft.AspNetCore.Builder
         }
 
 #if (!NETSTANDARD2_0)
-        public static IEndpointRouteBuilder MapSwagger(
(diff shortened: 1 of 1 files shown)


domaindrivendev:
@wu-yafeng I'm not overly familiar with these interfaces but I believe this would be a breaking change for anyone who is using the `MapSwagger` extension already?

wu-yafeng:
> @wu-yafeng I'm not overly familiar with these interfaces but I believe this would be a breaking change for anyone who is using the `MapSwagger` extension already?

This is a breaking changes with minimal impact. 

``` csharp

// before
app.UseEndpoint(endpoint=>{
    endpoint.MapSwagger().MapController();
});

// after
app.UseEndpoint(endpoint=>{
    endpoint.MapSwagger();
    endpoint..MapController();
});

```

martincostello:
Thanks for contributing - if you'd like to continue with this pull request, please open a new PR that does not make breaking changes that otherwise still solves your use case.

PR #2102, opened 2021-04-22
Fix null value handling in OpenApiAnyFactory.

djjsdtc:
When fixing bool value parsing issue (#2092) I found that my changes to the code did not handle the null value in json correctly, and if there is null values, it will crash the whole parsing process. Also, my test case did not cover this situation.

Here is my additional fix for handling null values, with modified test cases covering this. Please feel free to let me know if you have any suggestions or additional changing requirements.


domaindrivendev:
I believe this is already fixed by #2151. If not, please create a new issue describing the problem complete with exact repro steps.

PR #2125, opened 2021-05-21
Fixes #2117

### LINKED ISSUE #2117: dotnet swagger tofile fails if any parameters contain spaces (even if quoted)
Version 5.6.3 -> 6.1.4

1) Create a .NET Core 3.1 project that correctly references and uses Swashbuckle.AspNetCore.
2) Install the Swashbuckle.AspNetCore.Cli tool in the project
3) Run the tool as `dotnet swagger tofile --serializeasv2 --output "Path With\Space To\swagger.json" "Path With\Space To\SwashbuckeUsingBinary.dll" v1`

RESULT:
Usage: dotnet swagger _tofile [options] [startupassembly] [swaggerdoc]

startupassembly:
  

swaggerdoc:
  

options:
  --output:  
  --host:  
  --basepath:  
  --serializeasv2:  
  --yaml:  

NOTE:
The options output doesn't have any helpful display text; the "tofile" command runner has helpful display text on each option, but the "_tofile" command runner does not.  "_tofile" is running because "tofile" calls "_tofile" (for runtime configuration reasons) by invoking a new process using "dotnet exec", but it seems that the parameters enclosed by meaningful double-quotes for "tofile" aren't re-enclosed with meaningful double-quotes for "_tofile", so "_tofile" sees that there are far too many arguments, runs "PrintUsage", and returns.

EXPECT:
Swagger to be generated at "Path With\Space To\Swagger.json".

phil-allen-msft:
Fixes #2117

The main command "tofile", if it has arguments that are enclosed in double quotes, will receive them as independent elements in the args array.  When the call to dotnet exec is made to invoke the subcommand "_tofile", if there are any arguments that contain spaces, they will now be regarded as multiple arguments and the subcommand run will have a different set of arguments, ones that do not match the users' intentions.

Uses the existing EscapePath to enclose those parameters that contain spaces with double quotes.  The result is that if the paths that are arguments for the "tofile" command (startupassembly or --output) contain spaces, they should now be seen by the subcommand the same as they were in the main command.

### DIFF
--- a/src/Swashbuckle.AspNetCore.Cli/Program.cs
+++ b/src/Swashbuckle.AspNetCore.Cli/Program.cs
@@ -42,14 +42,24 @@ namespace Swashbuckle.AspNetCore.Cli
 
                     var depsFile = namedArgs["startupassembly"].Replace(".dll", ".deps.json");
                     var runtimeConfig = namedArgs["startupassembly"].Replace(".dll", ".runtimeconfig.json");
+                    var commandName = args[0];
 
-                    var subProcess = Process.Start("dotnet", string.Format(
-                        "exec --depsfile {0} --runtimeconfig {1} {2} _{3}", // note the underscore
+                    var subProcessArguments = new string[args.Length - 1];
+                    if (subProcessArguments.Length > 0)
+                    {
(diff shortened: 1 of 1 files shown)


domaindrivendev:
Excellent thank you 👍

PR #2128, opened 2021-05-27
Use the URL API to create a location-relative path to "oauth2-redirect.html

### LINKED ISSUE #1854: Default oauth2RedirectUrl problem with query parameter

ulrichb:
Fixes #1854

URL JS API: https://caniuse.com/url (no IE 11, but there are already other non-IE 11 usages like [startsWith()](https://caniuse.com/mdn-javascript_builtins_string_startswith))

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerUI/index.html
+++ b/src/Swashbuckle.AspNetCore.SwaggerUI/index.html
@@ -88,7 +88,7 @@
 
             // If oauth2RedirectUrl isn't specified, use the built-in default
             if (!configObject.hasOwnProperty("oauth2RedirectUrl"))
-                configObject.oauth2RedirectUrl = window.location.href.replace("index.html", "oauth2-redirect.html").split('#')[0];
+                configObject.oauth2RedirectUrl = (new URL("oauth2-redirect.html", window.location.href)).href;
 
             // Apply mandatory parameters
             configObject.dom_id = "#swagger-ui";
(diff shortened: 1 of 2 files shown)


ulrichb:
@domaindrivendev Manually tested this with the `OAuth2Integration` sample app.

Any more tests?

ulrichb:
@domaindrivendev I didn't check, but this fix _may_ also fix #2076.

ulrichb:
I've just tested that `.../swagger/` (no index.html) also works. => So #2076 should indeed be fixed.

Tested it via temporarily modifying: https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/2fc1ac36dff97b6e84870b35ef331ca03c01de71/src/Swashbuckle.AspNetCore.SwaggerUI/SwaggerUIMiddleware.cs#L55)

PR #2143, opened 2021-06-22
Check UseAllOfToExtendReferenceSchemas in all Schema Generations

forteddyt:
The particular issue I ran into is when a Dictionary has a reference as a value. I'm using `UseAllOfToExtendReferenceSchemas()`.
ie for
```csharp
public Dictionary<string, InputDataBinding> InputDataBindings { get; set; }
```
I would expect
```json
"inputDataBindings": {
  "type": "object", 
  "additionalProperties": {
    "allOf": [
      {
        "$ref": "#/definitions/InputDataBinding"
      }
    ]
  }
```
but instead get
```json
"inputDataBindings": {
  "type": "object", 
  "additionalProperties": {
    "$ref": "#/definitions/InputDataBinding"
  }
```

The code changes I made address this specific scenario (and the ones tested), but I'm nowhere near familiar enough with the Swashbuckle codebase to know if this has unintended consequences.

Also I'd appreciate feedback on how I implemented the testing of this - do they fit with your view of Swashbuckle testing?


domaindrivendev:
Could you elaborate on the actual problem you're trying to solve here? Extending a type schema with the `allOf` construct is only neccessary when you have additional "contextual" metadata to be associated with the schema. For example, a property or parameter may be annotated with a description or other metadata. But, in the case of a dictionary value - all you have is the type specificed as the generic argument - there's no other "contextual" metadata and hence no need to "extend" the generated schema. Does that make sense?

forteddyt:
My team uses AutoRest as our client generator tool, and so our swagger specs are created to fall inline with the requirements outline by AutoRest. The specific scenario we'd like this feature for is decorating properties with [x-nullable](https://github.com/Azure/autorest/tree/main/docs/extensions#x-nullable) - note that the example given is with `"type": "string"`, but the same applies to `"$ref": "blah"`. To be fair, I'm not sure if there are other scenarios in AutoRest that would make use of this - perhaps with mutability metadata?

Our current work around is done via an ISchemaFilter:
```csharp
public void Apply(OpenApiSchema schema, SchemaFilterContext context)
{
    if (schema.AdditionalProperties?.Reference != null)
    {
        schema.AdditionalProperties.AllOf ??= new List<OpenApiSchema>();
        schema.AdditionalProperties.AllOf.Add(new OpenApiSchema { Reference = schema.AdditionalProperties.Reference });
        schema.AdditionalProperties.Reference = null;
    }
}
```

domaindrivendev:
TBH this part of the code is so complex (needs a major refactor) as it is and I’m concerned introducing a change like this could have unforeseen affects. Given that it’s for  a very specific use case with AutoRest, I’m inclined to not merge and just recommend the workaround.

domaindrivendev:
Thank you for the time, effort and insights

forteddyt:
sounds good - thanks for all you do with this project :)

Issue #2145, opened 2021-06-24
.NET 6 minimal APIs and the Swashbuckle CLI


bradygaster:
We were doing some experimenting building a new API using the .NET 6 minimal web template. So far, Swashbuckle is working as expected. However. we noticed one bug in which the CLI fails Here's some code for the new minimal api:

```csharp
using Microsoft.AspNetCore.Builder;
using Microsoft.AspNetCore.Hosting;
using Microsoft.Extensions.Configuration;
using Microsoft.Extensions.DependencyInjection;
using Microsoft.Extensions.Hosting;
using Microsoft.OpenApi.Models;
using Microsoft.AspNetCore.Mvc;
using Microsoft.Extensions.Caching.Memory;
using System.Collections.Generic;
using System.Linq;

const string OpenApiDocumentName = "v1";
const string MemcacheKey = "items";
const string Greeting = "Hello {0}!";

var builder = WebApplication.CreateBuilder(args);
builder.Services.AddMemoryCache();
builder.Services.AddEndpointsApiExplorer();
builder.Services.AddSwaggerGen(options => options.SwaggerDoc(OpenApiDocumentName, new OpenApiInfo
{
    Title = "Minimal Swagger Test",
    Description = "Simple description",
    Version = "v1"
}));

var app = builder.Build();

if (app.Environment.IsDevelopment())
{
    app.UseDeveloperExceptionPage();
}

app.UseSwagger(setup => setup.SerializeAsV2 = true);

app.Services.GetService<IMemoryCache>().Set<List<TheApiModel>>(MemcacheKey, new List<TheApiModel>());

app.MapGet("/", (IMemoryCache memcache) =>
    memcache.Get<List<TheApiModel>>(MemcacheKey)
);

app.MapGet("/{target}", (string target,IMemoryCache memcache) =>
    memcache.Get<List<TheApiModel>>(MemcacheKey).FirstOrDefault(_ => _.Target == target)
);

app.MapPost("/create", (TheApiModel model, IMemoryCache memcache) =>
{
    var existing = memcache.Get<List<TheApiModel>>(MemcacheKey);
    existing.Add(model);
    memcache.Set<List<TheApiModel>>(MemcacheKey, existing);
    return string.Format(Greeting, model.Target);
});

app.MapGet("/version", () => typeof(Microsoft.AspNetCore.Http.RequestDelegateFactory)
    .Assembly.GetCustomAttributes(true)
[...]

PR #2147, opened 2021-06-25
Added HostFactoryResolver to detect new application patterns

davidfowl:
- The HostFactoryResolver is [shared source](https://github.com/dotnet/runtime/blob/7ed5a67c0bbb48bdce3cfd962e333c8f7db047a4/src/libraries/Microsoft.Extensions.HostFactoryResolver/src/HostFactoryResolver.cs) that is used by tools to resolve the entry point and service provider from applications that use Microsoft.Extensions.Hosting. It has been updated to handle a couple of new patterns introduced in .NET 6 and this change uses it as a fallback after detecting the other patterns. I've copied the source into this project since I didn't want to add a new nuget feed as a requirement to build this application. We'll need to keep this file in sync if there are changes.
- I had to make changes to make the shared source support C# 7.3 (and remove nullable support). I'd prefer to fix it so that the project targeted a newer C# version but I didn't want to introduce new requirements for building the repository.
- Introduce HostingApplication as a way to abstract how the IServiceProvider is retrieved when there's no special entry point.

Fixes https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/2145

PS: I should probably add some tests, but I wasn't sure where to do so

### DIFF
--- a/src/Swashbuckle.AspNetCore.Cli/Program.cs
+++ b/src/Swashbuckle.AspNetCore.Cli/Program.cs
@@ -15,10 +15,18 @@ namespace Swashbuckle.AspNetCore.Cli
 {
     public class Program
     {
+        private CommandRunner _runner;
+
         static int Main(string[] args)
         {
+            var program = new Program(args);
+            return program.Run(args);
+        }
+
+        public Program(string[] args)
+        {
             // Helper to simplify command line parsing etc.
-            var runner = new CommandRunner("dotnet swagger", "Swashbuckle (Swagger) Command Line Tools", Console.Out);
+            _runner = new CommandRunner("dotnet swagger", "Swashbuckle (Swagger) Command Line Tools", Console.Out);
 
             // NOTE: The "dotnet swagger tofile" command does not serve the request directly. Instead, it invokes a corresponding
             // command (called _tofile) via "dotnet exec" so that the runtime configuration (*.runtimeconfig & *.deps.json) of the
@@ -26,7 +34,7 @@ namespace Swashbuckle.AspNetCore.Cli
             // startupassembly and it's transitive dependencies. See https://github.com/dotnet/coreclr/issues/13277 for more.
 
             // > dotnet swagger tofile ...
-            runner.SubCommand("tofile", "retrieves Swagger from a startup assembly, and writes to file ", c =>
+            _runner.SubCommand("tofile", "retrieves Swagger from a startup assembly, and writes to file ", c =>
             {
                 c.Argument("startupassembly", "relative path to the application's startup assembly");
                 c.Argument("swaggerdoc", "name of the swagger doc you want to retrieve, as configured in your startup class");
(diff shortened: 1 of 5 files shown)


src/Swashbuckle.AspNetCore.Cli/Program.cs
```diff
@@ -140,10 +140,24 @@ private static IServiceProvider GetServiceProvider(Assembly startupAssembly)
...
+            try
+            {
+                return WebHost.CreateDefaultBuilder()
+                   .UseStartup(startupAssembly.GetName().Name)
+                   .Build()
+                   .Services;
+            }
+            catch
+            {
+                var serviceProvider = HostingApplication.GetServiceProvider(startupAssembly);
```
domaindrivendev, line 152: @davidfowl you mentioned that the `HostFactoryResolver` is the de facto way to resolve entry point and service provider for apps that use `Microsoft.Extensions.Hosting`. With that in mind, shouldn't it be the default approach here rather than the fallback?

domaindrivendev:
@davidfowl re tests - up until this point the logic in this area was fairly straightforward and so I was able to live without much test coverage, particularly given the hoops you have to jump through to test this low-level infrastructure stuff. But, as this MR adds a significant amount of logic to the CLI, I'm not sure if that rationale makes sense anymore.

As the source is pretty much copied, it seems we should probably copy the tests as well. Would this be feasible? If so, you could possibly add them to the `Swashbuckle.AspNetCore.IntegrationTests` project. Interested to get your thoughts?

src/Swashbuckle.AspNetCore.Cli/Program.cs
davidfowl: I wanted to keep the existing compatibility.

davidfowl:
Yes I agree this needs tests now. I'll take a look.

bradygaster:
Let me know if you need me to test anything out with private builds, etc. I have a few sample projects you could use to test some baseline scenarios (Power, HttpRepl, etc.).

BrennanConroy:
Ideally we would want to have an end to end test where the tool runs against a built minimal 6.0 app. It doesn't look like that infrastructure exists in this repo currently and we would also need to take a dependency on 6.0 bits (the repo would, not the tool).

What would your recommendation for testing be @domaindrivendev ?

bradygaster:
@BrennanConroy let's start with a test that makes sure a standard .net core 5.0 web api will work once the changes you make to make it work with minimal go in. Then, couple that test with a test to make sure minimal works whenever future changes are made to update anything on the standard Web API side. 

if @domaindrivendev has any requests beyond continual validation that both the former Web API/MVC way works and the new way works, he can add addendum here or to the PR when you submit.

BrennanConroy:
Just pushed a couple proof of concept tests.

src/Swashbuckle.AspNetCore.Cli/Program.cs
```diff
@@ -15,18 +15,26 @@ namespace Swashbuckle.AspNetCore.Cli
 {
```
domaindrivendev, line 15: I'm curious to know what the rationale is for the changes to this file - seems like it's just a subjective preference to have the logic in instance methods vs static Main or am I missing something?

BrennanConroy: I was following the pattern we have in the aspnetcore repo for some tools, can change to make Main public and call that instead if it's preferred.

domaindrivendev: Yeh - if the only material benefit is consistency with an external repo, I'd prefer to revert to how it was. Thanks.

Issue #2165, opened 2021-07-17
DefaultOperationIdSelector should be updated to fallback to EndpointName if AttributeRouteInfo isn't populated


DamianEdwards:
The new minimal API feature in ASP.NET Core 6 doesn't populate the `ApiDescription.ActionDescriptor.AttributeRouteInfo` property which is [used by Swashbuckle by default to set the operation ID](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/95cb4d370e08e54eb04cf14e7e6388ca974a686e/src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGeneratorOptions.cs#L64). Minimal APIs can have an endpoint name set via endpoint metadata however so if Swashbuckle's default behavior were updated to fallback to using the `EndpointNameMetadata` in `ApiDescription.ActionDescriptor.EndpointMetadata` then it would be simple for operation IDs to be configured on minimal APIs without having to configure Swashbuckle manually.

e.g. change in [src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGeneratorOptions.cs](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/master/src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGeneratorOptions.cs)
``` c#
namespace Swashbuckle.AspNetCore.SwaggerGen
{
    public class SwaggerGeneratorOptions
    {
        ...
        private string DefaultOperationIdSelector(ApiDescription apiDescription)
        {
            return
                apiDescription.ActionDescriptor.AttributeRouteInfo?.Name
                ?? (apiDescription.ActionDescriptor.EndpointMetadata
                        .FirstOrDefault(m => m is EndpointNameMetadata) as EndpointNameMetadata)?.EndpointName;
        }
    }
}
```

Example minimal API:
``` c#
app.MapGet("/", () => "Hello World")
   .WithMetadata(new EndpointNameMetadata("HelloWorldApi"));
```

It might be worth considering including `RouteNameMetadata` as a value to look for too, rather than relying solely on `AttributeRouteInfo`.

domaindrivendev:
Released with `v6.1.5`.

PR #2173, opened 2021-07-29
Support setting operation IDs from EndpointName metadata

captainsafia:
Fixes https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/2165.

Part of https://github.com/dotnet/aspnetcore/issues/34514.

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGeneratorOptions.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGeneratorOptions.cs
@@ -71,7 +71,7 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
 #if (!NETSTANDARD2_0)
             return
                 actionDescriptor.AttributeRouteInfo?.Name
-                ?? (actionDescriptor.EndpointMetadata.FirstOrDefault(m => m is EndpointNameMetadata) as EndpointNameMetadata)?.EndpointName;
+                ?? (actionDescriptor.EndpointMetadata.FirstOrDefault(m => m is IEndpointNameMetadata) as IEndpointNameMetadata)?.EndpointName;
 #else
             return actionDescriptor.AttributeRouteInfo?.Name;
 #endif
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGeneratorOptions.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGeneratorOptions.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Linq;
 using Microsoft.AspNetCore.Mvc.ApiExplorer;
 using Microsoft.OpenApi.Models;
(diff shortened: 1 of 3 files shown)


src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGeneratorOptions.cs
```diff
@@ -61,7 +62,19 @@ private bool DefaultDocInclusionPredicate(string documentName, ApiDescription ap
...
         private string DefaultOperationIdSelector(ApiDescription apiDescription)
         {
-            return apiDescription.ActionDescriptor.AttributeRouteInfo?.Name;
+            var actionDescriptor = apiDescription.ActionDescriptor;
+
+            // Resolve the operation ID from the route name and fallback to the
+            // endpoint name if no route name is available. This allows us to
+            // generate operation IDs for endpoints that are defined using
+            // minimal APIs.
+#if (!NETSTANDARD2_0)
```
captainsafia, line 71: This is ifdef'd out since the M.A.Routing 2.1.0 package does not contain the `EndpointNameMetadata` types that we need here. We could update the repo to rely on M.A.* 2.2.0 packages for everything if we wanted to avoid if-defing here but that seemed more appropriate for a separate PR.

bradygaster:
quick ping to @domaindrivendev. @captainsafia can't apply labels or milestones but this is one of the minimal items we discussed. Thanks!

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGeneratorOptions.cs
```diff
@@ -61,7 +62,19 @@ private bool DefaultDocInclusionPredicate(string documentName, ApiDescription ap
...
+            return
+                actionDescriptor.AttributeRouteInfo?.Name
+                ?? (actionDescriptor.EndpointMetadata.FirstOrDefault(m => m is EndpointNameMetadata) as EndpointNameMetadata)?.EndpointName;
```
captainsafia, line 74: We do `FirstOrDefault` here and not `SingleOrDefault` because of https://github.com/dotnet/aspnetcore/issues/34199.

test/Swashbuckle.AspNetCore.SwaggerGen.Test/SwaggerGenerator/SwaggerGeneratorTests.cs
```diff
@@ -77,6 +80,34 @@ public void GetSwagger_SetsOperationIdToRouteName_IfActionHasRouteNameMetadata()
             Assert.Equal("SomeRouteName", document.Paths["/resource"].Operations[OperationType.Post].OperationId);
         }
 
+        [Fact]
+        public void GetSwagger_SetsOperationIdToEndpointName_IfActionHasEndpointNameMetadata()
+        {
+            var methodInfo = typeof(FakeController).GetMethod(nameof(FakeController.ActionWithParameter));
+            var actionDescriptor = new ControllerActionDescriptor
```
davidfowl, line 87: Test it without a controller action descriptor.

PR #2182, opened 2021-08-10
AllOf-Support for Interface-Types

angelaki:
Using this PR Swashbuckle is able to use AllOf for interface types.

I think the `_generatorOptions.SubTypesSelector` is not required, what do you think? It is related to this issue and I finally found the time to analyse it: https://github.com/domaindrivendev/Swashbuckle.WebApi/issues/1409. I'd love to see this getting merged since I'd like to stay with the head.

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/ConfigureSchemaGeneratorOptions.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/ConfigureSchemaGeneratorOptions.cs
@@ -35,6 +35,7 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
             target.SchemaIdSelector = source.SchemaIdSelector;
             target.IgnoreObsoleteProperties = source.IgnoreObsoleteProperties;
             target.UseAllOfForInheritance = source.UseAllOfForInheritance;
+            target.IgnoreAllOfSubTypesSelector = source.IgnoreAllOfSubTypesSelector;
             target.UseOneOfForPolymorphism = source.UseOneOfForPolymorphism;
             target.SubTypesSelector = source.SubTypesSelector;
             target.DiscriminatorNameSelector = source.DiscriminatorNameSelector;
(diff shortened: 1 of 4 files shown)


angelaki:
Not there is a clean implementation of `SubTypesSelector` (still needs to be checked what solution is preferred) with the possibility to disable it for `allOf`. Quite nice imho, let's go to head 👍

angelaki:
Still someone needs to check this failed check for me. lockfileVersion?

angelaki:
Anything I can do here for this moving on?

angelaki:
I'd really love to see this happen before it's not compatible any more. Would someone **please** merge this?

angelaki:
Ok ... So let's get this going? Is anything wrong with this? I'd like to get it merged before it loses compatibility ...

domaindrivendev:
To be honest, this area of the code is super brittle and needs to be significantly re-worked before I'd be willing to accept new contributions. Apologies as I'm sure this is frustrating to hear.

dtaalbers:
@domaindrivendev Any idea on when (if) this will be fixed? I think this will help to solve my current issue. 

This setup:
```
public interface IFooBar { }
```

```
public class Foo: IFooBar {}
public class Bar: IFooBar {}
```

```
public class Instructions 
{
   public List<IFooBar> Steps { get; set; }
}
```
with the following settings:

```
c.GeneratePolymorphicSchemas();
c.UseAllOfForInheritance();
c.SelectSubTypesUsing(baseType =>
{
    if (baseType == typeof(IFooBar))
    {
        return new[]
        {
            typeof(Foo),
            typeof(Bar)
        };
    }
    return Enumerable.Empty<Type>();
});

```
generates the following in `swagger.json`. I'd expected (and need) the `oneOf` to be `allOf` so that when I generate a typescript client base on this it will use both types instead of the first.
```
{
    "Instructions": {
        "properties": {
            "steps": {
                "nullable": true,
                "type": "array",
                "items": {
                    "oneOf": [
                        {
                            "$ref": "#/components/schemas/Foo"
                        },
                        {
                            "$ref": "#/components/schemas/Bar"
                        }
                    ]
                }
            }
        },
        "type": "object",
        "additionalProperties": false
    }
}
```

PR #2183, opened 2021-08-10
Change how IsRequired for parameter descriptions work

davidfowl:
- For non-controllers, change the logic to look for the IsRequired flag on the ApiParameterDescription instead of the fallback logic. This is to avoid potentially breaking existing behavior for controller based API descriptions. In theory, this logic could be changed for controllers since MVC does use that logic but that's a behavior change that would likely need to be opted into.

~TODO: Add tests.~

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/ApiParameterDescriptionExtensions.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/ApiParameterDescriptionExtensions.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
 using System.Linq;
 using System.Reflection;
 using Microsoft.AspNetCore.Http;
@@ -11,6 +12,33 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
 {
     public static class ApiParameterDescriptionExtensions
     {
+        private static readonly Type[] RequiredAttributeTypes = new[]
+        {
+            typeof(BindRequiredAttribute),
+            typeof(RequiredAttribute)
+        };
+
+        public static bool IsRequiredParameter(this ApiParameterDescription apiParameter)
+        {
+            // This is the default logic for IsRequired
+            bool IsRequired() => apiParameter.IsFromPath() ||
+                                 apiParameter.CustomAttributes().Any(attr => RequiredAttributeTypes.Contains(attr.GetType()));
+
+            // This is to keep compatibility with MVC controller logic that has existed in the past
+            if (apiParameter.ParameterDescriptor is ControllerParameterDescriptor)
+            {
+                return IsRequired();
+            }
+
+            // For non-controllers, prefer the IsRequired flag if we're not on netstandard 2.0, otherwise fallback to the default logic.
+            return
+#if !NETSTANDARD2_0
+            apiParameter.IsRequired;
+#else
+            IsRequired();
+#endif
+        }
+
         public static ParameterInfo ParameterInfo(this ApiParameterDescription apiParameter)
         {
             var controllerParameterDescriptor = apiParameter.ParameterDescriptor as ControllerParameterDescriptor;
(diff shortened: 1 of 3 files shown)


Eneuman:
I think this change will also fix todays problems with APIs using routes like this:
`[HttpGet("test/{layer?}")]`

In current release; "layer" would be marked as "Required" but white this change, it will be marked as not required by the ApiExplorer.

The OpenApi 3.0 Specification says "`Also remember to add required: true, because path parameters are always required`" 
(https://swagger.io/docs/specification/describing-parameters/) so I'm not sure if this will break specification or not.
But from what I can tell it should work :)

davidfowl:
> The OpenApi 3.0 Specification says "Also remember to add required: true, because path parameters are always required"
(https://swagger.io/docs/specification/describing-parameters/) so I'm not sure if this will break specification or not.
But from what I can tell it should work :)

Good catch! I didn't realize the spec said that. Not sure what the best "path" forward here is (pun intended). Maybe we leave this only for body?

davidfowl:
Changed to set requires if we're looking at a path parameter location.

> Determines whether this parameter is mandatory. If the parameter location is "path", this property is REQUIRED and its value MUST be true. Otherwise, the property MAY be included and its default value is false.

davidfowl:
@domaindrivendev can you take a look?

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/ApiParameterDescriptionExtensions.cs
```diff
@@ -11,6 +12,33 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
...
     public static class ApiParameterDescriptionExtensions
     {
+        private static readonly Type[] RequiredAttributeTypes = new[]
+        {
+            typeof(BindRequiredAttribute),
+            typeof(RequiredAttribute)
+        };
+
+        public static bool IsRequiredParameter(this ApiParameterDescription apiParameter)
+        {
```
domaindrivendev, line 22: As defaulting to `true` for path parameters is applicable to controller and non-controller cases, I would consider starting this method with the following one-liner, and removing that check from the rest of the logic. I think this makes the behavior a bit more explicit and simplifies the rest of the code:

```csharp
if (apiParameter.IsFromPath()) return true;
```

test/Swashbuckle.AspNetCore.SwaggerGen.Test/SwaggerGenerator/SwaggerGeneratorTests.cs
```diff
@@ -246,6 +246,75 @@ public void GetSwagger_SetsParameterRequired_IfApiParameterIsBoundToPath()
             Assert.Equal(expectedRequired, operation.Parameters.First().Required);
         }
 
+        [Theory]
+        [InlineData(false)]
+        [InlineData(true)]
+        public void GetSwagger_SetsParameterRequired_IfParameterIsPath(bool isRequired)
```
domaindrivendev, line 252: We already have a test for this case (see `GetSwagger_SetsParameterRequired_IfApiParameterIsBoundToPath`), which makes this one redundant.

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/ApiParameterDescriptionExtensions.cs
davidfowl: Agreed!

test/Swashbuckle.AspNetCore.SwaggerGen.Test/SwaggerGenerator/SwaggerGeneratorTests.cs
```diff
@@ -246,6 +246,75 @@ public void GetSwagger_SetsParameterRequired_IfApiParameterIsBoundToPath()
...
+            );
+
+            var document = subject.GetSwagger("v1");
+
+            Assert.True(document.Paths["/resource"].Operations[OperationType.Post].Parameters[0].Required);
+        }
+
+        [Theory]
+        [InlineData(false)]
+        [InlineData(true)]
```
domaindrivendev, line 282: As this is really testing the non-controller case, I think the test name should reflect that.

davidfowl: Done

test/Swashbuckle.AspNetCore.SwaggerGen.Test/SwaggerGenerator/SwaggerGeneratorTests.cs
davidfowl: Done

davidfowl:
Ping...

PR #2189, opened 2021-08-12
Add support for using <remarks> on schema XML comments

smbecker:
Currently, schema types always set the `Description` to match the `<summary>` XML comment. However, `OpenApiSchema` does have a `Title` field and in many other places, `Title=<summary>` and `Description=<remarks>`. This PR updates the schema filter to align with that. It does however, default to the current behavior if the `<remarks>` node is not present in the extracted comments.


daxu7509:
hi, can you force this to master branch? I found it quite useful, but would like to use a more official nuget package other than fork my own one

martincostello:

PR #2191, opened 2021-08-12
Add support for resolving conflicting actions using URL fragments

smbecker:
I have several APIs that use the same URL and method but disambiguate based on the request content-type. This causes conflicts in the generated swagger document. The current pattern using `ConflictingActionsResolver` only allows you to choose which operation to use rather than providing a mechanism to support both. A common pattern for resolving this is to use URL fragments to force the key to be unique. One approach would be to use the operation ID as the fragment. The approach that I took here is to just use an incrementing index on conflicting actions. I added tests to ensure that it behaves as expected but I am certainly open to other approaches to disambiguating the paths.


domaindrivendev:
@smbecker thanks for submitting this. I'd like to understand the problem a little better before discussing solutions and figure out if there's another way to describe these operations using [OpenAPI's native support for media types](https://swagger.io/docs/specification/media-types/) rather than tacking on meaningless path fragments.

Could you please create a corresponding issue that describes the problem with as much detail as possible, including code/JSON examples, so we can talk through potential options. Then, when we're agreed on an approach we can link a new PR to that issue.

Issue #2194, opened 2021-08-13
ArgumentNullException at ...SwaggerGeneratorOptions.DefaultOperationIdSelector


icnocop:
Hi.

Upgrading from 6.1.4 to 6.1.5 causes `System.ArgumentNullException: Value cannot be null. (Parameter 'source')` to be thrown when generating swagger.json.

I've created a repository which can reproduce the issue here:
https://github.com/icnocop/Swashbuckle.AspNetCore.InheritanceExample

```
Swashbuckle.AspNetCore.SwaggerGen.SwaggerGeneratorException: Failed to generate Operation for action - WebApplication9.Controllers.AnimalsController.Post (WebApplication9). See inner exception
 ---> System.ArgumentNullException: Value cannot be null. (Parameter 'source')
   at System.Linq.ThrowHelper.ThrowArgumentNullException(ExceptionArgument argument)
   at System.Linq.Enumerable.TryGetFirst[TSource](IEnumerable`1 source, Func`2 predicate, Boolean& found)
   at System.Linq.Enumerable.FirstOrDefault[TSource](IEnumerable`1 source, Func`2 predicate)
   at Swashbuckle.AspNetCore.SwaggerGen.SwaggerGeneratorOptions.DefaultOperationIdSelector(ApiDescription apiDescription)
   at Swashbuckle.AspNetCore.SwaggerGen.SwaggerGenerator.GenerateOperation(ApiDescription apiDescription, SchemaRepository schemaRepository)
   --- End of inner exception stack trace ---
   at Swashbuckle.AspNetCore.SwaggerGen.SwaggerGenerator.GenerateOperation(ApiDescription apiDescription, SchemaRepository schemaRepository)
   at Swashbuckle.AspNetCore.SwaggerGen.SwaggerGenerator.GenerateOperations(IEnumerable`1 apiDescriptions, SchemaRepository schemaRepository)
   at Swashbuckle.AspNetCore.SwaggerGen.SwaggerGenerator.GeneratePaths(IEnumerable`1 apiDescriptions, SchemaRepository schemaRepository)
   at Swashbuckle.AspNetCore.SwaggerGen.SwaggerGenerator.GetSwagger(String documentName, String host, String basePath)
   at Swashbuckle.AspNetCore.Swagger.SwaggerMiddleware.Invoke(HttpContext httpContext, ISwaggerProvider swaggerProvider)
[...]

PR #2195, opened 2021-08-13
Fixed regression when looking for an operation id

davidfowl:
- Handle null EndpointMetadata
- Added test

Fixes https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/2194


captainsafia (approved):
Oof! Thanks for fixing this and adding the tests.

PR #2209, opened 2021-08-31
Fix parsing of optional params in route

captainsafia:
The current `RelativePathSansQueryString` method causes problems with the approach that minimal APIs use for setting an endpoint names.

With an endpoint like:

```csharp
app.MapGet("/bar/{foo?}", (string foo) => ...);
```

Will produce a `relativePath` that looks something like `/bar/{foo` which is incorrect.

To mitigate this, I used a negative lookahead regular expression to only split on the query string part and to ignore optional parameters.

`Regex.Split` is not as performant as a `string.Split` but I wasn't sure what other alternatives might be best here.

Is there a particular reason that we remove the query string part? There might be another alternative to approach here (e.g. removing the query string with the existing strategy in another codepath).

### DIFF
--- a/test/Swashbuckle.AspNetCore.SwaggerGen.Test/SwaggerGenerator/SwaggerGeneratorTests.cs
+++ b/test/Swashbuckle.AspNetCore.SwaggerGen.Test/SwaggerGenerator/SwaggerGeneratorTests.cs
@@ -59,9 +59,6 @@ namespace Swashbuckle.AspNetCore.SwaggerGen.Test
 
                     ApiDescriptionFactory.Create<FakeController>(
                         c => nameof(c.ActionWithNoParameters), groupName: "v1", httpMethod: "GET", relativePath: "resource/{id?}"),
-
-                    ApiDescriptionFactory.Create<FakeController>(
-                        c => nameof(c.ActionWithNoParameters), groupName: "v1", httpMethod: "POST", relativePath: "resource/?param1"),
                     
                     ApiDescriptionFactory.Create<FakeController>(
                         c => nameof(c.ActionWithNoParameters), groupName: "v1", httpMethod: "POST", relativePath: "resource/foo"),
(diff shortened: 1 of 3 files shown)


domaindrivendev:
@captainsafia this extension method was carried over when I ported the original WebApi version to Core. It's used to obtain the corresponding OpenAPI "path" for an operation and removes the query part because [OpenAPI does not allow query strings in the "path"](https://swagger.io/docs/specification/paths-and-operations/):

> Query string parameters must not be included in paths. They should be defined as query parameters instead.

With that said though, I'm wondering if the extension is required anymore because I don't believe`ApiDescription.RelativePath` can ever have a query part in it in the first place. At least for controller/action routing, I don't _think_ you can define a route template with a query parameter in it (e.g. `[HttpGet("/resource?foo={bar}")]`. For minimal APIs though I'm not sure.

Perhaps you could confirm for both cases, and if it's not possible in either, I think we can just remove the extension method and revert to using `ApiDescription.RelativePath` instead.

captainsafia:
> At least for controller/action routing, I don't think you can define a route template with a query parameter in it (e.g. [HttpGet("/resource?foo={bar}")]. For minimal APIs though I'm not sure.

Query params aren't supported in route templates, only route segment parameters. It's expected that users will use parameters on the action or a `FromQuery` attribute.

Given this, I can update the PR to remove the use of the extension method.

test/Swashbuckle.AspNetCore.SwaggerGen.Test/SwaggerGenerator/SwaggerGeneratorTests.cs
```diff
@@ -48,6 +48,37 @@ public void GetSwagger_GeneratesSwaggerDocument_ForApiDescriptionsWithMatchingGr
             Assert.Equal(new[] { OperationType.Post, OperationType.Get }, document.Paths["/resource"].Operations.Keys);
         }
 
+        [Fact]
+        public void GetSwagger_GeneratesSwaggerDocument_ForApiDescriptionsWithVariousRelativePaths()
```
domaindrivendev, line 52: Given the change, I'm not sure this test has much value anymore as it's all just based on direct assignment of the `RelativePath` property surfaced by API Explorer. What do you think?

captainsafia: It's not really necessary but I left it in to future-proof. As far as I can tell, it's the only test that validates with optional route params so it gives us coverage for this area in the event of any issues in the future.

PR #2210, opened 2021-09-01
Add support for TagsMetadata in endpoints

captainsafia:
Part of https://github.com/dotnet/aspnetcore/issues/35310.


captainsafia:
@domaindrivendev So, this PR adds support for the net6.0 TFM in the repo. As part of this, we'll need to update the test environment to use the 6.0 SDK for validation (specifically RC1). Is this a change you're OK with? If so, do you know if it is easy to do with AppVeyor? I'm not super familiar with their CI setup but happy to dig in if you want to bump to 6.0 now.

The (non-ideal) alternative is to wait until 6.0 GAs.

domaindrivendev:
@captainsafia - understood and happy for you to modify appveyor config as needed. It's currently using a pre-built Windows image and I would have expected this to have the .NET 6 preview supported out-of-the-box but according to the [docs it only supports up to `5.0.400`](https://www.appveyor.com/docs/windows-images-software/#net-framework). So, for now, we may need to install .net 6 with each run by adding an `install:` section to `appveyor.yml`

```yml
install:
- ps: <Put the powershell command to install $env:DOTNET_VERSION here>

build_script:
- ...
```

I'm not sure how easy it is to install .NET 6 from the command line so this is something you would have to look into. Looks like you may need to pull down and invoke [the dotnet install script](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-install-script)

Also, it's worth noting this will slow down every build. If the build speed starts to cause serious issues, we could look into [custom build images](https://www.appveyor.com/docs/custom-build-images/) at a later point. But TBH I'd be surprised if appveyor don't add support for .NET 6 preiews to the canned images fairly soon.

domaindrivendev:
@captainsafia - on a separate note, when I try to build this branch locally using the `6.0.100-preview.7.21379.14` SDK, it fails with the following error:

> /Users/richie/src/Swashbuckle.AspNetCore/src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGeneratorOptions.cs(89,91): error CS0246: The type or namespace name 'ITagsMetadata' could not be found (are you missing a using directive or an assembly reference?) [/Users/richie/src/Swashbuckle.AspNetCore/src/Swashbuckle.AspNetCore.SwaggerGen/Swashbuckle.AspNetCore.SwaggerGen.csproj]

Which version of the SDK are you using to build locally?

captainsafia:
> Which version of the SDK are you using to build locally?

Ah yes, this change was included in RC1 so a nightly installer from https://github.com/dotnet/installer will need to be used. We can always wait until next week when the RC1 SDK is released publicly and do the update thing if you'd like to avoid using a nightly.

domaindrivendev:
@captainsafia looks good. I just noticed the test failure and think it’s due to the following:

https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/9abccd3069df4cddedc2b3e2eb7ce52cb0ed56b0/src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/JsonSerializerDataContractResolver.cs#L147

captainsafia:
> @captainsafia looks good. I just noticed the test failure and think it’s due to the following:

Ah thank you! I thought it was an issue with the new JSON generator. It looks like this is the only ifdef that needed to be updated.

captainsafia:
OK! I think this is in a good state now. If you're cooling taking a dependency on the RC2 nightlies in the build, feel free to merge. Otherwise, we can sit on this until the build is out.

domaindrivendev:
@captainsafia is it RC1 or RC2 that included this change? I tried building this branch locally with `6.0.100-rc.1` an am getting the same error.

captainsafia:
@domaindrivendev RC2 contains the change which won't be out of nightly for a bit so we can hold this change until then.

In the meantime, this change should work with any SDK version `6.0.100-rc.2.21458.9` or higher.

domaindrivendev:
Sounds good - ping me here as soon as RC2 is out and I'll merge the PR

PR #2214, opened 2021-09-03
Enhancement for SwaggerUIMiddleware

### LINKED ISSUE #2204: Performance issues on SwaggerUIMiddleware
From the `SwaggerUIMiddleware`,  

We're using `Regex.IsMatch` to catch whether it's the swagger path, wondering that if we could use compiled Regex for better performance

- <https://github.com/domaindrivendev/Swashbuckle.AspNetCore/tree/25501836a9254b13f4f5dc706fd1df9ee6cf2c3b/src/Swashbuckle.AspNetCore.SwaggerUI/SwaggerUIMiddleware.cs#L55>
- <https://github.com/domaindrivendev/Swashbuckle.AspNetCore/tree/25501836a9254b13f4f5dc706fd1df9ee6cf2c3b/src/Swashbuckle.AspNetCore.SwaggerUI/SwaggerUIMiddleware.cs#L66>

And when I get a dump from my app, it seemed a lot of tasks are blocked on `Regex.IsMatch` on `SwaggerUIMiddlewar`

image

And happen to see `var htmlBuilder = new StringBuilder(new StreamReader(stream).ReadToEnd());` wondering that maybe 
`reader.ReadToEndAsync()` is better?

WeihanLi:
fixes #2204 

- Use `String.Equals` instead of `Regex.Match`
- Use `StreamReader` `ReadToEndAsync` method instead of `ReadToEnd`
- Specific the `LangVersion`


src/Swashbuckle.AspNetCore.SwaggerUI/SwaggerUIMiddleware.cs
```diff
@@ -67,7 +68,7 @@ public async Task Invoke(HttpContext httpContext)
                 return;
             }
 
-            if (httpMethod == "GET" && Regex.IsMatch(path, $"^/{Regex.Escape(_options.RoutePrefix)}/?index.html$",  RegexOptions.IgnoreCase))
```
domaindrivendev, line 70: @WeihanLi the change to this line breaks the case where the `RoutePrefix` is empty, which is a legit config option.

We could add a conditional, which will further complicate the code, but now that I think about it - I'd be amazed if the previous implementation based on a simple Regex pattern was truly causing performance problems. Unless you have a better way to support the empty `RoutePrefix`, I'm tempted to revert this change (although I'd retain the async change below).

WeihanLi: Thanks @domaindrivendev , seemed missed a `Trim('/')`, would add a test case for this, thanks

WeihanLi: Hi @domaindrivendev , I've opened another PR to fix this https://github.com/domaindrivendev/Swashbuckle.AspNetCore/pull/2323

PR #2215, opened 2021-09-05
Consider EndpointMetadata for annotations

martincostello:
Also consider the `EndpointMetadata` for operations when targeting .NET 5 and later for Swagger annotation attributes. This allows attributes applied to ASP.NET Core 6 Minimal Actions to be discovered.

I ran into this issue while migrating an ASP.NET Core 5 demo project with various customisations for the Swagger documentation to use ASP.NET Core 6 Minimal Actions using a daily build of the .NET 6 RC1 SDK (https://github.com/martincostello/api/pull/576, specifically this commit https://github.com/martincostello/api/pull/576/commits/75b086c171bd1d494c76df5f34cffc7e92c8912d in the `ApiModule` class for the relevant parts, for example [this endpoint](https://github.com/martincostello/api/blob/75b086c171bd1d494c76df5f34cffc7e92c8912d/src/API/ApiModule.cs#L47-L65)).

~I wasn't sure specifically what to do regarding additional tests as all but one of the current test projects only currently target .NET Core 3.1, and #2210 looks like it adds various .NET 6 specific things to the project. Happy to take guidance on what to do about that and/or wait until that PR is merged.~

With this change all the existing tests pass, and I've observed no unexpected changes to the Swagger JSON document with this fix added locally into my project.

/cc @captainsafia

### DIFF
--- a/src/Swashbuckle.AspNetCore.Annotations/AnnotationsOperationFilter.cs
+++ b/src/Swashbuckle.AspNetCore.Annotations/AnnotationsOperationFilter.cs
@@ -20,7 +20,7 @@ namespace Swashbuckle.AspNetCore.Annotations
                 actionAttributes = context.MethodInfo.GetCustomAttributes(true);
             }
 
-#if NET5_0_OR_GREATER
+#if NET6_0_OR_GREATER
             if (context.ApiDescription.ActionDescriptor.EndpointMetadata != null)
             {
                 metadataAttributes = context.ApiDescription.ActionDescriptor.EndpointMetadata;
--- a/src/Swashbuckle.AspNetCore.Annotations/AnnotationsOperationFilter.cs
+++ b/src/Swashbuckle.AspNetCore.Annotations/AnnotationsOperationFilter.cs
@@ -10,18 +10,38 @@ namespace Swashbuckle.AspNetCore.Annotations
     {
         public void Apply(OpenApiOperation operation, OperationFilterContext context)
         {
-            if (context.MethodInfo == null) return;
+            IEnumerable<object> controllerAttributes = Array.Empty<object>();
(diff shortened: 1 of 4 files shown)


src/Swashbuckle.AspNetCore.Annotations/AnnotationsOperationFilter.cs
```diff
@@ -10,18 +10,38 @@ public class AnnotationsOperationFilter : IOperationFilter
...
+            if (context.MethodInfo != null)
+            {
+                controllerAttributes = context.MethodInfo.DeclaringType.GetCustomAttributes(true);
+                actionAttributes = context.MethodInfo.GetCustomAttributes(true);
+            }
 
-            // NOTE: When controller and action attributes are applicable, action attributes should take precendence.
-            // Hence why they're at the end of the list (i.e. last one wins)
-            var controllerAndActionAttributes = controllerAttributes.Union(actionAttributes);
+#if NET5_0_OR_GREATER
```
captainsafia, line 23:
```suggestion
#if NET6_0_OR_GREATER
```

Should this only target minimal apps in .NET 6 and above?

martincostello: Yeah, probably. I went with this now now as the 6 SDK wasn't available yet in main and this didn't seem to do any harm.

I can update it once your 6 SDK updates are merged it.

martincostello:
Updated the `#if` for .NET 6 and added what I _think_ is OK for a unit test for this.

jonaskello:
We ran in to this issue too now when trying to use .NET 6 minimal API. Any news on this?

nilzzzzzz:
Hi @domaindrivendev ,

any plans to make a nuget release or shall we fork the repo for now? 

Thank you!

Nils

PR #2219, opened 2021-09-08
For STJ - don't set readonly if set via constructor

domaindrivendev:
It looks like STJ now supports deserialization of readonly properties that are set via constructor


chrisbbe:
This also seems to affect serialization, how are we now supposed to keep `"isReadOnly": true` in the OpenAPI specification when public get only properties are set in constructor?

PR #2231, opened 2021-09-15
Remove constraints from route parameters in path

captainsafia:
Fixes https://github.com/dotnet/aspnetcore/issues/36413.

This fix expands on the changes made in https://github.com/domaindrivendev/Swashbuckle.AspNetCore/pull/2209.

Currently, the route constraints that are included in the route template path are included in the generated schema which breaks things like the `curl` command generated in the UI. This resolves that issue by stripping out constraints/qualifiers from the path.

Note that there is work tracked in https://github.com/dotnet/aspnetcore/issues/36525 to ensure that these route constraints show up in the schema.

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/ApiDescriptionExtensions.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/ApiDescriptionExtensions.cs
@@ -47,5 +47,26 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
 
             customAttributes = Enumerable.Empty<object>();
         }
+
+        internal static string RelativePathSansParameterConstraints(this ApiDescription apiDescription)
+        {
+            var routeTemplate = apiDescription.RelativePath;
+            
+            while (routeTemplate.IndexOfAny(new[] { ':', '=', '?' }) != -1)
+            {
+                var startIndex = routeTemplate.IndexOfAny(new[] { ':', '=', '?' }) ;
+                var tokenStart = startIndex + 1;
+                findEndBrace:
+                    var endIndex = routeTemplate.IndexOf('}', tokenStart);
+                    if (endIndex < routeTemplate.Length - 1 && routeTemplate[endIndex + 1] == '}')
+                    {
+                        tokenStart = endIndex + 2;
+                        goto findEndBrace;
+                    }
+                routeTemplate = routeTemplate.Remove(startIndex, endIndex - startIndex);
+            }
+            
+            return routeTemplate;
+        }
     }
 }
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/ApiDescriptionExtensions.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/ApiDescriptionExtensions.cs
@@ -52,10 +52,15 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
         {
             var routeTemplate = apiDescription.RelativePath;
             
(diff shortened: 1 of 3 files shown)


src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/ApiDescriptionExtensions.cs
```diff
@@ -47,5 +47,26 @@ public static IEnumerable<object> CustomAttributes(this ApiDescription apiDescri
 
             customAttributes = Enumerable.Empty<object>();
         }
+
+        internal static string RelativePathSansParameterConstraints(this ApiDescription apiDescription)
+        {
```
domaindrivendev, line 52: Could this be done with a Regex replace? I _feel_ that would be a little cleaner but not sure about perf implications.

test/Swashbuckle.AspNetCore.SwaggerGen.Test/SwaggerGenerator/SwaggerGeneratorTests.cs
```diff
@@ -48,20 +48,22 @@ public void GetSwagger_GeneratesSwaggerDocument_ForApiDescriptionsWithMatchingGr
...
-        public void GetSwagger_GeneratesSwaggerDocument_ForApiDescriptionsWithVariousRelativePaths()
+        [Theory]
+        [InlineData("resources/{id}", "/resources/{id}")]
+        [InlineData("{category}/{product?}/{sku}", "/{category}/{product}/{sku}")]
+        [InlineData("{area=Home}/{controller:required}/{id=0:int}", "/{area}/{controller}/{id}")]
+        [InlineData("{category}/product/{group?}", "/{category}/product/{group}")]
+        [InlineData("{category:int}/product/{group:range(10, 20)?}", "/{category}/product/{group}")]
+        [InlineData("{person:int}/{ssn:regex(^\\d{{3}}-\\d{{2}}-\\d{{4}}$)}", "/{person}/{ssn}")]
+        [InlineData("{person:int}/{ssn:regex(^(?=.*kind)(?=.*good).*$)}", "/{person}/{ssn}")]
+        public void GetSwagger_GeneratesSwaggerDocument_ForApiDescriptionsWithConstrainedRelativePaths(string routeTemplate, string processedRouteTemplate)
```
domaindrivendev, line 59: Very nitpicky, but in general when I'm using `InlineData`, I have a naming convention whereby I prefix parameter names with `expected` if they're being used in an assertion. So in this case, I would propose renaming `processedRouteTemplate` to something like `expectedPath`.

captainsafia: Gotcha! Can do.

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/ApiDescriptionExtensions.cs
captainsafia: Perhaps! Admittedly, I'm by now means a regex expert but things get pretty gnarly when you try to capture all of the requirements for the matcher in regex form (matching '?', '=', or ':', handling regex constraints, etc).

In generally, we've avoid parsing routes with regex over in ASP.NET Core. We'll usually roll out token enumerator implementation ([ref](https://github.com/dotnet/aspnetcore/blob/b1560573192f70bfbfa7bb39be51aaafc0c91190/src/Framework/AspNetCoreAnalyzers/src/Analyzers/DelegateEndpoints/DetectMismatchedParameterOptionality.cs#L66)) to handle this.

I do agree that some code comments might make this a little bit easier to read/maintain.

PR #2235, opened 2021-09-21
Add method to configure Bearer authentication

### LINKED ISSUE #2234: Add extension method for Bearer authentication
This is how I add [Bearer authentication](https://swagger.io/docs/specification/authentication/bearer-authentication/):
```cs
services.AddSwaggerGen(c =>
{
    c.AddSecurityDefinition("bearerAuth", new OpenApiSecurityScheme
    {
        Type = SecuritySchemeType.Http,
        Scheme = "bearer",
        BearerFormat = "JWT",
        Description = "JWT Authorization header using the Bearer scheme."
    });
    c.AddSecurityRequirement(new OpenApiSecurityRequirement
    {
        {
            new OpenApiSecurityScheme
            {
                Reference = new OpenApiReference { Type = ReferenceType.SecurityScheme, Id = "bearerAuth" }
            },
            new string[] {}
        }
    });
});
```

That is way too much boilerplate code. Please provide a extension method so I can set it up with less code, I would like to do something like this:
```cs
services.AddSwaggerGen(c =>
{
    c.AddBearerAuthentication();
});
```

vanillajonathan:
A method to quickly and easily add Bearer authentication without lots of boilerplate code.

Closes #2234

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenOptionsExtensions.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenOptionsExtensions.cs
@@ -24,6 +24,32 @@ namespace Microsoft.Extensions.DependencyInjection
         }
 
         /// <summary>
+        /// Add Bearer authentication.
+        /// </summary>
+        /// <param name="bearerFormat">The Bearer token format. Defaults to "JWT".</param>
+        /// <param name="swaggerGenOptions"></param>
+        public static void AddBearerAuthentication(this SwaggerGenOptions swaggerGenOptions, string bearerFormat = "JWT")
+        {
+            swaggerGenOptions.AddSecurityDefinition("bearerAuth", new OpenApiSecurityScheme
+            {
+                Type = SecuritySchemeType.Http,
+                Scheme = "bearer",
+                BearerFormat = bearerFormat,
+                Description = "Authorization header using the Bearer scheme."
+            });
+            swaggerGenOptions.AddSecurityRequirement(new OpenApiSecurityRequirement
+            {
+                {
+                    new OpenApiSecurityScheme
+                    {
+                        Reference = new OpenApiReference { Type = ReferenceType.SecurityScheme, Id = "bearerAuth" }
+                    },
+                    new string[] {}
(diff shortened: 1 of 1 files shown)


src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenOptionsExtensions.cs
```diff
@@ -23,6 +23,32 @@ public static class SwaggerGenOptionsExtensions
...
+        /// Add Bearer authentication.
+        /// </summary>
+        /// <param name="bearerFormat">The Bearer token format. Defaults to "JWT".</param>
+        /// <param name="swaggerGenOptions"></param>
+        public static void AddBearerAuthentication(this SwaggerGenOptions swaggerGenOptions, string bearerFormat = "JWT")
+        {
+            swaggerGenOptions.AddSecurityDefinition("bearerAuth", new OpenApiSecurityScheme
+            {
+                Type = SecuritySchemeType.Http,
+                Scheme = "bearer",
```
WeihanLi, line 36:
```suggestion
                Scheme = "Bearer",
```

domaindrivendev:
I'm sure this won't be a popular answer but I'd prefer not go down the road of introducing "convenience" functions to simplify writing Swagger/Open API as that's not a core concern of Swashbuckle. First and foremost, Swashbuckle's goal is to automatically generate Swagger/Open API documents based on your APIs _actual_ implementation - "code first", "code is truth" and all that. And, in the cases where this auto-generated description is not sufficient, the secondary goal is to expose a Swagger/Open API object model so that the document can be tweaked as needed.

Unfortunately, the Swagger/Open API's semantics for defining security schemes is heavy on the boilerplate, so I fully understand the value in this PR. I just don't want Swashbuckle to take on the third concern of "making it easier to write Swagger/Open API". If you'd like, you could create a separate add-on project, which I'd be happy to reference from the readme, that adds this "convenience" method and others?

vanillajonathan:
Yeah, that is not a popular answer with me. I have to copy paste that boiler plate code in all my projects, it would be great with an extension method to simplify that, and it would also be of help to new users who else would have to figure it all out.

PR #2236, opened 2021-09-21
Ensure UseSwaggerUI & AddSwaggerGen defaults match

DamianEdwards:
Updated the defaults for `IApplicationBuilder.UseSwaggerUI` to better match the [defaults of `IServiceCollection.AddSwaggerGen`](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/master/src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/ConfigureSwaggerGeneratorOptions.cs#L51) so that the displayed document name in the Swagger UI matches the actual document name and title.

We intend to change the default webapi project template code in ASP.NET Core in .NET 6.0 to simply use the Swashbuckle defaults and this change makes the UX align more clearly.


domaindrivendev:
@DamianEdwards @bradygaster - I merged this quickly as it looked very straightforward but didn't realize that it's actually breaing the build and now master after being merged. Can you please take a look?

domaindrivendev:
Never mind - was a quick fix - cbb3912e

bradygaster:
Thanks @domaindrivendev!

PR #2238, opened 2021-09-23
Set `ApplicationName` to startup assembly in swagger CLI

captainsafia:


global.json
```diff
@@ -0,0 +1,5 @@
+{
+  "sdk": {
+    "version": "6.0.100-rc.1.21463.6"
```
davidfowl, line 3: This is a tricky one. Do you really want anyone contributing to the entire repo to require rc1?

captainsafia: At some point, this is going to be required for changes like https://github.com/domaindrivendev/Swashbuckle.AspNetCore/pull/2210. @domaindrivendev mentioned being OK with upgrading to a publicly available RC is fine. I think since we say our RCs are go-live, this should be fine.

The alternative is to remove the tests for this but I think we're gonna benefit greatly from having a test for top-level apps in the repo.

src/Swashbuckle.AspNetCore.Cli/HostingApplication.cs
```diff
@@ -65,7 +65,11 @@ void OnEntryPointExit(Exception exception)
             try
             {
                 // Get the IServiceProvider from the host
-                var services = ((IHost)factory(Array.Empty<string>())).Services;
+                var assemblyName = assembly.GetName()?.Name ?? string.Empty;
```
davidfowl, line 68: I think we want the full name here.

captainsafia: Better safe than sorry!

captainsafia:
@domaindrivendev Can you take a look at this when you get the chance? It resolves an issue with the Docker CLI and minimal apps that use endpoints and controllers.

domaindrivendev:
@captainsafia this appears to also include a lot of the changes from #2210 so it's hard to know what change is specific to this PR. Can you sumarize the code change Docker CLI issue so I can take a look at that and ignore the rest?

captainsafia:
> UPDATE: I see now that this depends on RC1 whereas #2210 depends on RC2. So presumably this could be merged now whereas the other still needs to wait for the official RC2 release. Is that correct?

Yes, correct. I realize this is not clear from looking at the code. This update uses the publicly available RC1 release.

> @captainsafia this appears to also include a lot of the changes from #2210 so it's hard to know what change is specific to this PR. Can you sumarize the code change Docker CLI issue so I can take a look at that and ignore the rest?

The relevant code changes are in this commit: https://github.com/domaindrivendev/Swashbuckle.AspNetCore/pull/2238/commits/b21d0771cf484dfca088ac01dbab0678988c9e74

daniatic:
@captainsafia @domaindrivendev this change leads to a bug when an application uses the new MapStaticAssets. When MapStaticAssets is called without the optional staticAssetsManifestPath parameter the implementation tries to create one by using the ApplicationName. When run by the swashbuckle.cli the ApplicationName is set to the full assembly name which is not a resolvable path and in turn throws an exception when it tries to resolve the path for the manifest file. Please check the excerpt below for the implementation of MapStaticAssets:
```
    public static StaticAssetsEndpointConventionBuilder MapStaticAssets(this IEndpointRouteBuilder endpoints, string? staticAssetsManifestPath = null)
    {
        ArgumentNullException.ThrowIfNull(endpoints);

        var environment = endpoints.ServiceProvider.GetRequiredService<IWebHostEnvironment>();
        staticAssetsManifestPath ??= $"{environment.ApplicationName}.staticwebassets.endpoints.json";

        staticAssetsManifestPath = !Path.IsPathRooted(staticAssetsManifestPath) ?
            Path.Combine(AppContext.BaseDirectory, staticAssetsManifestPath) :
            staticAssetsManifestPath;

        var result = MapStaticAssetsCore(endpoints, staticAssetsManifestPath);

        if (StaticAssetDevelopmentRuntimeHandler.IsEnabled(result.IsBuildManifest, endpoints.ServiceProvider))
        {
            StaticAssetDevelopmentRuntimeHandler.EnableSupport(endpoints, result, environment, result.Descriptors);
        }

        return result;
    }
```

I found a simple workaround by providing the staticAssetsManifestPath myself, but I think this could be improved by setting it to `assembly.GetName().Name` instead of `assembly.GetName().FullName`?

PR #2242, opened 2021-10-04
Feat: Allow to use xml localization with `xml:lang` attribute

unchase:
Allows to use localized xml-tags (`summary`, `remarks`, `example`, `param`) in swagger documentation with [`xml:lang`](https://docs.microsoft.com/en-us/dotNet/api/system.xml.xpath.xpathnavigator.xmllang?view=net-5.0) attribute like this:

```csharp
/// <summary>
/// WeatherForecastController
/// </summary>
/// <summary xml:lang="ru">
/// Контроллер погоды (на русском)
/// </summary>
[ApiController]
[Route("[controller]")]
public class WeatherForecastController : ControllerBase
{
    ...
}
```

```csharp
/// <summary>
/// WeatherForecast (eng)
/// </summary>
/// <summary xml:lang="ru">
/// WeatherForecast (на русском)
/// </summary>
/// <remarks>
/// WeatherForecast description (eng)
/// </remarks>
/// <remarks xml:lang="ru">
/// Описание WeatherForecast (на русском)
/// </remarks>
public class WeatherForecast
{
	...

	/// <summary>
	/// Summary (eng)
	/// </summary>
	/// <summary xml:lang="ru">
	/// Summary (на русском)
	/// </summary>
	/// <example>Some summary (eng)</example>
	/// <example xml:lang="ru">Некоторый summary (на русском)</example>
}
```

Just use `culture` parameter in `IncludeXmlComments`:

```csharp
{
	...

	services.AddSwaggerGen(c =>
	{
		c.SwaggerDoc("v1", new OpenApiInfo { Title = "Test API", Version = "1" });

		var xmlCommentsPath = Path.Combine(System.AppContext.BaseDirectory, "GenericControllers.xml");
		c.IncludeXmlComments(xmlCommentsPath, culture: new CultureInfo("ru"));

                ...
	});
}
```

image


unchase:
@domaindrivendev Can you take a look at this when you get time?

domaindrivendev:
There's a lot of new code/changes with this PR, to an area of the code that's already become quite complex and difficult to maintain. Furthermore, if you're looking to provide localized versions of the docs, I don't believe adding translated text for every supported language, inline with your code is a best practice for internationalizing an application. Ideally, the translations should be decoupled from your code so that translators can transalte and coders can code without stepping on eachothers toes. So, just having a separate XML comments file for each language feels like a better solution, and that's already supported as you can pass in a different file to `IncludeXmlComments` in the same way you're providing a different `CultureInfo` instances in your example above.

PR #2244, opened 2021-10-04
Enable custom schema sorting

forteddyt:
Typically, this would be done by DocumentFilters, but since the schemas are sorted after filters are applied, it's impossible to do this. Commit 224a0843bf5faa180ebd9a93d05e35f53c76a5ee moved the schema sorting to after filter application.

This PR adds a new SchemaComparer SwaggerGeneratorOptions property that will handle last-mile schema sorting.

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenOptionsExtensions.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenOptionsExtensions.cs
@@ -106,6 +106,18 @@ namespace Microsoft.Extensions.DependencyInjection
         }
 
         /// <summary>
+        /// Provide a custom comprarer to sort schemas with
+        /// </summary>
+        /// <param name="swaggerGenOptions"></param>
+        /// <param name="schemaComparer"></param>
+        public static void SortSchemasWith(
+            this SwaggerGenOptions swaggerGenOptions,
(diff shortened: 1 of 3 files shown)


domaindrivendev:
Could you please add a test for this - something similar to https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/master/test/Swashbuckle.AspNetCore.SwaggerGen.Test/SwaggerGenerator/SwaggerGeneratorTests.cs#L678

Also - the swagger generator config is typically exposed via convenience/extension methods on `SwaggerGenOptions`. For consistency, we should add a corresponding method for this config. Again, something similar to https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/master/src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenOptionsExtensions.cs#L101.

forteddyt:
I've added the convenience/extension method, but I'm having trouble figuring out how to add a test. Specifically, I don't know how to get the SwaggerGenerator.GetSwagger to populate its [swaggerDoc.Components.Schemas](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/master/src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs#L49). Any pointers on that?

dlstucki:
I think this feature is incomplete:
https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/2741

PR #2252, opened 2021-10-18
Check if custom mapping has base class

### LINKED ISSUE #2250: MapType should check for inheritance
The `MapType` extension method should check for inheritance of the type. This way you could reduce configuration.

Instead of:
```cs
options.MapType<Cat>(() => new OpenApiSchema { Type = "string", Format = "uuid" });
options.MapType<Dog>(() => new OpenApiSchema { Type = "string", Format = "uuid" });
// ...
options.MapType<Zebra>(() => new OpenApiSchema { Type = "string", Format = "uuid" });
```

the configuration could be just:
```cs
options.MapType<Animal>(() => new OpenApiSchema { Type = "string", Format = "uuid" });
```

If the classes `Cat`, `Dog`, ..., `Zebra` all inherited from the class `Animal`.

vanillajonathan:
This adds a check to see if the type has a base class which is inserted as custom type to map to.
This allows to map a base class and apply the mapping  to all the classes that inherit from that base class.

Closes #2250


domaindrivendev:
See https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/2250#issuecomment-1014812122

vanillajonathan:
😭️

PR #2259, opened 2021-10-25
Add support for generating schema from route constraints

captainsafia:


domaindrivendev:
@captainsafia - historically, I've tried to keep the Schema generator decoupled from ApiExplorer types and AspNetCore in general, because at that point they felt like separate concerns. So, when first seeing this PR I wasn't mad about adding `ApiParameterRouteInfo` to the interface signature.

But, over time it's become clear that the schema's can (and should) be dictated by many framework features (incl. route constraints) not just POC type definitions and attributes. With AspNetCore evolving, and my own increased understanding of Swashbuckle's problem space, it would be ideal to re-write many of it's underlying abstractions. But, who has time for that right 😉?

So, in lieue of that, I think this PR makese perfect sense and I'd be good to merge. It's currently in draft - are you good for me to merge?

captainsafia:
> So, in lieue of that, I think this PR makese perfect sense and I'd be good to merge. It's currently in draft - are you good for me to merge?

For this change to work end-to-end, users will need to be on a 7.0-preview1 build to consume the changes introduced in https://github.com/dotnet/aspnetcore/pull/37470.

The changes here are backwards-compatible with the previous version where the ApiExplorer didn't properly populate the `RouteInfo` property.

So, yes it's good to merge if you're comfortable that this is will only work on .NET 7 and above.

> But, over time it's become clear that the schema's can (and should) be dictated by many framework features (incl. route constraints) not just POC type definitions and attributes. With AspNetCore evolving, and my own increased understanding of Swashbuckle's problem space, it would be ideal to re-write many of it's underlying abstractions.

We're in a similar space at the moment with exploring some of the gaps in the API Explorer implementation. At the moment, it's look like we'll be relying heavily on EndpointMetadata to annotate a lot of things with our type implementations, but we'll see how things unfold come implementation time. 😅

Issue #2243, opened 2021-10-04
[BUG] Wrong parameter with odata functions


mferraricloudsurfers:
Using Swashbuckle.AspNetCore.Annotations **v6.2.2** odata function parameters are truncated like this example:

![image](https://user-images.githubusercontent.com/38034428/135851850-f9c9053d-2f2a-4c31-bbe2-c00ba06348f3.png)

with version 6.2.1 works correctly

![image](https://user-images.githubusercontent.com/38034428/135852010-8e564e7f-ef79-4daf-a22f-6a064975f5d3.png)

domaindrivendev:
Closing in favor of #2310

PR #2263, opened 2021-10-29
fix: constraints from route parameters in path

camillo-toselli:
Fix for [BUG] Wrong parameter with odata functions #2243


domaindrivendev:
Closing in favor of #2310

Issue #1726, opened 2020-06-27
Add MapSwaggerUI extension method for IEndpointRouteBuilder


edumserrano:
**Request**

Similar to the MapSwagger extension method in SwaggerBuilderExtensions https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/b7cf75e7905050305b115dd96640ddd6e74c7ac9/src/Swashbuckle.AspNetCore.Swagger/DependencyInjection/SwaggerBuilderExtensions.cs#L41-L42

Can we have a MapSwaggerUI added to SwaggerUIBuilderExtensions that works on IEndpointRouteBuilder ?https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/b7cf75e7905050305b115dd96640ddd6e74c7ac9/src/Swashbuckle.AspNetCore.SwaggerUI/SwaggerUIBuilderExtensions.cs#L10

The scenarios enable by this are described in #1705. Also note the comments in #1705 about returning IEndpointConventionBuilder instead of IEndpointRouteBuilder.

ylabade:
Any news about this issue ?

adurrans:
Is there any update for this issue?

PR #2266, opened 2021-11-04
add MapSwaggerUI for Endpoint routing

stefannikolei:
This solves #1726

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerUI/SwaggerUIBuilderExtensions.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerUI/SwaggerUIBuilderExtensions.cs
@@ -1,11 +1,16 @@
 ﻿using System;
-using Microsoft.AspNetCore.Hosting;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Options;
 using Swashbuckle.AspNetCore.SwaggerUI;
 
 #if NETSTANDARD2_0
 using IWebHostEnvironment = Microsoft.AspNetCore.Hosting.IHostingEnvironment;
+#endif
+
+#if (!NETSTANDARD2_0)
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.Routing;
 #endif
 
 namespace Microsoft.AspNetCore.Builder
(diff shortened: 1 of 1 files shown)


src/Swashbuckle.AspNetCore.SwaggerUI/SwaggerUIBuilderExtensions.cs
```diff
@@ -38,10 +43,38 @@ public static IApplicationBuilder UseSwaggerUI(this IApplicationBuilder app, Swa
...
                 var hostingEnv = app.ApplicationServices.GetRequiredService<IWebHostEnvironment>();
-                options.ConfigObject.Urls = new[] { new UrlDescriptor { Name = $"{hostingEnv.ApplicationName} v1", Url = "v1/swagger.json" } };
+                options.ConfigObject.Urls = new[]
+                    { new UrlDescriptor { Name = $"{hostingEnv.ApplicationName} v1", Url = "v1/swagger.json" } };
             }
 
             return app.UseSwaggerUI(options);
         }
+
+
```
domaindrivendev, line 53: The implementation here looks a little different to how it's been done for the [SwaggerMiddleware](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/master/src/Swashbuckle.AspNetCore.Swagger/DependencyInjection/SwaggerBuilderExtensions.cs#L42), which I know has gone through a couple of iterations. Could this be implemented in the same way or is there a good reason for them to be different?

stefannikolei: When the convention builder is returned, you are then able to further use it and add Authentication to it. Or what do you mean

domaindrivendev: Returning the builder makes sense. I’m referring to the rest of the method impl. Just do a diff to see what I mean

stefannikolei: Now I see what you mean. I will adjust it. The difference is that we try to get the settings from DI when the action is null

stefannikolei: Ok I got back to it and had a look at it. 

We need this implementation here. The problem is that otherwise the static files will produce an 401 when you try to get them. (This is prevented by the .Use(....)

If you like I can remove the else branch for getting the settings from DI. --> setupAction must then be provided for this method.

martincostello:
Thanks for contributing - if you'd like to continue with this pull request, please rebase against the default branch to pick up our new CI. Please also add tests.

PR #2287, opened 2021-12-01
Replace <c></c> with Markdown backticks instead of curly brackets

Quppa:
`<c></c>` tags are currently [replaced](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/master/src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsTextHelper.cs#L11) by [`{}`](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/master/src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsTextHelper.cs#L95), but this is [rather old code](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/commit/85d4d09707679b5a04c7a6b3770820d24b786127#diff-233d46370aa00ef71a1979e91ebef89acdf75f0d7280e89d424eb207eeff6002R9), and I couldn't find anything about this being standard practice in OpenAPI/Swagger (the only mention I could find of `{}` was in relation to [URL path parameters](https://swagger.io/docs/specification/paths-and-operations/)). Redoc displays curly brackets as-is.

This PR changes the behaviour to use GitHub-flavoured markdown, instead - `<c></c>` tags are replaced with single backticks and `<code></code>` tags (previously ignored) are replaced with triple backticks. Redoc supports this style of code formatting out of the box.


tanujsati:
thanks

PR #2306, opened 2022-01-08
Add the property Example to SwaggerSchemaAttribute

fsandrade:
This change add the feature of use an property example using SwaggerSchemaAttribute as we already have the XML option.

### DIFF
--- a/src/Swashbuckle.AspNetCore.Annotations/AnnotationsSchemaFilter.cs
+++ b/src/Swashbuckle.AspNetCore.Annotations/AnnotationsSchemaFilter.cs
@@ -1,10 +1,10 @@
-﻿using System;
+﻿using Microsoft.Extensions.DependencyInjection;
+using Microsoft.OpenApi.Models;
+using Swashbuckle.AspNetCore.SwaggerGen;
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
-using Microsoft.Extensions.DependencyInjection;
-using Microsoft.OpenApi.Models;
-using Swashbuckle.AspNetCore.SwaggerGen;
 
 namespace Swashbuckle.AspNetCore.Annotations
 {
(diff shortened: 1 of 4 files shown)


src/Swashbuckle.AspNetCore.Annotations/AnnotationsSchemaFilter.cs
```diff
@@ -96,6 +96,12 @@ private void ApplySchemaAttribute(OpenApiSchema schema, SwaggerSchemaAttribute s
 
             if (schemaAttribute.Title != null)
                 schema.Title = schemaAttribute.Title;
+
+            if (schemaAttribute.Example != null)
+                schema.Example = OpenApiAnyFactory.CreateFromJson(
+                    schemaAttribute.Example.All(char.IsDigit)
```
domaindrivendev, line 102: Rather than second guessing based on the contents of the provided example (i.e `All(char.IsDigit)`), I think it's safer (and consistent with the implementation for XML comments) to wrap the example in quotes based on the schema type:

```csharp
var exampleAsJson = (schema.ResolveType(context.SchemaRepository) == "string") && !exampleNode.Value.Equals("null")
    ? $"\"{exampleNode.ToString()}\""
    : exampleNode.ToString();
```

NOTE: `context` is available on the calling method so can be passed down here.

domaindrivendev: Other than that all looks good 👍

fsandrade: Hi, @domaindrivendev 
Thanks for your answer.
In my tests the context.SchemaRepository is comming null, so I can't evaluate the type from the schema.
Am I doing something wrong?

martincostello:

PR #2310, opened 2022-01-11
Filter Out Qualifiers Only between Curly Braces

### LINKED ISSUE #2239: Error on Swagger for relative paths with colon character
I have an API endpoint (.net core 3.1), with the following signature:

```
[HttpGet("{divisionName}/things:count", Name = "GetNumberOfThings")]
public async Task<IActionResult> GetNumberOfThings([FromRoute] [Required] string divisionName, [FromQuery] [Required] string groupBy)
```

When I try to go to the Swagger page, I get the following error:
image

The endpoint is working as expected, it is just when I try to access the Swagger.
Previously the Swagger was working, it was only when I updated Swashbuckle.AspNetCore to version 6.2.2.

I think it has to do with that new internal method 'RelativePathSansParameterConstraints', which is looking for a '}' after the ':'.

Can you please have a look?
 
Many thanks in advance.

### LINKED ISSUE #2245: System.ArgumentOutOfRangeException: Count cannot be less than zero. (Parameter 'count')
I have a route with double points (e.g. /api/v1/bill:confirm)
after updating from 6.2.1 to 6.2.2 the swagger.json file cannot be created anymore.
I get:
An unhandled exception has occurred while executing the request.
System.ArgumentOutOfRangeException: Count cannot be less than zero. (Parameter 'count')
   at System.String.Remove(Int32 startIndex, Int32 count)
   at Swashbuckle.AspNetCore.SwaggerGen.ApiDescriptionExtensions.RelativePathSansParameterConstraints(ApiDescription apiDescription)
   at Swashbuckle.AspNetCore.SwaggerGen.SwaggerGenerator.<>c.<GeneratePaths>b__6_0(ApiDescription apiDesc)
   at System.Linq.Lookup`2.Create(IEnumerable`1 source, Func`2 keySelector, IEqualityComparer`1 comparer)
   at System.Linq.GroupedEnumerable`2.GetEnumerator()
   at Swashbuckle.AspNetCore.SwaggerGen.SwaggerGenerator.GeneratePaths(IEnumerable`1 apiDescriptions, SchemaRepository schemaRepository)
   at Swashbuckle.AspNetCore.SwaggerGen.SwaggerGenerator.GetSwagger(String documentName, String host, String basePath)
   at Swashbuckle.AspNetCore.Swagger.SwaggerMiddleware.Invoke(HttpContext httpContext, ISwaggerProvider swaggerProvider)
   at Microsoft.AspNetCore.Diagnostics.DeveloperExceptionPageMiddleware.Invoke(HttpContext context)

### LINKED ISSUE #2247: Route's parameters divided by a colon are not shown in UI
My .NET Core controller's action route consists of parameters separated by a colon:
[**Route**("{param1}/{param2}**:**{param3}")]
public void Action(string param3, string param2, string param3)  { ... }

When I open the swagger.json in run-time, the _param3_ of that action is hidden in the shown URL and only '}' character is visible at the end:
[POST] {param1}/{param2}}

This issue is similar to #2239 a bit, but I don`t have a run-time exception and I have bindable parameters while there is just static text in the route.

Used Swashbuckle.AspNetCore version: 6.2.2

### LINKED ISSUE #2294: Composite keys result into broken links in swagger.json
Environment
.Net 6
Microsoft.AspNetCore.OData 8.0.4
Swashbuckle.AspNetCore  6.2.3

When an entity has a composite key defined, swagger.json generate paths as "/odata/Products(Key},OtherKey})"

It looks that the problem is in ApiDescriptionExtensions.RelativePathSansParameterConstraints
I would expect ideally the path would be "/odata/Products(Key={Key},OtherKey={OtherKey})"

Model:
```
public class ProductModel
    {
        public int Key { get; set; }
        public int OtherKey { get; set; }
    }
```
Startup.cs:
```
public class Startup
    {
        {  
            services.AddControllers()
            .AddOData(opt =>

            { 
                opt.Count().AddRouteComponents("odata", GetEdmModel());              
            }
            
            );

            services.AddSwaggerGen(c =>
            {
                c.SwaggerDoc("v1", new OpenApiInfo { Title = "odata8", Version = "v1" });
            });
        }

        public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
        {
            if (env.IsDevelopment())
            {
                app.UseDeveloperExceptionPage();
            }

            app.UseSwagger();
            app.UseSwaggerUI(c => c.SwaggerEndpoint("/swagger/v1/swagger.json", "odata8 v1"));

            app.UseRouting();

           app.UseEndpoints(endpoints =>
            {
                endpoints.MapControllers();
            });

        }

        private static IEdmModel GetEdmModel()
        {
            ODataConventionModelBuilder builder = new ODataConventionModelBuilder();

            builder.EntitySet<ProductModel>("Products");
            builder.EntityType<ProductModel>().HasKey(r => new { r.Key, r.OtherKey });

           return builder.GetEdmModel();
        }
    }
```

Controller:
```
public class ProductsController : ODataController
    {
        /// <summary>
        /// Gets a collection of products
       /// </summary>
        /// <returns>Collection of <see cref="ProductModel"/></returns>
        /// <response code="200">Gets all rproducts</response>
		[Produces("application/json")]
        [ProducesResponseType(typeof(ProductModel[]), StatusCodes.Status200OK)]
        [EnableQuery]
        public IQueryable<ProductModel> Get()
        {
            return new List<ProductModel> {
            new ProductModel{Key = 1 },
            new ProductModel{Key = 2 },
            new ProductModel{Key = 3 },
            }.AsQueryable();
        }

        [EnableQuery]

       public IActionResult Get(int keyKey, int keyOtherKey)
        {
            return Ok(new ProductModel { Key = keyKey, OtherKey = keyOtherKey });
        }

   }
```

chrisoverzero:
The current implementation can, for example, corrupt paths which
legitimately contain equals signs which are not indicating a default
value. This can come up in OData or when simulating matrix parameters.

Other contributors have taken a swing at this problem, but I'm hoping that
the small size of this solution will help its acceptance.

Fixes #2239
Fixes #2247
Fixes #2245
Fixes #2294

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/ApiDescriptionExtensions.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/ApiDescriptionExtensions.cs
@@ -2,16 +2,14 @@
 using System.Linq;
 using System.Collections.Generic;
 using System.Reflection;
-using System.Text;
 using Microsoft.AspNetCore.Mvc.ApiExplorer;
 using Microsoft.AspNetCore.Mvc.Controllers;
+using Microsoft.AspNetCore.Routing.Template;
 
 namespace Swashbuckle.AspNetCore.SwaggerGen
 {
     public static class ApiDescriptionExtensions
     {
-        private static readonly char[] Disallowed = { ':', '=', '?' };
-
         public static bool TryGetMethodInfo(this ApiDescription apiDescription, out MethodInfo methodInfo)
         {
             if (apiDescription.ActionDescriptor is ControllerActionDescriptor controllerActionDescriptor)
@@ -53,56 +51,11 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
 
         internal static string RelativePathSansParameterConstraints(this ApiDescription apiDescription)
         {
-            var routeTemplate = apiDescription.RelativePath;
-
-            // We want to filter out qualifiers that indicate a constraint (":")
-            // a default value ("=") or an optional parameter ("?")
-            if (routeTemplate.IndexOfAny(Disallowed) == -1)
-            {
-                // Quick-exit if there's nothing to remove.
-                return routeTemplate;
-            }
-
-            var modifiedTemplate = new StringBuilder(routeTemplate.Length);
-            var index = 0;
-            var isBetweenCurlyBraces = false;
-            while (index < routeTemplate.Length)
-            {
-                var current = routeTemplate[index];
-
-                if (current == '{')
-                {
-                    isBetweenCurlyBraces = true;
-                    goto next;
-                }
-
-                if (current == '}')
-                {
-                    isBetweenCurlyBraces = false;
-                    goto next;
-                }
-
-                if (isBetweenCurlyBraces && Disallowed.Contains(current))
-                {
-                    // Only find a single instance of a '}' after our start
-                    // character to avoid capturing escaped curly braces
-                    // in a regular expression constraint
-                findEndBrace:
-                    index = routeTemplate.IndexOf('}', index + 1);
-                    if (index < routeTemplate.Length - 1 && routeTemplate[index + 1] == '}')
-                    {
-                        index += 2;
-                        goto findEndBrace;
-                    }
-                    continue;
-                }
-
-            next:
-                modifiedTemplate = modifiedTemplate.Append(current);
-                index += 1;
-            }
-
-            return modifiedTemplate.ToString();
+            var routeTemplate = TemplateParser.Parse(apiDescription.RelativePath);
+            var sanitizedSegments = routeTemplate
+                .Segments
+                .Select(s => string.Concat(s.Parts.Select(p => p.Name != null ? $"{{{p.Name}}}" : p.Text)));
+            return string.Join("/", sanitizedSegments);
         }
     }
 }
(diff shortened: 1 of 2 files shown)


src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/ApiDescriptionExtensions.cs
```diff
@@ -51,28 +54,55 @@ public static IEnumerable<object> CustomAttributes(this ApiDescription apiDescri
...
+                modifiedTemplate = modifiedTemplate.Append(current);
+                index += 1;
             }
-            
-            return routeTemplate;
+
+            return modifiedTemplate.ToString();
         }
     }
 }
```
domaindrivendev, line 108: I feel this code is becoming increasingly complicated and brittle and wonder if we could simplify by leveraging the same route template parser that's used by ASP.NET Core routing (see https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.routing.template.templateparser?view=aspnetcore-6.0)? For example, something along the following lines:

```csharp
internal static string RelativePathSansParameterConstraints(this ApiDescription apiDescription)
{
    var routeTemplate = TemplateParser.Parse(apiDescription.RelativePath);

    var builder = new StringBuilder();
    foreach (var segment in routeTemplate.Segments)
    {
        foreach (var part in segment.Parts)
        {
            builder.Append(part.Name != null ? $"{{{part.Name}}}" : part.Text);
        }
        builder.Append("/");
    }
    
    return builder.ToString().TrimEnd('/');
}
```

chrisoverzero: Sure thing. (I admit that I was minimally solving my immediate problem.) Should I work something like that sample into this MR, or would you prefer to handle it yourself?

domaindrivendev: If you could work it into this PR that would be great 👍

PR #2326, opened 2022-01-30
Add default username for OAuth2 password flow.

pixellos:
I wanted to set default username for OAuth2 password flow.
Unfortunately, its not exposed.

Added possibility to do so.


pixellos:
@domaindrivendev When we can expect new nuget release?

PR #2338, opened 2022-02-09
Added possibility to register schema factory instances with more access to the type internals

### LINKED ISSUE #2333: Access DataContract and CLR Type in Open Generic Mapping
In https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/1599 and https://github.com/domaindrivendev/Swashbuckle.AspNetCore/pull/1600 there was support for open generics added. 

Unfortunately we can only register a simple `Func<OpenApiSchema>` and therefore we cannot access any details of the actual type being mapped. This is problematic if you have complex models where the generic type argument is used internally to encode some aspects. 

Some usecases: 
- A `Delta<TEntity>` object which has all properties of TEntity but as not required
- A `QueryResult<TEntity>` object which serializes as some custom object holding an array of `TEntity` inside the object. 

I have custom `JsonConverter`s for these classes because their serialization involves some complex logic. 

It would be great if we can have an overload like this: 
```cs
public interface ISchemaFactory
{
  OpenApiSchema CreateSchema(Type type, DataContract dataContract);
}
//  SwaggerGenOptionsExtensions
public static void MapType<TMappedType, TSchemaFactory>(this SwaggerGenOptions swaggerGenOptions) where TSchemaFactory : ISchemaFactory
{
}
```

Where `type` is the concrete type for which a schema is generated allowing access to the type arguments and `dataContract` is the related data contract which the `SchemaGenerator` acquired. The registration of the factory as type instead of instances should allow usage of injected service in the schema factory. 

That's the simple proposal, a more advanced version could try to use the Microsoft DI infrastructure directly. Instead of registering the factories on the `SwaggerGenOptions` we would register them in the DI container as transient services. 

My current workaround is to use SchemaFilters but this is not nice because a lot of types are potentially added to the overall schema which I have to clean again.

Danielku15:
Fixes #2333

I changed the approach to be more in-line on how other features like the filters. I introduced a new interface which can be registered and also a context which allows access to auxiliary services. The proposal in #2333 would be maybe a bit a larger initiative to bring more DI compatibility into Swashbuckle. 

This minimal solution would solve my intermediate needs quite well. 

I tried to stick to the current established pracices (conventions, level of documentation and testing) but can adapt the code if needed.

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenOptionsExtensions.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenOptionsExtensions.cs
@@ -184,6 +184,34 @@ namespace Microsoft.Extensions.DependencyInjection
         /// <summary>
         /// Provide a custom mapping, for a given type, to the Swagger-flavored JSONSchema
         /// </summary>
+        /// <param name="swaggerGenOptions"></param>
+        /// <param name="type">System type</param>
+        /// <param name="schemaFactory">A factory method that generates Schema's for the provided type</param>
+        public static void MapType(
+            this SwaggerGenOptions swaggerGenOptions,
+            Type type,
+            Func<SchemaFactoryContext, OpenApiSchema> schemaFactory)
+        {
+            swaggerGenOptions.SchemaGeneratorOptions.SchemaFactories.Add(type, new DelegatedSchemaFactory(schemaFactory));
+        }
+
+        /// <summary>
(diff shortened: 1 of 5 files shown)


domaindrivendev:
Hey thx for this contribution but I'd prefer not introduce yet another abstraction in this area of the code - it's complex enough as it is. For the specific usecase you're trying to solve you can create and register a `SchemaGenerator` "decorator" as I've described in the [related issue](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/2333#issuecomment-1035695675)

PR #2339, opened 2022-02-09
Allow rollForward to next .net major version in global.json

tbolon:
The current global.json file is locked to a specific (and outdated) SDK version.

[global.json offers the possibility](https://docs.microsoft.com/en-us/dotnet/core/tools/global-json?tabs=netcore3x#rollforward) to specify a SDK version but allow using a more recent version with the same major version number.

This PR fixes that.


global.json
```diff
@@ -1,5 +1,6 @@
 {
     "sdk": {
-      "version": "6.0.100-rc.2.21505.57"
+      "version": "6.0.100-rc.2.21505.57",
```
martincostello, line 3: RC2 isn't supported anymore, so might as well change this to one of the stable .NET 6 SDKs.

tbolon: Yes, I didn't want to change anything else, so I let it as-is.

PR #2384, opened 2022-03-23
Support new `UseInlineDefinitionsForObjects` flag

sol-wasserman:
# Problem

When a request/response has a nested object, there are two pieces of information missing from the generated doc file:

1. The property's `summary` is not displayed.
2. The `Nullable` marker is not shown for `nullable` properties.

---

## Here's a screenshot of what it looks like now:

image

## Here's a screenshot of the source code for the above:

image

As you can see, this property has a nice `summary`, and is `nullable`.

## Here's a screenshot of what it I would expect:

image

# Underlying Cause

It looks to me like the reason it's not generating that properly is because the object is only a reference, and Swagger doesn't support these on references. As [the Swagger specification](https://swagger.io/specification/#reference-object) says, regarding `$ref`s:

> This object cannot be extended with additional properties and any properties added SHALL be ignored.

# Proposed Solution

We have [previously](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/commit/9abdbf99a271de5decb2bd93c5c614632e750b32) introduced an option to `UseInlineDefinitionsForEnums`. Among other things, this will show the summary of a property that is an enum, and will also show the `Nullable` marker accordingly.

This PR adds a similar `UseInlineDefinitionsForObjects` option. When enabled, it'll inline the whole schema, and will properly show the `summary` and whether it's `nullable`:

image

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerUI/package-lock.json
+++ b/src/Swashbuckle.AspNetCore.SwaggerUI/package-lock.json
@@ -9,20 +9,20 @@
       "version": "1.0.0",
       "hasInstallScript": true,
       "dependencies": {
-        "swagger-ui-dist": "4.12.0"
+        "swagger-ui-dist": "4.14.0"
       }
     },
     "node_modules/swagger-ui-dist": {
-      "version": "4.12.0",
-      "resolved": "https://registry.npmjs.org/swagger-ui-dist/-/swagger-ui-dist-4.12.0.tgz",
-      "integrity": "sha512-B0Iy2ueXtbByE6OOyHTi3lFQkpPi/L7kFOKFeKTr44za7dJIELa9kzaca6GkndCgpK1QTjArnoXG+aUy0XQp1w=="
+      "version": "4.14.0",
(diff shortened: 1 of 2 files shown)


sol-wasserman:
@domaindrivendev Just checking in whether you got a chance to look at this.

Is there a better approach to achieving this? Or is there anything else I can improve in this PR?

Thanks in advance.

xinge-shizhiqiang:
我也遇到了这个问题 这个问题解决了吗

xinge-shizhiqiang:
升级到最新的6.5.0版本后 使用了UseAllOfToExtendReferenceSchemas()而不是UseInlineDefinitionsForObjects()问题解决了 谢谢老兄！



------------------&nbsp;原始邮件&nbsp;------------------
发件人: "Sol ***@***.***&gt;; 
发送时间: 2022年12月21日(星期三) 凌晨0:42
收件人: ***@***.***&gt;; 
抄送: ***@***.***&gt;; ***@***.***&gt;; 
主题: Re: [domaindrivendev/Swashbuckle.AspNetCore] Support new `UseInlineDefinitionsForObjects` flag (PR #2384)





 
This issue was accidentally closed. I've reopened it here: #2533.

PR #2416, opened 2022-05-17
Improve IsRequiredSpecified performance

Turnerj:
The `IsRequiredSpecified` method isn't _slow_ by itself but with how frequently it can be called, it kinda still is. For example in the code I'm working on, the total time within `IsRequiredSpecified` is 3 seconds:

image

When you look at the memory allocations too, it gets kinda crazy:

image

The problem here is that calling `GetCustomAttributesData` is expensive. How expensive depends on the member itself but here is the results for a basic benchmark:

image

Now, while the original code specifically checked if the user specified a `Required` value at all, this new code only checks that it isn't the default value (the default value is that it is not required). The original code also has a mistake here as it treats setting the `Required` value to anything means it is required which isn't true (as you _can_ set it to `Required.Default` which is to _not_ require something).

Technically `DisallowNull`, according to the Newtonsoft docs says "The property is not required..." but I'm not really wanting to modify that type of behaviour. It is a lot more clear cut that `Required.Default` means it _isn't_ required.

### DIFF
--- a/src/Swashbuckle.AspNetCore.Newtonsoft/SchemaGenerator/JsonPropertyExtensions.cs
+++ b/src/Swashbuckle.AspNetCore.Newtonsoft/SchemaGenerator/JsonPropertyExtensions.cs
@@ -23,10 +23,8 @@ namespace Swashbuckle.AspNetCore.Newtonsoft
             if (memberInfo.GetCustomAttribute<JsonRequiredAttribute>() != null)
                 return true;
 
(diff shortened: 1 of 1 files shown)


Turnerj:
I guess with the subtle behaviour change, the method name doesn't really line up 100% anymore however that really only matters if someone wanted `[JsonProperty(Required = Required.Default)]` to actually make the property required (which it shouldn't).

martincostello:

Issue #2374, opened 2022-03-14
Enabling nullabillity for project


Eneuman:
@domaindrivendev Would you be intrested in accepting a PR for enabling nullability checks in compile time?
I could enabling <Nullable>enable</Nullable> for all the projects and fix any nullabillity issues discovered?

I ran into a issue where I missed to check for nullabillity in the Nullable Reference type PR, and theese kinds if misses can be avoided by enabling nullable checkes at compile time :)

liesahead:
Any updates?

martincostello:
We'd be open to a PR to accept this, but if it's a big change it might have to wait for a release or two to be merged as we'd likely want to get bug fixes out without causing breaking changes at the same time.

martincostello:
https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/4df3d53e574fb988bc05727ac242fbe85754cfb5/Directory.Build.props#L22-L25

martincostello:
#3659 tried to add some limited nullability, but the public API analyzers make it difficult to do hyper-targeted adoption.

See https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/3707#issuecomment-3642700769.

PR #2418, opened 2022-05-18
Observe the route template constraints in the Swagger middleware

0xced:
The default route template, i.e. `swagger/{documentName}/swagger.{json|yaml}` which is used by the `SwaggerMiddleware` is problematic because it matches any file extension. Even though it *looks like* only `json` and `yaml` extensions are supported, actually **any** extension matches. Trying to hit the following endpoints all return the JSON swagger document:

* `swagger/v1/swagger.xml`
* `swagger/v1/swagger.yml`
* `swagger/v1/swagger.anything`

This is not a very big deal, until the `SwaggerUIMiddleware` is also used and one chooses to modify the default route to `swagger/{documentName}.{json|yaml}`.

This is the problematic configuration:

```csharp
var builder = WebApplication.CreateBuilder(args);
builder.Services.AddMvcCore().AddApiExplorer();
builder.Services.AddSwaggerGen(c => c.SwaggerDoc("v1", new OpenApiInfo { Title = "Test API", Version = "1" }));
var app = builder.Build();

app.UseSwagger(c => c.RouteTemplate = "swagger/{documentName}.{json|yaml}");
app.UseSwaggerUI(c => c.SwaggerEndpoint("v1.json", "Test API"));

app.Run();
```

At this point, the `SwaggerMiddleware` will try to serve `swagger/index.html` because the route template matches (`documentName` = `index` and `json|yaml` = `html`) but the `index` document doesn't exist and this results in a 404 instead of calling the next (SwaggerUI) middleware.

To fix this issue, the default route template has been modified to `swagger/{documentName}/swagger.{extension:regex(^(json|ya?ml)$)}`, leveraging ASP.NET Core [route constraints](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/routing?view=aspnetcore-6.0#route-constraints) and the constraints are actually enforced in the `SwaggerMiddleware` implementation.

The default route template has also been modified in the `MapSwagger` method to ensure that only `json`, `yaml` and `yml` extensions are supported by default.


martincostello:
Thanks for contributing - if you'd like to continue with this pull request, please rebase against the default branch to pick up our new CI. Are any more specific asserts needed in the integration test you've added?

0xced:
Great to see that `Swashbuckle.AspNetCore` is not dead! I rebased my branch onto `master`, ready for the GitHub workflows to run.

I don't think more asserts are warranted as the new integration test fully covers the problem described. (It would not pass without the changes in the route constraint handling.)

src/Swashbuckle.AspNetCore.Swagger/SwaggerMiddleware.cs
```diff
@@ -4,6 +4,9 @@
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Routing;
+#if !NETSTANDARD
```
martincostello, line 7: I think it would be a bit more readable to use something like `#if NET5_OR_GREATER` (or whatever the appropriate one is) for these, rather than _"not .NET Standard"_. Otherwise I had to go to check the project file to see if these also applied to any `net4x` TFMs (which it turns out we don't target).

src/Swashbuckle.AspNetCore.Swagger/SwaggerMiddleware.cs
```diff
@@ -57,7 +70,7 @@ public async Task Invoke(HttpContext httpContext, ISwaggerProvider swaggerProvid
                     filter(swagger, httpContext.Request);
                 }
 
-                if (Path.GetExtension(httpContext.Request.Path.Value) == ".yaml")
+                if (extension is ".yaml" or ".yml")
```
martincostello, line 73: Do we need to care about case-sensitivity here?

src/Swashbuckle.AspNetCore.Swagger/SwaggerMiddleware.cs
```diff
@@ -15,19 +18,29 @@ public class SwaggerMiddleware
...
+#if !NETSTANDARD
+            ,TemplateBinderFactory templateBinderFactory
+#endif
+            )
         {
             _next = next;
             _options = options ?? new SwaggerOptions();
             _requestMatcher = new TemplateMatcher(TemplateParser.Parse(_options.RouteTemplate), new RouteValueDictionary());
+#if !NETSTANDARD
+            _templateBinder = templateBinderFactory.Create(RoutePatternFactory.Parse(_options.RouteTemplate));
```
martincostello, line 37: Is there any danger that someone's existing custom `RouteTemplate` would start throwing an exception when passed through this code?

src/Swashbuckle.AspNetCore.Swagger/SwaggerMiddleware.cs
0xced: I think it's correct to use `#if !NETSTANDARD` as it mirrors the condition on the `$(TargetFramework)` in the csproj files. Also, there are many other places where `#if (!NETSTANDARD2_0)` is used.

Anyway, once #2798 is implemented this will become moot as all conditional compilation can be removed.

src/Swashbuckle.AspNetCore.Swagger/SwaggerMiddleware.cs
0xced: It would throw the line before in `TemplateParser.Parse(_options.RouteTemplate)` which uses `RoutePatternFactory.Parse` under the hood. Anyway, you'd want an invalid route template to throw.

src/Swashbuckle.AspNetCore.Swagger/SwaggerMiddleware.cs
martincostello: Happy to leave it as is, though I think the other one is more readable.

Regarding #2798, that might be a while yet.

src/Swashbuckle.AspNetCore.Swagger/SwaggerMiddleware.cs
```diff
@@ -15,19 +18,29 @@ public class SwaggerMiddleware
...
+#if !NETSTANDARD
+        private readonly TemplateBinder _templateBinder;
+#endif
 
         public SwaggerMiddleware(
             RequestDelegate next,
-            SwaggerOptions options)
+            SwaggerOptions options
+#if !NETSTANDARD
+            ,TemplateBinderFactory templateBinderFactory
```
martincostello, line 29: I just noticed this is a binary-breaking change. Unless we park this until v5, this should instead add a new constructor overload if there's a way to _somehow_ get this to work without the `TemplateBinderFactory`.

0xced: I added a new constructor in 2fd5bcf0233579574954324aec268c20131fd0cb and restored the existing one in order to avoid a binary breaking change.

I think it should be considered to make the `SwaggerMiddleware` class internal for the next major version as I'm pretty sure everybody uses either the `UseSwagger` or `MapSwagger` extension method and nobody instantiates a `SwaggerMiddleware` instance directly.

martincostello: I've opened #2805 to track this suggestion.

src/Swashbuckle.AspNetCore.Swagger/SwaggerMiddleware.cs
```diff
@@ -72,16 +88,37 @@ public async Task Invoke(HttpContext httpContext, ISwaggerProvider swaggerProvid
...
             documentName = null;
+            extension = null;
             if (request.Method != "GET") return false;
 
             var routeValues = new RouteValueDictionary();
-            if (!_requestMatcher.TryMatch(request.Path, routeValues) || !routeValues.ContainsKey("documentName")) return false;
+            if (_requestMatcher.TryMatch(request.Path, routeValues))
+            {
+#if !NETSTANDARD
+                if (!_templateBinder.TryProcessConstraints(request.HttpContext, routeValues, out _, out _))
```
martincostello, line 101: We should null guard this, as it's technically possible for someone using the TFM with this support to use the constructor that doesn't provide a value, and then it'll `NullReferenceException`.

0xced: Yes you're right! I just addressed this in f30392a4f97d694a2e53fd53bb462d29202a8a7d.

I don't know how to code without nullable reference types anymore. 😅 (This would have caught this error by the way.)

martincostello: We have an issue to track that it turns out: #2374

martincostello (approved):

PR #2427, opened 2022-05-27
#2426 - Added schema generator option for inline definitions for objects

### LINKED ISSUE #2426: [Feature] Provide the capability to use inline schemas for object types
Simple types are by default serialized inline in the Swagger JSON files output by the `SchemaGenerator` class. 

After [posting on StackOverflow](https://stackoverflow.com/questions/72387736/is-there-a-way-to-expand-references-in-swashbuckle-to-provide-inline-schemas/72397727#72397727) to see if there was a configuration option available to enable object inline schemas, and reading through the documentation for Swashbuckle, I came up short. 

A new config option in the generator options to `UseInlineDefinitions` would allow this small change to be a part of the wider Swashbuckle API layer than having to create a custom `SchemaGenerator` type to get around this.

This feature would enhance the capabilities of importing the generated Swagger JSON into Microsoft's Power Automate, for example, which has a limit on the number of reference types it will accept for a single connector.

jamesmcroft:
Fixes #2426 

Similar to the implementation for inline definitions for Enum values, this change allows object definitions to be configured for inline definitions rather than by reference.


sol-wasserman:
Duplicate of https://github.com/domaindrivendev/Swashbuckle.AspNetCore/pull/2384

PR #2429, opened 2022-05-27
#2428 set minLength for RequiredAttribute.AllowEmptyStrings=false

### LINKED ISSUE #2428: DataAnnotations.Required attribute AllowEmptyStrings=false should set minLength: 1
Hi, 
When using DataAnnotations.Required attribute, AllowEmptyStrings property of the attribute isn't taken into account. 
The[ AllowEmptyString is set to false by default](https://docs.microsoft.com/en-us/dotnet/api/system.componentmodel.dataannotations.requiredattribute.allowemptystrings?view=net-6.0#system-componentmodel-dataannotations-requiredattribute-allowemptystrings), which means that the [default model validation](https://docs.microsoft.com/en-us/aspnet/core/mvc/models/validation?view=aspnetcore-6.0) does not allow empty strings for any string property that has this attribute. This is not reflected in the schema.

```
[Required]
public string Label { get; set; }
```

Generates schema as
```
"label": {
  "type": "string"
}
```
Where I would expect
```
"label": {
  "minLength": 1,
  "type": "string",
}
```
Does this sound like a reasonable expectation?

binginsin:
Potentially fixes #2428
When using RequiredAttribute with AllowEmptyStrings=false on a string, a minLength:1 will be added to the schema to inform that the string must not be empty. If any MinLength attributes are set, they will override this value.

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/SchemaGenerator.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/SchemaGenerator.cs
@@ -66,12 +66,14 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
                 // Nullable, ReadOnly & WriteOnly are only relevant for Schema "properties" (i.e. where dataProperty is non-null)
                 if (dataProperty != null)
                 {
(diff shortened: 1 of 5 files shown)


domaindrivendev:
Makes sense thx 👍

PR #2442, opened 2022-06-14
Add support for SecuritySchemesSelector and default implementation

captainsafia:
Addresses part of https://github.com/dotnet/aspnetcore/issues/39761 and part of https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/1661.

We've been exploring the possibility of automatically generating `OpenApiSecuritySchemes` by examining authentication/authorization state in an ASP.NET Core app and auto-generating the appropriate annotations. 

- Adds an `IAsyncSwaggerProvider` interface and implementations of `GetSwaggerAsync` to `SwaggerGenerator` to support making async calls to the authentication system
- Adds a new `SecuritySchemesSelector` to `SwaggerGeneratorOptions` for deriving `OpenApiSecuritySchemes` from `AuthenticationSchemes`

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
@@ -3,6 +3,8 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
 using System.Text.RegularExpressions;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Authentication;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.Mvc.ApiExplorer;
 using Microsoft.AspNetCore.Mvc.ModelBinding;
@@ -11,11 +13,12 @@ using Swashbuckle.AspNetCore.Swagger;
 
 namespace Swashbuckle.AspNetCore.SwaggerGen
 {
-    public class SwaggerGenerator : ISwaggerProvider
+    public class SwaggerGenerator : ISwaggerProvider, IAsyncSwaggerProvider
     {
         private readonly IApiDescriptionGroupCollectionProvider _apiDescriptionsProvider;
         private readonly ISchemaGenerator _schemaGenerator;
         private readonly SwaggerGeneratorOptions _options;
+        private readonly IAuthenticationSchemeProvider _authenticationSchemeProvider;
 
         public SwaggerGenerator(
             SwaggerGeneratorOptions options,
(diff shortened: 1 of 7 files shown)


captainsafia:
@domaindrivendev Thoughts on this? Let me know if you'd like more context on the change.

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
```diff
@@ -27,7 +30,48 @@ public class SwaggerGenerator : ISwaggerProvider
...
+            IApiDescriptionGroupCollectionProvider apiDescriptionsProvider,
+            ISchemaGenerator schemaGenerator,
+            IAuthenticationSchemeProvider authentiationSchemeProvider) : this(options, apiDescriptionsProvider, schemaGenerator)
+        {
+            _authenticationSchemeProvider = authentiationSchemeProvider;
+        }
+
+        public async Task<OpenApiDocument> GetSwaggerAsync(string documentName, string host = null, string basePath = null)
+        {
+            var (applicableApiDescriptions, swaggerDoc, schemaRepository) = GetSwaggerDocumentWithoutFilters(documentName, host, basePath);
```
domaindrivendev, line 44: What's the reasoning for pulling the document filter processing out of the shared method? The first 7 lines of both `GetSwaggerAsync` and `GetSwagger` look identical so just wondering why all of that can't be pulled up into the shared method? And, if there's a good reason, might be good to put a comment there because it's not clear.

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
```diff
@@ -47,20 +91,22 @@ public OpenApiDocument GetSwagger(string documentName, string host = null, strin
...
                 },
                 SecurityRequirements = new List<OpenApiSecurityRequirement>(_options.SecurityRequirements)
             };
 
-            var filterContext = new DocumentFilterContext(applicableApiDescriptions, _schemaGenerator, schemaRepository);
-            foreach (var filter in _options.DocumentFilters)
+            return (applicableApiDescriptions, swaggerDoc, schemaRepository);
+        }
+
+        private async Task<Dictionary<string, OpenApiSecurityScheme>> GetOpenApiSecuritySchemes()
```
domaindrivendev, line 101: As the `OpenApi` prefix is redundant, and for consistency with the other methods in this class, can we rename this method to `GenerateSecuritySchemes`

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGeneratorOptions.cs
```diff
@@ -102,5 +108,28 @@ private string DefaultSortKeySelector(ApiDescription apiDescription)
         {
             return TagsSelector(apiDescription).First();
         }
+
```
domaindrivendev, line 111: I would prefer to see this mapping of ASP.NET schemes to Open API schemes in the `SwaggerGenerator` rather than a config helper. Can you elaborate on the usecase for exposing this as an option at all? 

Also, I generally try to avoid functions that mutate parameters (see line 121). I think it would be better to have this method accept the ASP.NET schemes only and return the corresponding OpenAPI schemes. So, the option is just "bring your own" mapping of ASP.NET to Open API schemes. But again, I would avoid exposing this option unless there's a strong need vs just adding schemes via the `SecuritySchemes` property that's already on this class.

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGeneratorOptions.cs
```diff
@@ -61,6 +63,10 @@ public SwaggerGeneratorOptions()
 
         public IList<IDocumentFilter> DocumentFilters { get; set; }
```
domaindrivendev, line 65: As per my comment below - I'd prefer not accept the type that's also being returned. To be more consistent with the other "*Selector" properties in this class, could we change this signature to the following and update the default selector accordingly?
```csharp
public Func<IEnumerable<AuthenticationScheme>, Dictionary<string, OpenApiSecurityScheme>>
```

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
captainsafia: I did this as a bit of a "premature optimization" in the event that support for async filters is fully pursued per https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/1661. I've removed it since it is not explicitly required for this change.

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
captainsafia: Good call!

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGeneratorOptions.cs
captainsafia: I've updated this and moved the logic for favoring schemes provided explicitly via options over those produced by the selector in the case where scheme names collide.

captainsafia:
> @captainsafia - added some comments

I've addressed some of the feedback here. This work emerged as a result of explorations that we've been making into seeing whether it is feasible to introspect all the registered authentication schemes on an ASP.NET Core application and automatically generated the OpenApiSecurityScheme definitions for them. Turns out, this is a little gnarly, largely because authentication configuration is difficult to introspect in an ASP.NET Core app. We opted to support just the JWT-bearer based authentication which we've made easier to introspect by reading options from from config (as of .NET 7 Preview 5).

Issue #2164, opened 2021-07-15
IncludeXmlComments performance issues


coronabytes:
Version 6.1.4 on .NET 5

We were using the following code to include multiple XMLs for comments (model dlls, app parts etc).
This was working for years until a dev referenced an external commercial library with it's own XML comment file (130k lines / 7 MB) and the generation time for the swagger.json went from ~600ms to 40s+ **for each request**.

Of course filtering out this file fixed the issue, however I don't think performance degradation should be that bad for a mere 130k lines of additional xml and even if it takes 40s why the swagger.json is it taking so long for subsequent requests.

I havn't looked at your code, yet it feels like you're not using a dictionary lookup and just search the xml nodes over and over again and the end result is not cached at all.

```charp
c.IncludeXmlComments(() =>
{
    try
    {
        var directory = Path.GetDirectoryName(typeof(Startup).Assembly.Location);

        var xdocs = Directory.EnumerateFiles(directory, "*.xml")
            .Select(x => XDocument.Load(x).Element("doc")?.Descendants()).ToList();

        var doc = new XDocument(new XElement("doc", xdocs));
        return new XPathDocument(doc.CreateReader());
    }
    catch (Exception)
    {
        var doc = new XDocument(new XElement("doc"));
        return new XPathDocument(doc.CreateReader());
    }
}, true);
```

Update:
```csharp
var paramNode = _xmlNavigator.SelectSingleNode(
                $"/doc/members/member[@name='{methodMemberName}']/param[@name='{parameterInfo.Name}']");
```

eww xpath.. that explains it

domaindrivendev:
If you have ideas around improving the performance of this particular feature, I would suggest submitting a PR for it. If you think the changes are going to be significant, I would start with a 1 pager refactor proposal.

coronabytes:
Yeah thats a rather significant change.
I will consider writing a proposal in the next few weeks or so

Just take a note that the current implementation is O(N^2) and it could be O(N) parse and O(1) lookup

stevendarby:
> If you have ideas around improving the performance of this particular feature, I would suggest submitting a PR for it.

@domaindrivendev I've raised a PR for it, please review when you get a chance- thanks!
https://github.com/domaindrivendev/Swashbuckle.AspNetCore/pull/2443

domaindrivendev:
@stevendarby at a glance your PR approach makes sense to me. I’ll try do a full review and either merge or provide feedback later today.

[...]

PR #2443, opened 2022-06-19
Improve IncludeXmlComments performance

stevendarby:
This PR seeks to address the performance issues highlighted in https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/2164

The slow performance mainly stems from the use of SelectSingleNode to search for member nodes. By creating a dictionary of all the member nodes once, the nodes can later be looked up against this dictionary much more efficiently.

In one app I work on - with a few hundred endpoints and lots of XML comments - this took swagger generation down from 25 seconds to 4 seconds.

I've tried to keep changes to a minimum and backwards compatible to increase chances of an early patch or minor release. For example, the existing set of constructors on the filters remain for external backwards compatibility but are no longer used within swashbuckle itself. The new set of constructors take a dictionary instead, which allows one to be created in the extension method and shared between all of the filters. 

You may also get some further gains by fully parsing the XML document into a model of some kind, but I think that would be unnecessary as it was really the search for member nodes that was the killer.

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenOptionsExtensions.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenOptionsExtensions.cs
@@ -413,13 +413,15 @@ namespace Microsoft.Extensions.DependencyInjection
             bool includeControllerXmlComments = false)
         {
             var xmlDoc = xmlDocFactory();
-            swaggerGenOptions.ParameterFilter<XmlCommentsParameterFilter>(xmlDoc);
-            swaggerGenOptions.RequestBodyFilter<XmlCommentsRequestBodyFilter>(xmlDoc);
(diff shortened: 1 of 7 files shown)


stevendarby:
> For example, the constructor signatures are unchanged, even though another option may be to generate the dictionary once in the `IncludeXmlComments` extensions which add all these filters, and pass that dictionary to the filters. Would adding a 2nd set of constructors which take a dictionary be acceptable? The old constructors could remain for external backwards compatibility but probably not used within swashbuckle itself.

Have pushed this idea

stevendarby:
@domaindrivendev  @coronabytes

martincostello:
If you'd like to reopen this PR (and rebase it), we'll take a look into it for you.

stevendarby:
Hi @martincostello, I partly closed this because I had some concerns about memory consumption and lost confidence in it, and didn't have the time (or support) to properly investigate and measure it. There's a chance I may be able to rebase and tidy up but would need a good test :)

martincostello:
Yep - we'd definitely need tests if the changes aren't covered by existing ones 😄

Issue #2449, opened 2022-06-26
More warnings that setting OAuthConfigObject.ClientSecret will cause the client secret to show up in the html


zippy1981:
I was working on customizing the swagger UI when I noticed at the bottom of the HTML is some inline javascript that included the clientID and client secret. We were using user name and password OpenID auth against AAD so we didn't actually need the client secret. However sometime our app uses its client secret to do Bearer Auth to other things so its in the app settings, and we passed it down to swagger until we learned this was bad and harmful.

While this is ultimately swagger-ui rendering the secret, swashbuckle should do more.to warn carelessly adding the client secret. At the very least the xmldoc comments should warn you this is bad. The examples should warn you explictly not to set it whenever they set client secret. I'd even go as far as making ClientSecret a field backed property so you can stick a `#WARNING` in the setter that the user could ignore of they really need to expose the client secret.

martincostello:
#2449

PR #2450, opened 2022-06-26
Updated the comments around client secret.

zippy1981:
Addresses #2449


zippy1981:
@captainsafia can I get eyes on this because it's helping people not configure swashbuckle to expose client secrets.

Issue #2319, opened 2022-01-24
Maybe not handling DateOnly method parameters correctly?


cleames:
Hi,

Doing a .NET 6.0 solution in VS2022. Swashbuckle.AspNetCore 6.2.3 comes with it when you say you want OpenAPI documentation. One of the new features of C# 10 is being able to use DateOnly (and TimeOnly) fields, as compared to only having DateTime fields in the past. Passing in a DateTime to my test method on the Get, it works as expected:
![image](https://user-images.githubusercontent.com/93782097/150819329-f1d604f0-9c69-4a2f-bb9d-783d25d6efb8.png)

When trying a DateOnly, I see unexpected behavior:
![image](https://user-images.githubusercontent.com/93782097/150819502-65d97170-177d-4529-ace8-347c8f62efec.png)

My C# call for that looked like this:     public IActionResult Foo(DateOnly? param1).
If I change it to force it to take it from the query, meaning     public IActionResult Foo([FromQuery] DateOnly? param1)...I see this:
![image](https://user-images.githubusercontent.com/93782097/150819895-ebe3691d-ae66-4339-8061-bd924d11a797.png)

I was expecting a few things...
1. To not have to explicitly state to get it from the query.
2. To be able to enter a date only, not in parts, like we can do for datetimes.

Is this considered an issue, or does it work "as expected" and I'm just missing something here?

Thanks in advance for any thoughts on this!

mmillican:
There is a 3rd party contribution that handles this: https://github.com/maxkoshevoi/DateOnlyTimeOnly.AspNet#swagger-support. Would be nice though if it was built-in.

pdevito3:
also seeing this. thanks for linking the repo @mmillican, going to give that a whirl. agreed that it would be great to have this built in though

DamianEdwards:
Can also workaround by configuring Swashbuckle with a custom type mapping for `DateOnly`:

```c#
builder.Services.AddSwaggerGen(
    c.MapType<DateOnly>(() => new OpenApiSchema
    {
        Type = "string",
        Format = "date"
    });
);
```

captainsafia:
It seems that perhaps adding an explicit mapping here would do the trick:

[...]

PR #2480, opened 2022-08-15
Set DateOnly as string type with date format in schema

captainsafia:
Addresses #2319

### DIFF
--- a/src/Swashbuckle.AspNetCore.Newtonsoft/SchemaGenerator/NewtonsoftDataContractResolver.cs
+++ b/src/Swashbuckle.AspNetCore.Newtonsoft/SchemaGenerator/NewtonsoftDataContractResolver.cs
@@ -209,7 +209,8 @@ namespace Swashbuckle.AspNetCore.Newtonsoft
             [ typeof(Uri) ] = Tuple.Create(DataType.String, "uri"),
             [ typeof(TimeSpan) ] = Tuple.Create(DataType.String, "date-span"),
 #if NET6_0_OR_GREATER
-            [ typeof(DateOnly) ] = Tuple.Create(DataType.String, "date")
+            [ typeof(DateOnly) ] = Tuple.Create(DataType.String, "date"),
+            [ typeof(TimeOnly) ] = Tuple.Create(DataType.String, "time")
 #endif
         };
     }
--- a/src/Swashbuckle.AspNetCore.Newtonsoft/SchemaGenerator/NewtonsoftDataContractResolver.cs
+++ b/src/Swashbuckle.AspNetCore.Newtonsoft/SchemaGenerator/NewtonsoftDataContractResolver.cs
@@ -207,7 +207,10 @@ namespace Swashbuckle.AspNetCore.Newtonsoft
             [ typeof(DateTimeOffset) ] = Tuple.Create(DataType.String, "date-time"),
             [ typeof(Guid) ] = Tuple.Create(DataType.String, "uuid"),
             [ typeof(Uri) ] = Tuple.Create(DataType.String, "uri"),
-            [ typeof(TimeSpan) ] = Tuple.Create(DataType.String, "date-span")
+            [ typeof(TimeSpan) ] = Tuple.Create(DataType.String, "date-span"),
+#if NET6_0_OR_GREATER
+            [ typeof(DateOnly) ] = Tuple.Create(DataType.String, "date")
+#endif
         };
     }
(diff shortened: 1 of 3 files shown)


src/Swashbuckle.AspNetCore.Newtonsoft/SchemaGenerator/NewtonsoftDataContractResolver.cs
```diff
@@ -207,7 +207,10 @@ private IEnumerable<DataProperty> GetDataPropertiesFor(JsonObjectContract jsonOb
             [ typeof(DateTimeOffset) ] = Tuple.Create(DataType.String, "date-time"),
             [ typeof(Guid) ] = Tuple.Create(DataType.String, "uuid"),
             [ typeof(Uri) ] = Tuple.Create(DataType.String, "uri"),
-            [ typeof(TimeSpan) ] = Tuple.Create(DataType.String, "date-span")
+            [ typeof(TimeSpan) ] = Tuple.Create(DataType.String, "date-span"),
+#if NET6_0_OR_GREATER
+            [ typeof(DateOnly) ] = Tuple.Create(DataType.String, "date")
```
maxkoshevoi, line 212: What about `TimeOnly`?

captainsafia: AFAIK, OpenAPI/JSON schema doesn't have support for time only schema. We could get away with doing a string type and providing a pattern for the format but that seemed like a little more effort than it was worth given the resolver doesn't seem to be set up for this.

maxkoshevoi: I'm mainly concerned with how Swagger displays `TimeOnly` parameters. It should do it as a single field (since it's a single field under the hood)

image

Instead, currently it displays it as multiple fields, which makes it somewhat difficult to use:

image

Not sure that ASP.NET Core 7 will be able to bind `string` to `TimeOnly` out-of-the-box (since https://github.com/dotnet/aspnetcore/issues/34591#issuecomment-1225997439 got moved to ASP.NET Core 8), but same applies to `DateOnly`. If it won't be able to do it for `TimeOnly`, it won't be able to do it for `DateOnly` as well.

But NSwag resolves this dynamically somehow. If it sees `TypeConverter` to `string`, it automatically switches to displaying those types as one field.

captainsafia: > Not sure that ASP.NET Core 7 will be able to bind string to TimeOnly out-of-the-box (since https://github.com/dotnet/aspnetcore/issues/34591#issuecomment-1225997439 got moved to ASP.NET Core 8), but same applies to DateOnly. If it won't be able to do it for TimeOnly, it won't be able to do it for DateOnly as well.

At the moment, minimal APIs is the only framework implementation that supports `TimeOnly`/`DateOnly` in routing. Although what the framework does is fairly independently from what Swashbuckle/NSwag does. We could totally add `TimeOnly` support here, we'd just have to figure out what schema to write. It looks like NSwag is using a "time" format for the string.

https://github.com/RicoSuter/NJsonSchema/blob/a1053c83dd798defdd0a83f98f379cadadcb1e64/src/NJsonSchema/Generation/DefaultReflectionService.cs#L151-L155

maxkoshevoi: I also used `time` in the NuGet: https://github.com/maxkoshevoi/DateOnlyTimeOnly.AspNet/blob/d239472270aaac83196a2ce1d701e60e3af670b9/DateOnlyTimeOnly.AspNet.Swashbuckle/Extensions/SwaggerGenOptionsExtensions.cs#L18

Seems like a logical name given `date-time` and `date` exists

captainsafia: Added support for `TimeOnly`.

maxkoshevoi (approved):

aheubusch:
@maxkoshevoi Ping. Is it possible to re run the checks so this gets merged?

captainsafia:
cc: @domaindrivendev should be able to help with that.

PR #2515, opened 2022-10-06
Fix incorrect serialization of double and float when using Swashbuckle.AspNetCore.Cli

### LINKED ISSUE #2105: Set culture of Json Serializer to InvariantCulture
Version 6.1.1

I noticed a bug where the examples in our swagger.json files that were generated did not include correctly formatted values for the data type "double". Instead of being formatted as "123.123" it was being formatted as "123,123". The issue was that our application is using the sv-SE culture `CultureInfo.DefaultThreadCurrentCulture = new CultureInfo("sv-SE");`.

I imagine a lot of applications want to specify their own culture, but shouldn't the generated swagger.json always be using InvariantCulture so we won't have to face these issues?

I know this is an easy fix for Newtonsoft.Json, but since Swashbuckle is using System.Text.Json, I'm not sure what the best way to solve this is, otherwise I would have submitted a PR.

dinar007007:
Fixes #2105 

When serializing to a file, the CultureInfo.InvariantCulture used.

### DIFF
--- a/src/Swashbuckle.AspNetCore.Cli/Program.cs
+++ b/src/Swashbuckle.AspNetCore.Cli/Program.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Reflection;
 using System.Diagnostics;
+using System.Globalization;
 using System.IO;
 using System.Runtime.Loader;
 using System.Linq;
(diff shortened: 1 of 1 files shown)


dinar007007:
@domaindrivendev can you be so kind as to comment on the pull request?

martincostello:
Thanks for contributing - if you'd like to continue with this pull request, please rebase against the default branch to pick up our new CI. We'd also like a test.

PR #2522, opened 2022-10-22
Replacing thread-safe version dictionary objects for SchemaRepository

shmilyyan:
I am trying to replace the GeneratePaths part of the logic of SwaggerGenerator with concurrent execution. I accidentally found that the dictionary storage structure in the SchemaRepository object is not thread-safe, which causes the logic of concurrent construction to not proceed normally, prompting an exception when adding a dictionary, here by replacing the storage structure solves this problem


martincostello:
Thanks for contributing - if you'd like to continue with this pull request, please rebase against the default branch to pick up our new CI. Do the existing tests appropriately cover this change?

PR #2526, opened 2022-10-25
Handle Stream and PipeReader content types correctly

martincostello:
- Do not generate schema models for `Stream` and `PipeReader`.
- Only allocate the array once, rather than on every call.
- Fix warning by removing unused parameter from private method.

/cc @captainsafia

Relates to item 3 in https://github.com/dotnet/aspnetcore/issues/44677.


martincostello:
I'm declaring bankruptcy on this PR. Will come back to it maybe if/when this project shows some signs of life in the future.

martincostello:
Moved to #2784.

PR #2533, opened 2022-11-03
Support new `UseInlineDefinitionsForObjects` flag

sol-wasserman:
# Problem

When a request/response has a nested object, there are two pieces of information missing from the generated doc file:

1. The property's `summary` is not displayed.
2. The `Nullable` marker is not shown for `nullable` properties.

---

## Here's a screenshot of what it looks like now:

image

## Here's a screenshot of the source code for the above:

image

As you can see, this property has a nice `summary`, and is `nullable`.

## Here's a screenshot of what it I would expect:

image

# Underlying Cause

It looks to me like the reason it's not generating that properly is because the object is only a reference, and Swagger doesn't support these on references. As [the Swagger specification](https://swagger.io/specification/#reference-object) says, regarding `$ref`s:

> This object cannot be extended with additional properties and any properties added SHALL be ignored.

# Proposed Solution

We have [previously](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/commit/9abdbf99a271de5decb2bd93c5c614632e750b32) introduced an option to `UseInlineDefinitionsForEnums`. Among other things, this will show the summary of a property that is an enum, and will also show the `Nullable` marker accordingly.

This PR adds a similar `UseInlineDefinitionsForObjects` option. When enabled, it'll inline the whole schema, and will properly show the `summary` and whether it's `nullable`:

image

> Copied from https://github.com/domaindrivendev/Swashbuckle.AspNetCore/pull/2384 which was accidentally closed.

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/SchemaGenerator.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/SchemaGenerator.cs
@@ -244,7 +244,7 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
                 case DataType.Object:
                     {
                         schemaFactory = () => CreateObjectSchema(dataContract, schemaRepository);
-                        returnAsReference = true;
+                        returnAsReference = !_generatorOptions.UseInlineDefinitionsForObjects;
                         break;
                     }
 
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/ConfigureSchemaGeneratorOptions.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/ConfigureSchemaGeneratorOptions.cs
@@ -32,6 +32,7 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
         {
             target.CustomTypeMappings = new Dictionary<Type, Func<OpenApiSchema>>(source.CustomTypeMappings);
             target.UseInlineDefinitionsForEnums = source.UseInlineDefinitionsForEnums;
+            target.UseInlineDefinitionsForObjects = source.UseInlineDefinitionsForObjects;
             target.SchemaIdSelector = source.SchemaIdSelector;
             target.IgnoreObsoleteProperties = source.IgnoreObsoleteProperties;
             target.UseAllOfForInheritance = source.UseAllOfForInheritance;
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenOptionsExtensions.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenOptionsExtensions.cs
@@ -205,6 +205,14 @@ namespace Microsoft.Extensions.DependencyInjection
         }
 
         /// <summary>
+        /// Generate inline schema definitions (as opposed to referencing a shared definition) for complex objects
+        /// </summary>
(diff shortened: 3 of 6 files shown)


ColKrumpler:
@domaindrivendev could this be evaluated to be merged in?

ColKrumpler:
@domaindrivendev could this be evaluated to be merged in?

ColKrumpler:
@domaindrivendev could this be evaluated to be merged in?

src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/SchemaGenerator.cs
```diff
@@ -244,7 +244,7 @@ private OpenApiSchema GenerateConcreteSchema(DataContract dataContract, SchemaRe
                 case DataType.Object:
                     {
                         schemaFactory = () => CreateObjectSchema(dataContract, schemaRepository);
-                        returnAsReference = true;
+                        returnAsReference = !_generatorOptions.UseInlineDefinitionsForObjects;
```
jamesmcroft, line 247: From the original PR for this change #2426, you'll want to add support for self-referencing types:

```csharp
returnAsReference = !_generatorOptions.UseInlineDefinitionsForObjects ||
                                        dataContract.ObjectProperties.Any(prop =>
                                            prop.MemberType.IsAssignableFrom(dataContract.UnderlyingType) ||
                                            prop.MemberType.GenericTypeArguments.Any(type =>
                                                type.IsAssignableFrom(dataContract.UnderlyingType)));
```

jamesmcroft: The test for this to also validate the change:

```csharp
        [Fact]
        public void GenerateSchema_SupportsOption_UseReferenceForSelfReferenceTypesWhenUseInlineDefinitionsForObjects()
        {
            var subject = Subject(
                configureGenerator: c => c.UseInlineDefinitionsForObjects = true
            );

            var schema = subject.GenerateSchema(typeof(SelfReferencingType), new SchemaRepository());

            Assert.NotNull(schema.Reference);
            Assert.Equal("SelfReferencingType", schema.Reference.Id);
        }
```

martincostello:

martincostello (commented):
This seems reasonable to me - I think I've even researched this exact lack of summary (and maybe the nullability) myself before in the past.

If have no specific feedback other than than it looks like there's some comments from jamesmcroft that should be looked at.

sol-wasserman:
Good catch @jamesmcroft. However, your code doesn't correctly handle self referencing objects. Think of this simple example:

```c#
public record Car(Driver Driver);

public record Driver(Car Car);
```


Additionally, your solution will use a reference when not called for. For example:

```c#
public class MetadataAboutComplexType<T>
{
    public string Namespace { get; set; }
}
```

...should be inlined but will now incorrectly just be a reference.

martincostello:
Can you add a test that exhibits the above is handled correctly, whether or not the non-test code is changed? If it then isn't, we should fix that somehow.

sol-wasserman:
@jamesmcroft I got a little stuck trying to solve this. Your linked piece of code correctly solves for self-references, but I couldn't figure out how to solve for circular references that may be buried deeper into the tree.

Do you have any ideas?

cervengoc:
Thank you for this PR, this is exactly what I was investigating. However, I'd like to suggest a little generalization.

I would like to have this option as a callback receiving at least the contract type which is about to be generated. So that way one can configure this in a much more flexible way, for example using this mode only for several types.

NicolasHumann:
any news ?

PR #2535, opened 2022-11-07
Include InferSecuritySchemes in deep copy of SwaggerGeneratorOptions

danroth27:


martijn1279:
#2540 -> same PR, only that one some extra fixes

PR #2540, opened 2022-11-09
Fix bug in the extension method 'InferSecuritySchemes' so it will actually enable it

martijn1279:
When using the extension method `InferSecuritySchemes` it wont actually do anything because its not being deepcopied to the new object

```cs
options.InferSecuritySchemes(schemes => new Dictionary<string, OpenApiSecurityScheme>
{
    ["OAuth2"] = new OpenApiSecurityScheme
    {
        Type = SecuritySchemeType.OAuth2,
        Name = "OAuth2",
    },
});
```


martincostello:
Thanks for contributing - if you'd like to continue with this pull request, please rebase against the default branch to pick up our new CI. We'd also like a test.

martincostello:
Superseded by #2796.

PR #2542, opened 2022-11-10
Update to .NET 7.0

eslamisepehr:
Hello 👋

Update Projects to .NET 7.0.

Notice:
- One test (`Validate_ThrowsException_IfExpectedContentIsNotPresent`) failed when updated to .NET 7.0.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -8,7 +8,7 @@ init:
 - git config --global core.autocrlf true
 
 environment:
-  DOTNET_VERSION: "7.0.0"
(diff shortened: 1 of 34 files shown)


htuomola:
I was curious why the build fails (it fails to download .net 7). Seems that it relies on global.json version but dotnet-install script doesn't honor "rollForward" policy in global.json but tries to download 7.0.0 instead. That doesn't exist, correct version would be 7.0.100. 
So, I think that could be used in global.json instead - or change build to call dotnet-install script with parameter `-channel 7.0` instead?

Issue #2530, opened 2022-10-31
Root tags object not showing


greg-signi:
I set the attribute [Tags("...")] in the Controler and despite having my openapi.json object with tags under the designated methods, eg:

`"paths": {
    "/sessions": {
      "post": {
        "tags": [
          "Authentication Session"
        ],
        "summary": "Create a session",`
        
The root object tags is not created. Any idea what's missing ?

domaindrivendev:
I'm not familiar with the `TagsAttribute` you're referring to. Could you please provide more info (sample code, expected output etc.).

greg-signi:
Hello, either the [Tags("User")] from Microsoft.AspNetCore.Http or `[SwaggerOperation(Tags = new[] { "User" }] from Swashbuckle.Swagger.Annotations`

And they work in some instance since they are able to get appended to the method information `"paths": { "/sessions": { "post": { "**tags**": [ "Authentication Session" ], "summary": "Create a session",` 

But the array "tags" at the root level object, remains empty. I did a workaround where I create a `IDocumentFilter` and fill the array `swaggerDoc.Tags` by hand.. But I shouldn't have to do this right ?

yingxinj:
Hello, please see the linked PR https://github.com/domaindrivendev/Swashbuckle.AspNetCore/pull/2565 that adds support for the `TagsAttribute` - let me know if this would be helpful for resolving your issue, or if you had a different approach to implementing a workaround @greg-signi?

personball:
Hi, any update for ” the array tags at the root level object, that encompasses all the tags available in the api, remains empty“ ？

martincostello:
https://github.com/domaindrivendev/Swashbuckle.AspNetCore/pull/2565#issuecomment-2053577323

Issue #2162, opened 2021-07-11
Controllers not sorted alphabetically when using includeControllerXmlComments


lonix1:
## Environment
aspnet: 5.0.301
Swashbuckle.AspNetCore: 6.1.4

# Background
I created a minimal swashbuckle implementation. All I have is:
````cs
o.IncludeXmlComments(path, includeControllerXmlComments: true)
o.CustomSchemaIds(x => x.FullName);
````

And I added `<summary>` xml comments to my controllers.

## Problem
If I use `includeControllerXmlComments:false` then the controllers are sorted alphabetically.

But if I use `includeControllerXmlComments:true` then the controllers are not sorted alphabetically.

**Comment by lscorcia** [comment]
+1 for this.
As a workaround, you can use the following document filter:

```
/// <summary>
/// Workaround for https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/2162
/// When adding XML controller descriptions to the swagger description document,
/// controllers are listed out of alphabetical order.
///
/// This filter explicitly reorders them.
/// </summary>
public class TagReorderDocumentFilter : IDocumentFilter
{
  /// <summary>
  /// Allows customization of the swagger description document
  /// </summary>
  /// <param name="swaggerDoc">The generated swagger description document</param>
  /// <param name="context">Context information</param>
  public void Apply(OpenApiDocument swaggerDoc, DocumentFilterContext context)
  {
    swaggerDoc.Tags = swaggerDoc.Tags
      .OrderBy(tag => tag.Name)
      .ToList();
  }
}
```

To apply the filter add it during Swagger initialization:
```
builder.Services.AddSwaggerGen(config => {
  ...

  // Workaround: After adding XML controller descriptions, they are listed out of alphabetical order
  config.DocumentFilter<TagReorderDocumentFilter>();
});
```

lonix1:
In newer versions, this seems to work without the need for any customisation.

ASP.NET Core 5.0.11
Swashbuckle.AspNetCore 6.2.3

lscorcia:
Unfortunately the issue is still presente. ASP.NET Core 6.0 + Swashbuckle.AspNetCore 6.3.0.

lonix1:
Hey @lscorcia I've reopened for you.

lscorcia:
Thanks @lonix1 ! I prepared a small repro project at https://github.com/lscorcia/UnsortedControllersRepro . Please let me know if you need any additional information.

[...]

Issue #1757, opened 2020-07-27
Unable to change group order with `OrderActionsBy` when only some group have description 


yyjdelete:
Swashbuckle.AspNetCore 5.5.1

When `IncludeXmlComments(..., includeControllerXmlComments: true)` is used, but only some controllers have comments, it's unable to change the order of groups with `OrderActionsBy`(the ones with description is always the first)

Seems `Swashbuckle.AspNetCore` only gen tags with Description in OpenApiDocument.Tags, and according to https://swagger.io/specification/#fixed-fields,`The tags that are not declared MAY be organized randomly or based on the tools' logic.`

Is there an options to force `Swashbuckle.AspNetCore` generate all tags?

yingxinj:
I have just opened a PR #2565 that would fix this issue by adding all Controller tags to the root tags object, not just tags with description, please let me know if it is helpful :) the root tags are then sorted by name as a sensible default, but having the tags object be a complete list means that you can sort them in other ways if preferred.

martincostello:
#2565

Issue #467, opened 2017-08-18
TagActionsBy custom tag or controller name


djwales:
Hi,

I was looking for a solution where I could group the actions of one or more controllers under a custom tag. I did a bit of searching round and found this [section in the documentation](https://github.com/domaindrivendev/Swashbuckle.AspNetCore#customize-operation-tags-eg-for-ui-grouping). I dug into it and managed to produce a solution where I assign a custom attribute to controller(s) that I want to be tagged together. I then created an extension method for the ApiDescription class that will either extract the name in my custom attribute or use the name of the controller if the attribute is not present. This has worked, however, it was a little painful getting the name of the controller when my attribute was not present. I ended up having to duplicate some of the code in the [ApiDescriptionExtensions](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/1b8e109cdb2e21b57584ef4596a0d1a13af8b7eb/src/Swashbuckle.AspNetCore.SwaggerGen/Generator/ApiDescriptionExtensions.cs) class. Ideally, I would have just used the ```ControllerName``` extension method but this has been marked as internal. 

I have two questions:

1. Is there a reason why the ```ControllerName``` extension method is marked as internal? If not, would it be possible to make it public?
2. Is there an alternative approach to what I have done which achieves the same result?

Here is the code that I have produced to achieve this. 

```csharp
public class SwaggerGroupAttribute : Attribute
{
    public string GroupName { get; }

    public SwaggerGroupAttribute(string groupName)
    {
        GroupName = groupName;
    }
}

public static class ApiDescriptionExtensions
{
    public static string GroupBySwaggerGroupAttribute(this ApiDescription api)
    {
        var groupNameAttribute = (SwaggerGroupAttribute)api.ControllerAttributes().SingleOrDefault(attribute => attribute is SwaggerGroupAttribute);

        // ------
        // Lifted from ApiDescriptionExtensions
        var actionDescriptor = api.GetProperty<ControllerActionDescriptor>();
            
        if (actionDescriptor == null)
        {
            actionDescriptor = api.ActionDescriptor as ControllerActionDescriptor;
            api.SetProperty(actionDescriptor);
        }
        // ------

        return groupNameAttribute != null ? groupNameAttribute.GroupName : actionDescriptor?.ControllerName;
    }
}

// In Startup.cs
public void ConfigureServices(IServiceCollection services)
{
    // ...
    services.AddSwaggerGen(c => 
    {
[...]

PR #2565, opened 2022-12-05
Support for associating XML Comments with custom tags defined using TagsAttribute

yingxinj:
## Description & reference issues

As of .NET 6.0, Swagger/OpenAPI supports using [TagsAttribute](https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.http.tagsattribute?view=aspnetcore-7.0) to rename each group of endpoints, overriding the default value of the controller name. (See also e.g. [the top voted answer by Raul on this SO question](https://stackoverflow.com/questions/37076173/is-there-a-way-change-the-controllers-name-in-the-swagger-ui-page))

The existing implementation of Swashbuckle's XmlCommentsDocumentFilter associates the controller's `<summary>` XML comment with the controller name. It currently has two limitations:

1. It does not support using TagsAttribute - it only links the summary node with the controller name. See the last four comments on #467, these last comments refer to the new TagsAttribute instead of TagActionsBy.

2. It only adds tags for controllers with a summary node. If the user has summary nodes for some controllers but not others, then the root tags object will only include a partial list of tags. This results in the side-effect that tags are not displayed in alphabetical order. Instead, tags with descriptions are displayed first, followed by tags without descriptons. Related issues:
    - #2162  
    - #1757

Another linked issue #2530  is a result of both these limitations.

This PR fixes both these issues. If IncludeXmlComments is true: 
 - then the resulting tag names & descriptions will support the new TagsAttribute, falling back to the default (controller name) if TagsAttribute is not used.
 - then the resulting root tags object will contain all controller-level tags, regardless of whether they have an XML summary node or not. Those without a summary node will not have any "description" field.

## Tests

Tests are included for the target framework .NET 6.0, which supports TagsAttribute. The SwaggerGen.Tests project currently only targets .NET 6.0, so I did not add unit tests for earlier frameworks. I built the project in .NET 5.0 and did a manual test of this case to check it works.

## Potential improvements

This PR still has some limitations:
 - TagsAttribute theoretically allows you to specify more than one custom tag e.g. by `[Tags("first tag", "second tag")]` however I don't see this as a primary use case in generating the swagger file, and there is no way of providing separate XML summaries for the two tags, so this is not supported in this PR - only the first tag is supported.
 - TagsAttribute can be specified at the method level, if so, such tags would still be omitted from the root tags object.

If it's actually useful, I would be happy to add support for these cases.

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenOptionsExtensions.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenOptionsExtensions.cs
@@ -422,7 +422,9 @@ namespace Microsoft.Extensions.DependencyInjection
         /// <param name="xmlDocFactory">A factory method that returns XML Comments as an XPathDocument</param>
         /// <param name="includeControllerXmlComments">
         /// Flag to indicate if controller XML comments (i.e. summary) should be used to assign Tag descriptions.
-        /// Don't set this flag if you're customizing the default tag for operations via TagActionsBy.
+        /// If customizing the default tag for operations via TagsAttribute, only the first Tag per controller will be
+        /// associated with the description.
+        /// However, don't set this flag if you're customizing the default tag for operations via TagActionsBy.
         /// </param>
         public static void IncludeXmlComments(
             this SwaggerGenOptions swaggerGenOptions,
@@ -446,7 +448,9 @@ namespace Microsoft.Extensions.DependencyInjection
         /// <param name="filePath">An absolute path to the file that contains XML Comments</param>
         /// <param name="includeControllerXmlComments">
         /// Flag to indicate if controller XML comments (i.e. summary) should be used to assign Tag descriptions.
-        /// Don't set this flag if you're customizing the default tag for operations via TagActionsBy.
+        /// If customizing the default tag for operations via TagsAttribute, only the first Tag per controller will be
+        /// associated with the description.
+        /// However, don't set this flag if you're customizing the default tag for operations via TagActionsBy.
         /// </param>
         public static void IncludeXmlComments(
             this SwaggerGenOptions swaggerGenOptions,
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/KeyValuePairExtensions.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/KeyValuePairExtensions.cs
@@ -1,0 +1,14 @@
+using System.Collections.Generic;
+
+namespace Swashbuckle.AspNetCore.SwaggerGen
+{
+    public static class KeyValuePairExtensions
+    {
+        // Explicit deconstruct required for older .NET frameworks
+        public static void Deconstruct<TKey, TValue>(this KeyValuePair<TKey, TValue> kvp, out TKey key, out TValue value)
+        {
+            key = kvp.Key;
+            value = kvp.Value;
+        }
+    }
+}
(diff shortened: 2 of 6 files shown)


NilkOne:
Great !
When will this fix added on NuGet?

silkfire:
@domaindrivendev Any ETA on this?

Skintkingle:
We could really do with this change.

sirmeepington:
This would be a lifesaver 👍

DenKn:
This fix does not cover all cases. For example it won't work

```
 public abstract class ConfigurationControllerBase : ControllerBase
{
}

[Tags("ConfigurationController ")]
public class ConfigurationController : ConfigurationControllerBase 
{
}
```

martincostello:

src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenOptionsExtensions.cs
```diff
@@ -422,7 +422,9 @@ public static void SupportNonNullableReferenceTypes(this SwaggerGenOptions swagg
         /// <param name="xmlDocFactory">A factory method that returns XML Comments as an XPathDocument</param>
         /// <param name="includeControllerXmlComments">
         /// Flag to indicate if controller XML comments (i.e. summary) should be used to assign Tag descriptions.
-        /// Don't set this flag if you're customizing the default tag for operations via TagActionsBy.
+        /// If customizing the default tag for operations via TagsAttribute, only the first Tag per controller will be
+        /// associated with the description.
+        /// However, don't set this flag if you're customizing the default tag for operations via TagActionsBy.
```
martincostello, line 427: Let's make `TagsAttribute` and `TagActionsBy` `<see .... />` references in these two comments.

src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/KeyValuePairExtensions.cs
```diff
@@ -0,0 +1,14 @@
...
+{
+    public static class KeyValuePairExtensions
+    {
+        // Explicit deconstruct required for older .NET frameworks
+        public static void Deconstruct<TKey, TValue>(this KeyValuePair<TKey, TValue> kvp, out TKey key, out TValue value)
+        {
+            key = kvp.Key;
+            value = kvp.Value;
+        }
+    }
```
martincostello, line 13: It would be better to only add this polyfill for the TFMs that need it. Also, can it be `internal`?

src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsDocumentFilter.cs
```diff
@@ -21,34 +23,63 @@ public XmlCommentsDocumentFilter(XPathDocument xmlDoc)
...
-                        });
-                    }
+                    Name = controllerInfo.CustomTagName ?? controllerName,
+                    Description = description
+                });
+            }
+            swaggerDoc.Tags = swaggerDoc.Tags.OrderBy(x => x.Name).ToList();
+        }
+
+        private class ControllerInfo
```
martincostello, line 50:
```suggestion
        private sealed class ControllerInfo
```

src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsDocumentFilter.cs
```diff
@@ -21,34 +23,63 @@ public XmlCommentsDocumentFilter(XPathDocument xmlDoc)
...
+        private static ControllerInfo GetControllerInfo(IGrouping<string, ControllerActionDescriptor> group)
+        {
+            var controllerInfo = new ControllerInfo
+            {
+                ControllerType = group.First().ControllerTypeInfo.AsType()
+            };
+
+#if NET6_0_OR_GREATER
+            controllerInfo.CustomTagName =
+                group.First().MethodInfo.DeclaringType?.GetCustomAttribute<TagsAttribute>()?.Tags[0];
```
martincostello, line 65: This looks like it'll throw an exception if `Tags` is empty or `null`.

PR #2610, opened 2023-02-11
Added Swashbuckle.AspNetCore.Annotations.SwaggerIgnoreAttribute

jcracknell:
Application of the SwaggerIgnoreAttribute suppresses the annotated property during schema generation, without altering the serialization behavior as with JsonIgnoreAttribute.

This is useful in a scenario where you want to capture and reject unhandled properties, without advertising that you accept `additionalProperties` (which is the opposite of what is intended):

```cs
public abstract class RequestBase : IValidatableObject {
    [SwaggerIgnore]
    [JsonExtensionData]
    public IDictionary<string, JsonElement>? UnsupportedProperties { get; set; }

    public virtual IEnumerable<ValidationResult> Validate(ValidationContext validationContext) {
        if(UnsupportedProperties is { Count: not 0 })
            foreach(var entry in UnsupportedProperties)
                yield return new ValidationResult($"Unsupported property: {entry.Key}.", new[] { entry.Key });
    }
}
```

I added the annotation to the Swashbuckle.AspNetCore.SwaggerGen project as it is much easier to filter the properties in the JsonSerializerDataContractResolver than after the fact in a filter, as you would need to regenerate the data contract for the type and then remove all of the annotated properties and derived side effects.

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/Annotations/SwaggerIgnoreAttribute.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/Annotations/SwaggerIgnoreAttribute.cs
@@ -1,0 +1,16 @@
+using System;
+
+// ReSharper disable once CheckNamespace
+namespace Swashbuckle.AspNetCore.Annotations
+{
+    /// <summary>
+    /// Causes the annotated member to be ignored during schema generation.
+    /// Does not alter serialization behavior.
+    /// </summary>
+    /// <remarks>
+    /// Can be used in combination with <see cref="System.Text.Json.Serialization.JsonExtensionDataAttribute"/>
+    /// to capture and invalidate unsupported properties.
+    /// </remarks>
+    [AttributeUsage(AttributeTargets.Property)]
+    public class SwaggerIgnoreAttribute : Attribute { }
+}
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/Annotations/SwaggerIgnoreAttribute.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/Annotations/SwaggerIgnoreAttribute.cs
@@ -11,6 +11,6 @@ namespace Swashbuckle.AspNetCore.Annotations
     /// Can be used in combination with <see cref="System.Text.Json.Serialization.JsonExtensionDataAttribute"/>
     /// to capture and invalidate unsupported properties.
     /// </remarks>
-    [AttributeUsage(AttributeTargets.Method | AttributeTargets.Property)]
+    [AttributeUsage(AttributeTargets.Method | AttributeTargets.Parameter | AttributeTargets.Property)]
     public class SwaggerIgnoreAttribute : Attribute { }
 }
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/Annotations/SwaggerIgnoreAttribute.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/Annotations/SwaggerIgnoreAttribute.cs
@@ -12,5 +12,5 @@ namespace Swashbuckle.AspNetCore.Annotations
     /// to capture and invalidate unsupported properties.
     /// </remarks>
     [AttributeUsage(AttributeTargets.Method | AttributeTargets.Parameter | AttributeTargets.Property)]
-    public class SwaggerIgnoreAttribute : Attribute { }
+    public sealed class SwaggerIgnoreAttribute : Attribute { }
 }
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/Annotations/SwaggerIgnoreAttribute.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/Annotations/SwaggerIgnoreAttribute.cs
@@ -11,6 +11,6 @@ namespace Swashbuckle.AspNetCore.Annotations
     /// Can be used in combination with <see cref="System.Text.Json.Serialization.JsonExtensionDataAttribute"/>
     /// to capture and invalidate unsupported properties.
     /// </remarks>
-    [AttributeUsage(AttributeTargets.Property)]
+    [AttributeUsage(AttributeTargets.Method | AttributeTargets.Property)]
     public class SwaggerIgnoreAttribute : Attribute { }
 }
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
@@ -321,6 +321,7 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
                 {
                     return (!apiParam.IsFromBody() && !apiParam.IsFromForm())
                         && (!apiParam.CustomAttributes().OfType<BindNeverAttribute>().Any())
+                        && (!apiParam.CustomAttributes().OfType<SwaggerIgnoreAttribute>().Any())
                         && (apiParam.ModelMetadata == null || apiParam.ModelMetadata.IsBindingAllowed);
                 });
 
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
@@ -9,6 +9,7 @@ using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.Mvc.ApiExplorer;
 using Microsoft.AspNetCore.Mvc.ModelBinding;
 using Microsoft.OpenApi.Models;
+using Swashbuckle.AspNetCore.Annotations;
 using Swashbuckle.AspNetCore.Swagger;
 #if NET7_0_OR_GREATER
 using Microsoft.AspNetCore.Http.Metadata;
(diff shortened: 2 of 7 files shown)


Havunen:
Maybe the attribute should be called something else from "ignore" ie. SwaggerHiddenAttribute to indicate it is not the same thing.

martincostello:

jcracknell:
Done. I've also added commits to support the attribute on operations and parameters.

I'm aware of the current recommended approach of excluding methods using ApiExplorerSettingsAttribute, but there's something to be said for consistency, and there's also a difference between excluding the operation from Swashbuckle and excluding it from the ApiExplorer.

src/Swashbuckle.AspNetCore.SwaggerGen/Annotations/SwaggerIgnoreAttribute.cs
```diff
@@ -0,0 +1,16 @@
...
+{
+    /// <summary>
+    /// Causes the annotated member to be ignored during schema generation.
+    /// Does not alter serialization behavior.
+    /// </summary>
+    /// <remarks>
+    /// Can be used in combination with <see cref="System.Text.Json.Serialization.JsonExtensionDataAttribute"/>
+    /// to capture and invalidate unsupported properties.
+    /// </remarks>
+    [AttributeUsage(AttributeTargets.Method | AttributeTargets.Parameter | AttributeTargets.Property)]
```
martincostello, line 14: System.Text.Json also supports fields for serialisation, so should this too?

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
```diff
@@ -83,6 +84,7 @@ public OpenApiDocument GetSwagger(string documentName, string host = null, strin
             var applicableApiDescriptions = _apiDescriptionsProvider.ApiDescriptionGroups.Items
                 .SelectMany(group => group.Items)
                 .Where(apiDesc => !(_options.IgnoreObsoleteActions && apiDesc.CustomAttributes().OfType<ObsoleteAttribute>().Any()))
+                .Where(apiDesc => !apiDesc.CustomAttributes().OfType<SwaggerIgnoreAttribute>().Any())
```
martincostello, line 87: Could this be `!HasAttribute<T>()`? I just saw it in the previous file and it looks terser than this.

test/Swashbuckle.AspNetCore.SwaggerGen.Test/Fixtures/SwaggerIngoreAnnotatedType.cs
```diff
@@ -0,0 +1,18 @@
+using System.Collections.Generic;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using Swashbuckle.AspNetCore.Annotations;
+
+namespace Swashbuckle.AspNetCore.SwaggerGen.Test.Fixtures {
```
martincostello, line 6:
```suggestion
namespace Swashbuckle.AspNetCore.SwaggerGen.Test.Fixtures
{
```

test/Swashbuckle.AspNetCore.SwaggerGen.Test/SchemaGenerator/JsonSerializerSchemaGeneratorTests.cs
```diff
@@ -786,6 +787,21 @@ public void GenerateSchema_HonorsSerializerAttribute_JsonExtensionData()
             Assert.Null(schema.AdditionalProperties.Type);
         }
 
+        [Fact]
+        public void GenerateSchema_HonorsAttribute_SwaggerIgnore() {
```
martincostello, line 791:
```suggestion
        public void GenerateSchema_HonorsAttribute_SwaggerIgnore()
        {
```

src/Swashbuckle.AspNetCore.SwaggerGen/Annotations/SwaggerIgnoreAttribute.cs
jcracknell: I actually didn't realize JsonSerializerOptions had been added for this in .NET 5, however it looks as though significant portions of the Swashbuckle codebase only consider properties (e.g. looking at the CustomAttributes extension used elsewhere), so I'm hesitant to take on this change.

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
jcracknell: The behavior of this particular extension appears to be provided by the builtin PropertyInfo.IsDefined(Type). I've removed this extension on PropertyInfo, and added extensions with the same name on ApiDescription and ApiParameterDescription, as in this case the extension is required to handle e.g. property and parameter sourced API parameters.

It also occurred to me that the ApiExplorer provided models include the instantiated attributes as metadata, however I am unclear how this might affect synthetically generated instances in unit tests.

src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/PropertyInfoExtensions.cs
```diff
@@ -5,12 +5,6 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
 {
     public static class PropertyInfoExtensions
     {
-        public static bool HasAttribute<TAttribute>(this PropertyInfo property)
-            where TAttribute : Attribute
-        {
-            return property.GetCustomAttribute<TAttribute>() != null;
-        }
-
```
martincostello, line 13: Removing this is a breaking change because it's `public`.

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/ApiDescriptionExtensions.cs
```diff
@@ -33,12 +33,26 @@ public static bool TryGetMethodInfo(this ApiDescription apiDescription, out Meth
...
+        public static bool HasAttribute(this ApiDescription apiDescription, Type attributeType)
+        {
+            if(!apiDescription.TryGetMethodInfo(out var methodInfo))
+                return false;
+
+            return methodInfo.IsDefined(attributeType)
+                || true == methodInfo.DeclaringType?.IsDefined(attributeType);
+        }
+
+
```
martincostello, line 49:
```suggestion
        public static bool HasAttribute(this ApiDescription apiDescription, Type attributeType)
        {
            if (!apiDescription.TryGetMethodInfo(out var methodInfo))
                return false;

            return methodInfo.IsDefined(attributeType)
                || methodInfo.DeclaringType?.IsDefined(attributeType) is true;
        }

```

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/ApiDescriptionExtensions.cs
```diff
@@ -33,12 +33,26 @@ public static bool TryGetMethodInfo(this ApiDescription apiDescription, out Meth
             return false;
         }
 
+        public static bool HasAttribute<T>(this ApiDescription apiDescription)
+            where T : Attribute =>
+            apiDescription.HasAttribute(typeof(T));
+
+        public static bool HasAttribute(this ApiDescription apiDescription, Type attributeType)
```
martincostello, line 40: Missing null check on `apiDescription`.

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/ApiParameterDescriptionExtensions.cs
```diff
@@ -66,6 +66,20 @@ public static PropertyInfo PropertyInfo(this ApiParameterDescription apiParamete
...
+        public static bool HasAttribute<T>(this ApiParameterDescription apiParameter)
+            where T : Attribute =>
+            apiParameter.HasAttribute(typeof(T));
+
+        public static bool HasAttribute(this ApiParameterDescription apiParameter, Type attributeType)
+        {
+            if(apiParameter.PropertyInfo() is { } propertyInfo)
+                return propertyInfo.IsDefined(attributeType);
+            if(apiParameter.ParameterInfo() is { } parameterInfo)
+                return parameterInfo.IsDefined(attributeType);
```
martincostello, line 78:
```suggestion
            if (apiParameter.PropertyInfo() is { } propertyInfo)
                return propertyInfo.IsDefined(attributeType);
            if (apiParameter.ParameterInfo() is { } parameterInfo)
                return parameterInfo.IsDefined(attributeType);
```

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/ApiParameterDescriptionExtensions.cs
```diff
@@ -66,6 +66,20 @@ public static PropertyInfo PropertyInfo(this ApiParameterDescription apiParamete
                 : null;
         }
 
+        public static bool HasAttribute<T>(this ApiParameterDescription apiParameter)
+            where T : Attribute =>
+            apiParameter.HasAttribute(typeof(T));
+
+        public static bool HasAttribute(this ApiParameterDescription apiParameter, Type attributeType)
```
martincostello, line 73: Missing null check.

src/Swashbuckle.AspNetCore.SwaggerGen/Annotations/SwaggerIgnoreAttribute.cs
```diff
@@ -0,0 +1,16 @@
...
+    public class SwaggerIgnoreAttribute : Attribute { }
```
martincostello, line 15: We can always unseal it if a use case for extension ever appears.

```suggestion
    public sealed class SwaggerIgnoreAttribute : Attribute { }
```

jcracknell:
You're correct that PropertyInfo.HasAttribute is public API and dropping it would be a breaking change, so I've dropped the commit. Unfortunately it's also going to be highly infectious thanks to Intellisense, so I'd suggest marking it as obsolete so people are aware, but I'll leave that up to you.

I've also dropped the commit adding extensions on Api{,Parameter}Description, as I had not realized that these are ApiExplorer types, and public extension methods are inappropriate here for the same reason.

martincostello (approved):

martincostello:
Thanks for the contribution - I'm not going to merge this immediately as I'm not sure if our next release is going to be a patch release of just small changes/tweaks/fixes, or a minor release with some new features too (like this).

Once we know the next release is going to be minor (or major), I'll merge this.

joesdu:
it's look like [this](https://github.com/joesdu/EasilyNET/blob/main/src/EasilyNET.WebCore.Swagger/SwaggerFilters/SwaggerPropertyIgnoreFilter.cs) ?

PR #2615, opened 2023-03-04
feat(pathbase): Add support for app.UsePathBase

underwoodcode:
This is a minor change to add support for PathBase.

In cases where the API needs to use app.UsePathBase("some/path"), getting Swagger UI to conform to this path with RoutePrefix was a little involved.  This change adds a PathBase property that should be set to the same value defined in app.UsePathBase().

From there, it checks the incoming request to see if path base is defined and the PathBase property is populated.  

If PathBase is populated, but the incoming request path does not contain PathBase, then Swagger UI will not be displayed.

If PathBase is populated and the incoming request path contains the same PathBase value, then Swagger UI will be displayed.

If PathBase is not populated and the incoming request does not contain a PathBase value, then the normal workflow will display the Swagger UI.


martincostello:
Thanks for contributing - if you'd like to continue with this pull request, please rebase against the default branch to pick up our new CI. Tests would also be required.

underwoodcode:
Thanks!  I’ll get it updated

PR #2635, opened 2023-04-20
Fixes #1661 - Make IDocumentFilter async compatible 

### LINKED ISSUE #1661: Make IDocumentFilter async compatible

schehlmj:
This is my take on the fix for this.  Rather than extending IDocumentFilter, I created a new IDocumentAsyncFilter that is configured separately.   I chosen to name IDocumentAsyncFilter to help make it clearer for developers.   It continues to use a method named "Apply" like other filters, and since they Type itself makes it clear that it is for Async, I did not name the method "ApplyAsync".  In cases where GetSwaggerAsync() is called, the IDocumentFilters will run first and then the IDocumentAsyncFilters.  For GetSwagger(), only the IDocumentFilters are run. If IDocumentAsyncFilters are configured, it will throw an exception stating to use GetSwaggerAsync() instead.

I had also considered just extending the IDocumentFilter to interface to have an async ApplyAsync() method with a default implementation, but that was not possible with the targeted runtimes.

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/ConfigureSwaggerGeneratorOptions.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/ConfigureSwaggerGeneratorOptions.cs
@@ -46,6 +46,9 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
             _swaggerGenOptions.DocumentFilterDescriptors.ForEach(
                 filterDescriptor => options.DocumentFilters.Add(CreateFilter<IDocumentFilter>(filterDescriptor)));
 
+            _swaggerGenOptions.DocumentAsyncFilterDescriptors.ForEach(
+                filterDescriptor => options.DocumentAsyncFilters.Add(CreateFilter<IDocumentAsyncFilter>(filterDescriptor)));
+
             if (!options.SwaggerDocs.Any())
             {
                 options.SwaggerDocs.Add("v1", new OpenApiInfo { Title = _hostingEnv.ApplicationName, Version = "1.0" });
@@ -69,6 +72,7 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
             target.ParameterFilters = new List<IParameterFilter>(source.ParameterFilters);
             target.OperationFilters = new List<IOperationFilter>(source.OperationFilters);
             target.DocumentFilters = new List<IDocumentFilter>(source.DocumentFilters);
+            target.DocumentAsyncFilters = new List<IDocumentAsyncFilter>(source.DocumentAsyncFilters);
         }
 
         private TFilter CreateFilter<TFilter>(FilterDescriptor filterDescriptor)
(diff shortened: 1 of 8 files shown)


schehlmj:
The check failing is:
Invoke-WebRequest "https://dot.net/v1/dotnet-install.ps1" -OutFile "./dotnet-install.ps1"
Invoke-WebRequest : The underlying connection was closed: An unexpected error occurred on a send.

martincostello:

mauve:
I rebased and added support for the other filters in this PR: #2938

martincostello:
I'm going to close this in favour of #2938. Thanks for the initial work on this @schehlmj.

Issue #2656, opened 2023-05-30
Add Server default with schema/host


timheuer:
The swagger.[json|yaml] should emit a `servers` object in the OpenAPI specification using the running server scheme/host by default.  This helps complete the default configuration with little overrides quickly.

It saves from writing the code of `AddServer()` when the default host of the API should be at least one of the servers.  Or give an opt-out way.

martincostello:
https://github.com/domaindrivendev/Swashbuckle.AspNetCore/pull/2663#issuecomment-2053575975

PR #2663, opened 2023-06-05
Fix for #2656 - Add default server

timheuer:
This adds host detection in the middleware to populate the default schema/host for the swagger to populate the default Servers instance


Havunen:
When the open api document is generated during build the request origin is not what is going to be the real host of the API.

martincostello:
Thanks for contributing - if you'd like to continue with this pull request, please rebase against the default branch to pick up our new CI. Would you also be able to add a test?

PR #2672, opened 2023-06-16
💚Move tests to net7 to fix build

schnerring:
- Test fixes: cherry-picked from #2644
- Add some preprocessor symbols for .NET 7 support


schnerring:
Since creating the PR, I haven't looked at this again. I guess it would be good timing to sync this with the upcoming .NET 8 changes? But I haven't checked if this is tracked anywhere.

martincostello:

From looking at the description only, some parts of this will have been addressed by #2775.

martincostello:
Thanks for the contribution. The TFM updates are covered by the changes in #2799 to add support for .NET 8, and the other changes look like they were covered with #2775.

PR #2697, opened 2023-08-25
Set Nullable=true for array of nullable type in generated open api (fix #2628)

### LINKED ISSUE #2628: Array of nullable type do not set Nullable=true in generated open api
As this test expose https://github.com/worming004/Swashbuckle.AspNetCore/commit/3506ff1ff614579af70ade83be4aff93d90bceb9

```
[Theory]
[InlineData(typeof(int?[]), "integer", "int32", true)]
public void GenerateSchema_GeneratesArraySchema_IfEnumerableType(
    Type type,
    string expectedItemsType,
    string expectedItemsFormat,
    bool expectedItemsNullable)
{
    var schema = Subject().GenerateSchema(type, new SchemaRepository());

    Assert.Equal("array", schema.Type);
    Assert.NotNull(schema.Items);
    Assert.Equal(expectedItemsType, schema.Items.Type);
    Assert.Equal(expectedItemsFormat, schema.Items.Format);
    Assert.Equal(expectedItemsNullable, schema.Items.Nullable); // test fail here
}
```

The array's item is not set to nullable. Item is expected to be nullable because input type is `int?`

vdekercd:
In answer to [issue #2628](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/2628), if the array's item type is nullable. the Item type will be nullable in generated open api


martincostello:

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/ISchemaGenerator.cs
```diff
@@ -12,6 +12,7 @@ public interface ISchemaGenerator
             SchemaRepository schemaRepository,
             MemberInfo memberInfo = null,
             ParameterInfo parameterInfo = null,
-            ApiParameterRouteInfo routeInfo = null);
+            ApiParameterRouteInfo routeInfo = null,
+            bool isEffectiveTypeNeeded = true);
```
martincostello, line 16: This is a binary-breaking change we cannot make outside of a major version (and then only if we deem it worthy of the break).

src/Swashbuckle.AspNetCore.Newtonsoft/SchemaGenerator/NewtonsoftDataContractResolver.cs
```diff
@@ -201,32 +201,50 @@ private List<DataProperty> GetDataPropertiesFor(JsonObjectContract jsonObjectCon
         private static readonly Dictionary<Type, Tuple<DataType, string>> PrimitiveTypesAndFormats = new()
```
martincostello, line 201: I wonder if there's a way to refactor this here and in the other resolve to avoid the duplication.

PR #2710, opened 2023-09-16
Configure non-nullable types as required

### LINKED ISSUE #2036: Configure non-nullable types as required
In my project we do not distinguish between `required` and `nullable: false`. If a type is non-nullable, it is also required. This is regardless if it is a value type or reference type.

It would be great if one could optionally configure Swashbuckle to set all non-nullable properties to required. E.g. like: `setupAction.MarkNonNullableTypesAsRequired();` This way, I could remove all my `[Required]` annotations on the non-nullable C# properties.

This is a subsequent request following  #1686. As _originally posted by @felschr in https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/1686#issuecomment-674074694_ , treating non-nullable as required is the default for [nullable contexts](https://docs.microsoft.com/en-us/aspnet/core/mvc/models/validation?view=aspnetcore-3.1#non-nullable-reference-types-and-required-attribute) in .NET validation.

blouflashdb:
Closes #2036

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenOptionsExtensions.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenOptionsExtensions.cs
@@ -310,6 +310,15 @@ namespace Microsoft.Extensions.DependencyInjection
         }
 
         /// <summary>
+        /// Enable detection of non nullable reference types to set Required flag accordingly on schema properties
+        /// </summary>
+        /// <param name="swaggerGenOptions"></param>
+        public static void MarkNonNullableReferenceTypesAsRequired(this SwaggerGenOptions swaggerGenOptions)
+        {
+            swaggerGenOptions.SchemaGeneratorOptions.MarkNonNullableReferenceTypesAsRequired = true;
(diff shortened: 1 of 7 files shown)


blouflashdb:
Seems like the build pipeline is broken but this should have nothing to do with my changes.

piskov:
@blouflashdb there’s one more thing to consider: default values

```
public string Prop { get; init; } = "default value";
```

It’s non-nullable, yet it shoudn’t be required as it has a default value.

See how [ASP validation is working](https://github.com/dotnet/aspnetcore/pull/38801) as a result of this [issue](https://github.com/dotnet/aspnetcore/issues/18403)

blouflashdb:
You are right, I should take a look at the implementation in AspNetCore to match the behavior in Swashbuckle.

kpko:
Hi, this feature would help a lot, especially with generating APIs from swagger documents :) Is there a way to help this move forward?

martincostello:

martincostello:
Contributes to #2793.

martincostello:
https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/2036#issuecomment-2054039124

Issue #2712, opened 2023-09-26
CLI is running all registered IHostedService


desjoerd:
Currently the CLI is running all registered Hosted Services (`IHostedService`). I see that it's needed to start the application because otherwise the `app.Use*`/`app.Map*` configuration is not run. Because it's starting the application (with a NoopServer) it is running all registered HostedServices.

Because the CLI is almost always run in the context of a build it would be nice to not run those registered hosted services. This would also solve instances of for example HangFire or other things which start an IHostedService.

I've got a fix which I would like to discuss. It's a bit dirty but it works in my tests.

0xced:
Related: [dotnet-getdocument issues in .NET7](https://github.com/dotnet/aspnetcore/issues/45025#issuecomment-1320526859) on the ASP.NET Core repository.

The workaround I describe there requires modifying the application code so your pull request is probably a much better approach!

I'm just a bit worried that this project looks currently inactive. Last commit and last comment from @domaindrivendev dates back to January, 11. In the meantime, issues and pull requests are piling on. 😕 I hope this is only temporary. 🤞

martincostello:
https://github.com/domaindrivendev/Swashbuckle.AspNetCore/pull/2713#issuecomment-2053583005

martincostello:
Thanks for your help with this - the fix is available in [Swashbuckle.AspNetCore 6.6.2](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/releases/tag/v6.6.2).

PR #2713, opened 2023-09-26
Do not run IHostedService implementations

desjoerd:
Currently the CLI is running all registered Hosted Services (`IHostedService`). I see that it's needed to start the application because otherwise the `app.Use*`/`app.Map*` configuration is not run. Because it's starting the application (with a NoopServer) it is running all registered HostedServices.

Because the CLI is almost always run in the context of a build it would be nice to not run those registered hosted services. This would also solve instances of for example HangFire or other things which start an IHostedService.

Please let me know if you want to have this change or if I need to modify something.

Related issue: #2712

### DIFF
--- a/Swashbuckle.AspNetCore.sln
+++ b/Swashbuckle.AspNetCore.sln
@@ -92,6 +92,8 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Swashbuckle.AspNetCore.Test
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "MinimalApp", "test\WebSites\MinimalApp\MinimalApp.csproj", "{3D0126CB-5439-483C-B2D5-4B4BE111D15C}"
 EndProject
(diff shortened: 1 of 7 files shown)


siriak (approved):
I had a similar situation where we have validation on config values and when during build in CI environment we would generate swagger.json, it'd crash because of missing config files. So, this PR wouldn't solve my problem, but if it works for other people in their cases, I still think it's worthy!

desjoerd:
Thanks for the review, I see the build is currently (still) failing which is blocking the merge. I think #2730 would solve that.

martincostello:

desjoerd:
> Thanks for contributing - if you'd like to continue with this pull request, please rebase against the default branch to pick up our new CI.

Does this mean you're going to pick up active development?

martincostello:
See https://github.com/domaindrivendev/Swashbuckle.AspNetCore/discussions/2778 🙂

desjoerd:
I will update my PR, also the implementation was a bit too eager (which I fixed in Nswag) so I will put that in as well. Probably somewhere in the coming week.

martincostello:
I've pulled this changes into #2880.

@desjoerd if you could clarify what the missing fix to stop it being "too eager" is, I can apply that change and then move the new PR out of draft.

desjoerd:
Awesome, I totally got swamped in work. I've added a review :).

PR #2720, opened 2023-10-03
Ignore methods on Controller without routes

whoaskedfrfr:


Havunen:
Are you sure this is what you wanted to do? all public method in controller types are also callable from the client side. They need to be annotated with NonActionAttribute to make them not publicly visible / callable: https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.mvc.nonactionattribute?view=aspnetcore-8.0

martincostello:
Please open a new pull request with an explanation of the use case as noted above including tests if you'd like to continue with this contribution.

PR #2724, opened 2023-10-08
Remove reflection-based JSON serialization from SwaggerUIMiddleware

### LINKED ISSUE #2550: `SwaggerUIMiddleware` is broken when used with .NET Native AOT
A .NET 7 app can be published as a native executable by setting `<PublishAot>true</PublishAot>` in the csproj.

I noticed in my test app that the Swagger UI middleware is broken when native AOT is enabled.

The problem is here:
https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/eede4717ebd80d3e31bd06a6cb47b6b387b18500/src/Swashbuckle.AspNetCore.SwaggerUI/SwaggerUIMiddleware.cs#L121-L131

`JsonSerializer.Serialize` internally attempts to generate dynamic code which isn't compatible with native AOT. It throws an error, and a 500 response is returned to the browser.

A fix could be to update these serializer calls to use System.Text.Json source gen - https://devblogs.microsoft.com/dotnet/try-the-new-system-text-json-source-generator/. There are other serializer usages in Swashbuckle that you should test out.

VMelnalksnis:
Fixes #2550

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerUI/SwaggerUIMiddleware.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerUI/SwaggerUIMiddleware.cs
@@ -1,5 +1,4 @@
-﻿using System;
-using System.Reflection;
+﻿using System.Reflection;
 using System.Threading.Tasks;
 using System.Text.RegularExpressions;
 using System.Text;
@@ -14,7 +13,6 @@ using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using Microsoft.Extensions.FileProviders;
 using Microsoft.AspNetCore.StaticFiles;
-using Microsoft.AspNetCore.Http.Extensions;
 using System.Linq;
 
 #if NETSTANDARD2_0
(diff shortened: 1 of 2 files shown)


HakamFostok:
Can I ask a question, I saw the bug and I reviewed the PR, I know this is not my business, but I just want to wonder, if possible, why this PR is not merged yet.
I faced this problem in my Application, that led me to here.
Thank you very much for your efforts.

TurcanStanislav:
Hi everyone. @VMelnalksnis, @bdukes can I ask whether you will have time soon to fix the issues with the workflow and merge this PR? It would be great if you would manage to do this because this will help a lot of people. Thanks!

VMelnalksnis:
I see that CI is failing because .NET 6 SDK is missing:
image
Looks like a fix for this was already provided in #2672, though I'm not sure what is the correct fix - changing the test target framework or installing the correct SDK. Either way, I can't merge PRs so someone else will have to review/merge/publish new version.

martincostello:
Thanks for contributing - if you'd like to continue with this pull request, please rebase against the default branch to pick up our new CI. We'd also appreciate tests for this scenario to prevent regressions.

VMelnalksnis:
Ok, I'll check out the current tests and see where I can add them.

martincostello:
> Ok, I'll check out the current tests and see where I can add them.

I was looking at this myself yesterday and came up with this that builds on top of #2799: https://github.com/domaindrivendev/Swashbuckle.AspNetCore/pull/2800/commits/880fc5fa26f3a69cdf56843c4fa04c2b83b33357

If you'd like to cherry-pick those changes I'm happy to close my PR in favour of this one if it's making basically the same updates.

VMelnalksnis:
I see that the PR already adds `net8.0` targets to tests, and I guess fixes all AOT/reflection based issues, not just in this one place. It would make sense to go with that PR, and I'm just happy to see the issue resolved.

PR #2727, opened 2023-10-13
Fix that XML comment examples do not show up if the type is string and the example contains quotation marks

### LINKED ISSUE #2088: XML example comment does not show if has quotation mark, escaping nor unicode substitute does not solve
### Q&A 
 - OS: Windows10
 - Browser: Chrome
 - Version: 89.0.4389.114 (Official Build) (64-bit) (cohort: Stable)
 - Method of installation: ASP.NET Core API Project with Swashbuckle Nuget pkg install
 - Swagger-UI version: Swashbuckle.AspNetCore.SwaggerUI (6.1.2)
 - Swagger/OpenAPI version: OpenAPI 3.0

### Content & configuration
1. Install ASP.NET Core (.net 5.0) API project template, install this swashbuckle nuget pkg.
2. Enable XML comments for documenting the API with Swagger. (this works for everything except this quotation marks issue, so I believe it's done correctly)
3. Add xml example tags to a data model class' string-type property that contains a quotation mark (I'm trying to use it as the annotation for "inches"). 
4. optional: Add other xml example tags to other model properties that don't have the problematic quotation marks, as a control.
5. run the project and check the swagger page's examples; while other parameters may have an example from your xml tags, the parameter with a quotation mark will just show the default "string" for text.

```csharp
namespace Catalog.API.Entities
{
	public class Product
	{
		/// <example>602d2149e773f2a3990b47f5</example>
		[BsonId]
		[BsonRepresentation(BsonType.ObjectId)]
		public string Id { get; set; }

		/// <example>Tablets</example>
		public string Category { get; set; }

/** here comes the problematic tag, with the quotes after '10.5' */
		/// <example>10.5" touchscreen tablet with keyboard and stylus</example>

		/// <example>Lorem ipsum ...</example>
		public string Description { get; set; }
/** more stuff */
	}
}
```
and here's what the "Example Value" looks like in the Swagger page:

```json
{
  "id": "602d2149e773f2a3990b47f5",
  "name": "Surface Go",
  "category": "Tablets",
  "summary": "string", // not the text from the example tag
  "description": "Lorem ipsum dolor sit amet, consectetur adipisicing elit. Ut, tenetur natus doloremque laborum quos iste ipsum rerum obcaecati impedit odit illo dolorum ab tempora nihil dicta earum fugiat. Temporibus, voluptatibus. Lorem ipsum dolor sit amet, consectetur adipisicing elit.",
  "imageFile": "product-1.png",
  "price": 490.99
}
```

### Solution attempts
I tried escaping the quote using: 
1. `&quot;` 
2. `U+0022`
3. and others listed here: https://unicode-table.com/en/0022/ 

None of those worked. I eventually settled for representing "inches" as "in" instead of using quotation mark.

#### Here's how it looks when using "in" instead of a quotation mark to indicate "inches"
image

#### Here's how it looks when I use a quotation mark to indicate "inches"
image

dldl-cmd:
Fix that XML comment examples do not show up if the type is string and the example contains quotation marks

Resolves #2088

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsExampleHelper.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsExampleHelper.cs
@@ -1,0 +1,27 @@
+﻿using System.Text.Json;
+using Microsoft.OpenApi.Any;
+using Microsoft.OpenApi.Models;
+
+namespace Swashbuckle.AspNetCore.SwaggerGen
+{
+    public static class XmlCommentsExampleHelper
+    {
+        public static IOpenApiAny Create(
+            SchemaRepository schemaRepository,
+            OpenApiSchema schema,
+            string exampleString)
+        {
+            var isStringType =
+                (schema.ResolveType(schemaRepository) == "string") &&
+                !exampleString.Equals("null");
+
+            var exampleAsJson = isStringType
+                    ? JsonSerializer.Serialize(exampleString)
+                    : exampleString;
+
+            var example = OpenApiAnyFactory.CreateFromJson(exampleAsJson);
+
+            return example;
+        }
+    }
+}
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsExampleHelper.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsExampleHelper.cs
@@ -4,7 +4,7 @@ using Microsoft.OpenApi.Models;
 
 namespace Swashbuckle.AspNetCore.SwaggerGen
 {
-    public static class XmlCommentsExampleHelper
+    internal static class XmlCommentsExampleHelper
     {
         public static IOpenApiAny Create(
             SchemaRepository schemaRepository,
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsExampleHelper.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsExampleHelper.cs
@@ -13,7 +13,7 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
         {
             var isStringType =
                 (schema?.ResolveType(schemaRepository) == "string") &&
-                !exampleString.Equals("null");
+                !string.Equals(exampleString, "null");
 
             var exampleAsJson = isStringType
                     ? JsonSerializer.Serialize(exampleString)
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsExampleHelper.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsExampleHelper.cs
@@ -12,7 +12,7 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
             string exampleString)
         {
             var isStringType =
-                (schema.ResolveType(schemaRepository) == "string") &&
+                (schema?.ResolveType(schemaRepository) == "string") &&
                 !exampleString.Equals("null");
 
             var exampleAsJson = isStringType
--- a/test/Swashbuckle.AspNetCore.SwaggerGen.Test/XmlComments/XmlCommentsExampleHelperTests.cs
+++ b/test/Swashbuckle.AspNetCore.SwaggerGen.Test/XmlComments/XmlCommentsExampleHelperTests.cs
@@ -11,7 +11,7 @@ namespace Swashbuckle.AspNetCore.SwaggerGen.Test
         [Fact]
         public void Create_BuildsOpenApiArrayJson__When_NotStringTypeAndDataIsArray()
         {
-            OpenApiSchema schema = new OpenApiSchema();
+            var schema = new OpenApiSchema();
 
             IOpenApiAny example = XmlCommentsExampleHelper.Create(
                 schemaRepository,
@@ -19,6 +19,8 @@ namespace Swashbuckle.AspNetCore.SwaggerGen.Test
                 "[\"one\",\"two\",\"three\"]");
 
 
+            Assert.NotNull(example);
+            Assert.IsType<OpenApiArray>(example);
             var output = (OpenApiArray)example;
             Assert.Equal(3, output.Count);
             Assert.Equal("one",   ((OpenApiString)output[0]).Value);
@@ -36,6 +38,8 @@ namespace Swashbuckle.AspNetCore.SwaggerGen.Test
             IOpenApiAny example = XmlCommentsExampleHelper.Create(
                 schemaRepository, schema, exampleString);
 
+            Assert.NotNull(example);
+            Assert.IsType<OpenApiString>(example);
             Assert.Equal(((OpenApiString)example).Value, exampleString);
         }
 
@@ -48,6 +52,8 @@ namespace Swashbuckle.AspNetCore.SwaggerGen.Test
             IOpenApiAny example = XmlCommentsExampleHelper.Create(
                 schemaRepository, schema, "null");
 
+            Assert.NotNull(example);
+            Assert.IsType<OpenApiNull>(example);
             Assert.Equal(AnyType.Null, ((OpenApiNull)example).AnyType);
         }
 
@@ -58,6 +64,8 @@ namespace Swashbuckle.AspNetCore.SwaggerGen.Test
 
             IOpenApiAny example = XmlCommentsExampleHelper.Create(schemaRepository, schema, "[]");
 
+            Assert.NotNull(example);
+            Assert.IsType<OpenApiArray>(example);
             Assert.Empty((OpenApiArray)example);
         }
     }
--- a/test/Swashbuckle.AspNetCore.SwaggerGen.Test/XmlComments/XmlCommentsExampleHelperTests.cs
+++ b/test/Swashbuckle.AspNetCore.SwaggerGen.Test/XmlComments/XmlCommentsExampleHelperTests.cs
@@ -50,5 +50,15 @@ namespace Swashbuckle.AspNetCore.SwaggerGen.Test
 
             Assert.Equal(AnyType.Null, ((OpenApiNull)example).AnyType);
         }
+
+        [Fact]
+        public void Create_AllowsSchemaToBeNull()
+        {
+            OpenApiSchema schema = null;
+
+            IOpenApiAny example = XmlCommentsExampleHelper.Create(schemaRepository, schema, "[]");
+
+            Assert.Empty((OpenApiArray)example);
+        }
     }
 }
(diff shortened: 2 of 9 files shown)


src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsExampleHelper.cs
```diff
@@ -0,0 +1,27 @@
...
+namespace Swashbuckle.AspNetCore.SwaggerGen
+{
+    public static class XmlCommentsExampleHelper
+    {
+        public static IOpenApiAny Create(
+            SchemaRepository schemaRepository,
+            OpenApiSchema schema,
+            string exampleString)
+        {
+            var isStringType =
```
bbarry, line 14: Should this check if schema is null before calling `.ResolveType()`?

I don't think it happens in the intended use but I have a version of `XmlCommentsOperationFilter` which handles response tags like:

```xml
<response code="400" contentType="application/problem+json" type="Library.ProblemDetails">
  <example>{ "message": "There was a problem", "statusCode": 400, errors: [] }</example>
  <description>Bad Request - data validation error</description>
</response>
```

and with my operation filter calling this method, my unit tests fail with a null ref on schema here when the type cannot be found.

dldl-cmd: @bbarry yes, it should check if `schema` is null before calling `.ResolveType()`. Thanks for finding that issue. I have updated the PR.

martincostello:

S-Tasker:
@dldl-cmd Does this happen to fix general issues with example parsing, such as backslashes in the string leading to the example being omitted completely?

S-Tasker:
@martincostello It looks like the expected CI builds didn't run here - they're awaiting approval.

src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsExampleHelper.cs
```diff
@@ -0,0 +1,27 @@
...
+                (schema?.ResolveType(schemaRepository) == "string") &&
+                !exampleString.Equals("null");
```
martincostello, line 16: The static `string.Equals()` method is robust to nulls that the instance method.

src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsExampleHelper.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System.Text.Json;
+using Microsoft.OpenApi.Any;
+using Microsoft.OpenApi.Models;
+
+namespace Swashbuckle.AspNetCore.SwaggerGen
+{
+    public static class XmlCommentsExampleHelper
```
martincostello, line 7: Make this internal, we don't need to expose new public APIs for this.

test/Swashbuckle.AspNetCore.SwaggerGen.Test/XmlComments/XmlCommentsExampleHelperTests.cs
```diff
@@ -0,0 +1,64 @@
...
+        public void Create_BuildsOpenApiArrayJson__When_NotStringTypeAndDataIsArray()
+        {
+            OpenApiSchema schema = new OpenApiSchema();
+
+            IOpenApiAny example = XmlCommentsExampleHelper.Create(
+                schemaRepository,
+                schema,
+                "[\"one\",\"two\",\"three\"]");
+
+
```
martincostello, line 21:
```suggestion
            var schema = new OpenApiSchema();

            IOpenApiAny example = XmlCommentsExampleHelper.Create(
                schemaRepository,
                schema,
                "[\"one\",\"two\",\"three\"]");

```

test/Swashbuckle.AspNetCore.SwaggerGen.Test/XmlComments/XmlCommentsExampleHelperTests.cs
```diff
@@ -0,0 +1,64 @@
...
+        }
+
+        [Fact]
+        public void Create_AllowsSchemaToBeNull()
+        {
+            OpenApiSchema schema = null;
+
+            IOpenApiAny example = XmlCommentsExampleHelper.Create(schemaRepository, schema, "[]");
+
+            Assert.Empty((OpenApiArray)example);
```
martincostello, line 61: Assert that `example` is an `OpenApiArray` first.

test/Swashbuckle.AspNetCore.SwaggerGen.Test/XmlComments/XmlCommentsExampleHelperTests.cs
```diff
@@ -0,0 +1,64 @@
...
+                schemaRepository,
+                schema,
+                "[\"one\",\"two\",\"three\"]");
+
+
+            var output = (OpenApiArray)example;
+            Assert.Equal(3, output.Count);
+            Assert.Equal("one",   ((OpenApiString)output[0]).Value);
+            Assert.Equal("two",   ((OpenApiString)output[1]).Value);
+            Assert.Equal("three", ((OpenApiString)output[2]).Value);
```
martincostello, line 26: Here and in the other methods, assert that the example isn't null as is of the right type, rather than doing casts without checking first.

dldl-cmd: done

test/Swashbuckle.AspNetCore.SwaggerGen.Test/XmlComments/XmlCommentsExampleHelperTests.cs
dldl-cmd: done

src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsExampleHelper.cs
dldl-cmd: done

test/Swashbuckle.AspNetCore.SwaggerGen.Test/XmlComments/XmlCommentsExampleHelperTests.cs
dldl-cmd: done

src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsExampleHelper.cs
dldl-cmd: done

test/Swashbuckle.AspNetCore.SwaggerGen.Test/XmlComments/XmlCommentsExampleHelperTests.cs
```diff
@@ -0,0 +1,72 @@
...
+
+            IOpenApiAny example = XmlCommentsExampleHelper.Create(
+                schemaRepository,
+                schema,
+                "[\"one\",\"two\",\"three\"]");
+
+
+            Assert.NotNull(example);
+            Assert.IsType<OpenApiArray>(example);
+            var output = (OpenApiArray)example;
```
martincostello, line 24:
```suggestion
            var output = Assert.IsType<OpenApiArray>(example);
```

test/Swashbuckle.AspNetCore.SwaggerGen.Test/XmlComments/XmlCommentsExampleHelperTests.cs
martincostello: This isn't done - the casts are still present.

The asserts should be something like this:

```csharp
var item = Assert.IsType<OpenApiArray>(output[0]);
Assert.Equal("one", item.Value);
```

test/Swashbuckle.AspNetCore.SwaggerGen.Test/XmlComments/XmlCommentsExampleHelperTests.cs
```diff
@@ -0,0 +1,72 @@
...
+            string exampleString = "example string with special characters\"<>\r\n\"";
+            OpenApiSchema schema = new OpenApiSchema { Type = "string" };
+            schemaRepository.AddDefinition("test", schema);
+            
+            IOpenApiAny example = XmlCommentsExampleHelper.Create(
+                schemaRepository, schema, exampleString);
+
+            Assert.NotNull(example);
+            Assert.IsType<OpenApiString>(example);
+            Assert.Equal(((OpenApiString)example).Value, exampleString);
```
martincostello, line 43:
```suggestion
            var actual = Assert.IsType<OpenApiString>(example);
            Assert.Equal(exampleString, actual.Value);
```

test/Swashbuckle.AspNetCore.SwaggerGen.Test/XmlComments/XmlCommentsExampleHelperTests.cs
```diff
@@ -0,0 +1,72 @@
...
+        {
+            OpenApiSchema schema = new OpenApiSchema { Type = "string" };
+            schemaRepository.AddDefinition("test", schema);
+
+            IOpenApiAny example = XmlCommentsExampleHelper.Create(
+                schemaRepository, schema, "null");
+
+            Assert.NotNull(example);
+            Assert.IsType<OpenApiNull>(example);
+            Assert.Equal(AnyType.Null, ((OpenApiNull)example).AnyType);
```
martincostello, line 57:
```suggestion
            var actual = Assert.IsType<OpenApiNull>(example);
            Assert.Equal(AnyType.Null, actual.AnyType);
```

test/Swashbuckle.AspNetCore.SwaggerGen.Test/XmlComments/XmlCommentsExampleHelperTests.cs
```diff
@@ -0,0 +1,72 @@
...
+        [Fact]
+        public void Create_AllowsSchemaToBeNull()
+        {
+            OpenApiSchema schema = null;
+
+            IOpenApiAny example = XmlCommentsExampleHelper.Create(schemaRepository, schema, "[]");
+
+            Assert.NotNull(example);
+            Assert.IsType<OpenApiArray>(example);
+            Assert.Empty((OpenApiArray)example);
```
martincostello, line 69:
```suggestion
            var actual = Assert.IsType<OpenApiArray>(example);
            Assert.Empty(actual);
```

test/Swashbuckle.AspNetCore.SwaggerGen.Test/XmlComments/XmlCommentsExampleHelperTests.cs
```diff
@@ -0,0 +1,72 @@
...
+        {
+            var schema = new OpenApiSchema();
+
+            IOpenApiAny example = XmlCommentsExampleHelper.Create(
+                schemaRepository,
+                schema,
+                "[\"one\",\"two\",\"three\"]");
+
+
+            Assert.NotNull(example);
```
martincostello, line 22:
```suggestion
                "[\"one\",\"two\",\"three\"]");

            Assert.NotNull(example);
```

test/Swashbuckle.AspNetCore.SwaggerGen.Test/XmlComments/XmlCommentsExampleHelperTests.cs
dldl-cmd: I hope, that it is now as expected

S-Tasker:
Is this something that mind find its way into v6.6 by any chance?

martincostello (approved):

Issue #1630, opened 2020-04-20
Timespan handling changed - no longer creates "date-span"


OliverRC:
In moving from 5.0.0-rc2 to 5.3.1 I noticed the following difference:
![image](https://user-images.githubusercontent.com/1888486/79754505-f5d3b400-8317-11ea-9875-c0ba524a40c9.png)

Looking at [JsonSerializerDataContractResolver.cs](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/master/src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/JsonSerializerDataContractResolver.cs) it seems like there is no handling for Timespan.

Was this specifically removed to no longer handle them as "date-span" anymore?

```
        private static readonly Dictionary<Type, Tuple<DataType, string>> PrimitiveTypesAndFormats = new Dictionary<Type, Tuple<DataType, string>>
        {
            [ typeof(bool) ] = Tuple.Create(DataType.Boolean, (string)null),
            [ typeof(byte) ] = Tuple.Create(DataType.Integer, "int32"),
            [ typeof(sbyte) ] = Tuple.Create(DataType.Integer, "int32"),
            [ typeof(short) ] = Tuple.Create(DataType.Integer, "int32"),
            [ typeof(ushort) ] = Tuple.Create(DataType.Integer, "int32"),
            [ typeof(int) ] = Tuple.Create(DataType.Integer, "int32"),
            [ typeof(uint) ] = Tuple.Create(DataType.Integer, "int32"),
            [ typeof(long) ] = Tuple.Create(DataType.Integer, "int64"),
            [ typeof(ulong) ] = Tuple.Create(DataType.Integer, "int64"),
            [ typeof(float) ] = Tuple.Create(DataType.Number, "float"),
            [ typeof(double) ] = Tuple.Create(DataType.Number, "double"),
            [ typeof(decimal) ] = Tuple.Create(DataType.Number, "double"),
            [ typeof(byte[]) ] = Tuple.Create(DataType.String, "byte"),
            [ typeof(string) ] = Tuple.Create(DataType.String, (string)null),
            [ typeof(char) ] = Tuple.Create(DataType.String, (string)null),
            [ typeof(DateTime) ] = Tuple.Create(DataType.String, "date-time"),
            [ typeof(DateTimeOffset) ] = Tuple.Create(DataType.String, "date-time"),
[...]

PR #2754, opened 2024-01-08
Add TimeSpan support for JsonSerializer

RuguKi:
TimeSpan is currently supported by NewtonsoftSerializer but not by JsonSerializer. 
This changes adds the missing support  which helps STJ users.


uTeisT:
As a note, this is probably by-design ( #1630 ) hoping that it'd be handled by STJ.  But `TimeSpan`s are still not resolved properly and shown as `"$ref": "#/definitions/System.TimeSpan"` in swagger.json.

`duration` may be a better choice of format as specified in https://spec.openapis.org/registry/format/ but I decided to keep it aligned with [NewtonsoftDataContractResolver.cs](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/8f363f7359cb1cb8fa5de5195ec6d97aefaa16b3/src/Swashbuckle.AspNetCore.Newtonsoft/SchemaGenerator/NewtonsoftDataContractResolver.cs)

RuguKi:
Abandoning this PR as this is probably already fixed in STJ but the packages have not been updated here.

PR #2762, opened 2024-01-30
.NET 8 targetframework

Havunen:


Havunen:
Continued here: https://github.com/Havunen/DotSwashbuckle

PR #2796, opened 2024-04-14
Fix configuration properties not being copied

### LINKED ISSUE #2741: SwaggerGeneratorOptions.SchemaComparer isn't calling the configured Comparer
I'm trying to custom sort my schemas and SwaggerGeneratorOptions.SchemaComparer doesn't seem to have any effect.
 
Could the problem be that SchemaComparer getting skipped in this DeepCopy?

https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/8f363f7359cb1cb8fa5de5195ec6d97aefaa16b3/src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/ConfigureSwaggerGeneratorOptions.cs#L55-L65

Using Swashbuckle.AspNetCore.SwaggerGen.6.5.0.0

martincostello:
Fix properties not being copied during clone.

I looked into testing the method directly, but that involved what I thought was too much complication to get `[InternalsVisibleTo]` working. Instead I've added a test which will fail if new properties are added to act as a signpost to update the `DeepCopy()` methods.

- Resolves #2540.
- Resolves #2741.


domaindrivendev (approved):
good call @martincostello and the fact we need to do this is bringing back memories of all the things I'd love to refactor in this lib lol

PR #2799, opened 2024-04-14
Add .NET 8 support

### LINKED ISSUE #2771: DateOnly not rendering correctly in Swagger
I _think_ this may be a regression of #2319, but in investigating [this stackoverflow question](https://stackoverflow.com/q/77834867/634824), I found that `DateOnly` fields are still not showing correctly in the Swagger UI in ASP.NET 8 latest.

To reproduce:
- Download and install .NET 8  SDK 8.0.2 (SDK 8.0.201)
- `dotnet new webapi`
- `dotnet run`
- Browse to Swagger endpoint, i.e. `http://localhost:5261/swagger`

Notice `DateOnly` and  `DayOfWeek` have been broken out into properties as if custom types, rather than being a `date`.

image

Apply the workaround:

```csharp
builder.Services.AddSwaggerGen(options => {
    options.MapType<DateOnly>(() => new OpenApiSchema { 
        Type = "string",
        Format = "date" });
});
```

Then it works:
image

This shouldn't be necessary though, right?

### LINKED ISSUE #2791: Add WeatherForecast as an integration test
https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/2771#issuecomment-2053952428

A lot of issues seem to be reproduced with ASP.NET Core's Weather Forecast app from the project templates.

We should add a copy of it (from .NET 8 most likely) as an integration test project and add some tests to verify its behaviour.

### LINKED ISSUE #2792: Support .NET 8
Tracking issue to add support for `net8.0`.

This issue does _not_ cover adding support for any new language features added as part of .NET 8 (or 7), such as `required` members (see #2555), new ASP.NET Core features etc. Those will be treated as separate feature requests.

### LINKED ISSUE #2794: Use Regex source generator
For `net7.0` and later we can look to leverage the [Regex source generator]() to improve performance.

For example:

https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/830c5a6f7468bce6b716d84c034a258dbbcbfe98/src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsTextHelper.cs#L10-L13

martincostello:
- Add support for .NET 8.
- Bump version to `6.6.0`.

Resolves #2792
Resolves #2791
Resolves #2771
Resolves #2794

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenServiceCollectionExtensions.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenServiceCollectionExtensions.cs
@@ -1,6 +1,5 @@
 ﻿using System;
 using System.Text.Json;
-using Microsoft.AspNetCore.Mvc;
 using Microsoft.Extensions.ApiDescriptions;
 using Microsoft.Extensions.DependencyInjection.Extensions;
 using Microsoft.Extensions.Options;
@@ -16,7 +15,7 @@ namespace Microsoft.Extensions.DependencyInjection
             Action<SwaggerGenOptions> setupAction = null)
         {
             // Add Mvc convention to ensure ApiExplorer is enabled for all actions
-            services.Configure<MvcOptions>(c =>
+            services.Configure<AspNetCore.Mvc.MvcOptions>(c =>
                 c.Conventions.Add(new SwaggerApplicationConvention()));
 
             // Register custom configurators that takes values from SwaggerGenOptions (i.e. high level config)
@@ -33,8 +32,16 @@ namespace Microsoft.Extensions.DependencyInjection
             services.TryAddTransient<ISerializerDataContractResolver>(s =>
             {
 #if (!NETSTANDARD2_0)
-                var serializerOptions = s.GetService<IOptions<JsonOptions>>()?.Value?.JsonSerializerOptions
+                var serializerOptions =
+                    s.GetService<IOptions<AspNetCore.Mvc.JsonOptions>>()?.Value?.JsonSerializerOptions
+#if NET8_0_OR_GREATER
+                    ?? s.GetService<IOptions<AspNetCore.Http.Json.JsonOptions>>()?.Value?.SerializerOptions
+#endif
+#if NET7_0_OR_GREATER
+                    ?? JsonSerializerOptions.Default;
+#else
                     ?? new JsonSerializerOptions();
+#endif
 #else
                 var serializerOptions = new JsonSerializerOptions();
 #endif
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenServiceCollectionExtensions.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenServiceCollectionExtensions.cs
@@ -29,23 +29,10 @@ namespace Microsoft.Extensions.DependencyInjection
             services.TryAddTransient(s => s.GetRequiredService<IOptions<SwaggerGeneratorOptions>>().Value);
             services.TryAddTransient<ISchemaGenerator, SchemaGenerator>();
             services.TryAddTransient(s => s.GetRequiredService<IOptions<SchemaGeneratorOptions>>().Value);
-            services.TryAddTransient<ISerializerDataContractResolver>(s =>
+            services.AddSingleton<JsonSerializerOptionsProvider>();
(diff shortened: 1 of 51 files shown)


test/Swashbuckle.AspNetCore.IntegrationTests/SwaggerIntegrationTests.cs
```diff
@@ -112,5 +111,66 @@ public async Task SwaggerEndpoint_ReturnsCorrectPriceExample_ForDifferentCulture
...
+            using var client = application.CreateDefaultClient();
+
+            using var swaggerResponse = await client.GetFromJsonAsync<JsonDocument>("/swagger/v1/swagger.json");
+
+            var weatherForecase = swaggerResponse.RootElement
+                .GetProperty("components")
+                .GetProperty("schemas")
+                .GetProperty("WeatherForecast");
+
+            Assert.Equal("object", weatherForecase.GetProperty("type").GetString());
```
Nysosis, line 137: Nunit has support for doing [multiple assertions](https://docs.nunit.org/articles/nunit/writing-tests/assertions/multiple-asserts.html) in a test, might be worth wrapping all these in it so that if more than one regression occurs at once, they are identified in a single execution?

martincostello: We can maybe improve that in the future, but as we use xunit we wouldn't improve the assertions by pulling in NUnit too. Using something like Shouldly or FluentAssertions to extend xunit's capabilities would be better IMHO.

Nysosis: Apologies, that was me being in a primarily nunit world, not thinking to check the library actually used here when checking the idea locally, but xunit also supports the same syntax.

martincostello: No problem - I'll take a look to see what there is. In my own projects I use Shouldly so I'm not used to using xunit's own built-in assertions.

martincostello: Thanks for the suggestion, I've refactored this to use `Assert.Multiple()`.

src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenServiceCollectionExtensions.cs
```diff
@@ -33,8 +32,16 @@ public static class SwaggerGenServiceCollectionExtensions
             services.TryAddTransient<ISerializerDataContractResolver>(s =>
             {
 #if (!NETSTANDARD2_0)
-                var serializerOptions = s.GetService<IOptions<JsonOptions>>()?.Value?.JsonSerializerOptions
+                var serializerOptions =
```
dnperfors, line 35: It might be a good idea to document this order of options somewhere, because not documenting it will most likely raise questions about what options are used.

src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenServiceCollectionExtensions.cs
```diff
@@ -33,8 +32,16 @@ public static class SwaggerGenServiceCollectionExtensions
...
+                    s.GetService<IOptions<AspNetCore.Mvc.JsonOptions>>()?.Value?.JsonSerializerOptions
+#if NET8_0_OR_GREATER
+                    ?? s.GetService<IOptions<AspNetCore.Http.Json.JsonOptions>>()?.Value?.SerializerOptions
+#endif
+#if NET7_0_OR_GREATER
+                    ?? JsonSerializerOptions.Default;
```
dnperfors, line 41: Would it be sensible / useful  to be able to pass in an instance of `JsonSerializerOptions`.

martincostello: Possibly.

dnperfors (approved):

test/WebSites/CustomDocumentSerializer/Controllers/WeatherForecastController.cs
mesMess, line 1: No logging separate logging for weather? :)

mesMess (approved):
Targeting necessary libraries in support of .NET 8
Dropping extraneous class objects (envs)
Method removal/cleanup

PR #2800, opened 2024-04-14
Support native AoT with SwaggerUI

### LINKED ISSUE #2550: `SwaggerUIMiddleware` is broken when used with .NET Native AOT

martincostello:
Add support for using SwaggerUI with native AoT.

SwaggerGen still does not support native AoT, though it happens to work for simple cases from testing the sample app locally.

Resolves #2550.

Depends on #2799.


mesMess (approved):
Targeting frameworks for AoT and .NET 8 support
Wrapped using contexts (using)
Method cleanup
Other changes

Issue #2500, opened 2022-09-13
swaggergen not using custom json converter


trampster:
We are using mongodb which has an ObjectId type (struct). We use a JsonConverter to serialize this as a string, This converter is registered using builder.Services.AddJsonOptions. It works perfectly in our actual API.

However when the swagger is generated the examples have it serialised as an object:
```
"id": {
        "timestamp": 0,
        "creation_time": "2022-09-13T22:36:23.872Z"
      }
```

Things we have tried which didn't work:
- using an example tag in the xml comments in the model
- using a ISchemaFilter to change the type and set the example

mrentier:
Ran into same problem, different use case. I found two spots I had to band-aid. First is that primitive types are not configurable. I was adding DateOnly, which would be considered a primitive. Following custom JsonSerializerDataContractResolver fixed that for me:
` internal class CustomJsonSerializerDataContractResolver : ISerializerDataContractResolver
    {
        private readonly JsonSerializerOptions _serializerOptions;
        private readonly JsonSerializerDataContractResolver _resolver;
        private static readonly Dictionary<Type, Tuple<DataType, string>> CustomPrimitiveTypesAndFormats = new Dictionary<Type, Tuple<DataType, string>>
        {            
             [typeof(DateOnly)] = Tuple.Create(DataType.String, "date"),
            [typeof(TimeOnly)] = Tuple.Create(DataType.String, "time"),           
        };

        public CustomJsonSerializerDataContractResolver(JsonSerializerOptions serializerOptions)
        {
            _serializerOptions = serializerOptions;
            _resolver = new JsonSerializerDataContractResolver(serializerOptions);
        }

        public DataContract GetDataContractForType(Type type)
        {
            if (type.IsOneOf(typeof(object), typeof(JsonDocument), typeof(JsonElement)))
            {
                return DataContract.ForDynamic(type, JsonConverterFunc);
            }

            if (CustomPrimitiveTypesAndFormats.ContainsKey(type))
            {
                Tuple<DataType, string> tuple = CustomPrimitiveTypesAndFormats[type];
                return DataContract.ForPrimitive(type, tuple.Item1, tuple.Item2, JsonConverterFunc);
            }          
            return _resolver.GetDataContractForType(type);
        }

        private string JsonConverterFunc(object value)
        {
            return JsonSerializer.Serialize(value, _serializerOptions);
        }

[...]

PR #2804, opened 2024-04-16
Extend built-in supported types

### LINKED ISSUE #2611: Support Int128 and UInt128 in schema generation
.NET 7 adds support for representing 128-bit integers via the Int128 and UInt128 types.

When provided as inputs for an API, the schema generator interprets the type as a string. When emitted as part of the response, the type is treated as an object.

We probably want to represent this type using the `{ "type": "integer" }` schema type.

See https://github.com/dotnet/aspnetcore/issues/43119 for more info.

martincostello:
- Ensure `DateOnly` and `TimeOnly` with for Newtonsoft.Json.
- Add System.Text.Json support for `TimeSpan` and `Version`.
- Add Int128 and UInt128 support.
- Handle `AnnotationsDataType.Upload`.
- Create `AnnotationsDataType` mapping dictionary only once.

Resolves #2611
Relates to #2500

### DIFF
--- a/src/Swashbuckle.AspNetCore.Newtonsoft/SchemaGenerator/NewtonsoftDataContractResolver.cs
+++ b/src/Swashbuckle.AspNetCore.Newtonsoft/SchemaGenerator/NewtonsoftDataContractResolver.cs
@@ -33,9 +33,10 @@ namespace Swashbuckle.AspNetCore.Newtonsoft
 
             if (jsonContract is JsonPrimitiveContract && !jsonContract.UnderlyingType.IsEnum)
             {
-                var primitiveTypeAndFormat = PrimitiveTypesAndFormats.TryGetValue(jsonContract.UnderlyingType, out var format)
-                    ? format
-                    : Tuple.Create(DataType.String, (string)null);
+                if (!PrimitiveTypesAndFormats.TryGetValue(jsonContract.UnderlyingType, out var primitiveTypeAndFormat))
+                {
+                    primitiveTypeAndFormat = Tuple.Create(DataType.String, (string)null);
+                }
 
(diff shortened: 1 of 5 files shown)


src/Swashbuckle.AspNetCore.Newtonsoft/SchemaGenerator/NewtonsoftDataContractResolver.cs
```diff
@@ -186,7 +197,7 @@ private IEnumerable<DataProperty> GetDataPropertiesFor(JsonObjectContract jsonOb
             return dataProperties;
         }
 
-        private static readonly Dictionary<Type, Tuple<DataType, string>> PrimitiveTypesAndFormats = new Dictionary<Type, Tuple<DataType, string>>
+        private static readonly Dictionary<Type, Tuple<DataType, string>> PrimitiveTypesAndFormats = new()
```
remcolam, line 200:
```suggestion
        private static readonly Dictionary<Type, Tuple<DataType, string>> PrimitiveTypesAndFormats = new
```

domaindrivendev (approved):

vierlijner (commented):
Please add a #if NET6_0_OR_GREATER ... #endif (snippet) around the test inline data for dateonly and timeonly in the testfiles, like the example below. I can't suggest this change anymore.
` #if NET6_0_OR_GREATER
       [InlineData(typeof(DateOnly), "string", "date")]
       [InlineData(typeof(TimeOnly), "string", "time")] 
#endif`

martincostello:
There was no need to do that in the `Swashbuckle.AspNetCore.Newtonsoft.Test` project because it already targets a minimum of .NET 6 so the condition would always have evaluated to true.

PR #2715, opened 2023-09-27
Change links to https in README

304NotModified:
Because we like privacy

PR #2808, opened 2024-04-16
Add NuGet package READMEs

martincostello:
- Add a simple common README file for all NuGet packages.
- Cherry-pick changes from #2715 to update README links from HTTP to HTTPS.

### DIFF
--- a/package-readme.md
+++ b/package-readme.md
@@ -1,0 +1,16 @@
+# Swashbuckle.AspNetCore
+
+[Swagger](https://swagger.io) tooling for APIs built with ASP.NET Core.
+
+Generate beautiful API documentation, including a UI to explore and test operations, directly from your routes, controllers and models.
+
+[![Build status](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/actions/workflows/build.yml/badge.svg?branch=master&event=push)](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/actions?query=workflow%3Abuild+branch%3Amaster+event%3Apush) [![Code coverage](https://codecov.io/gh/domaindrivendev/Swashbuckle.AspNetCore/branch/master/graph/badge.svg)](https://codecov.io/gh/domaindrivendev/Swashbuckle.AspNetCore)
+
+## Release notes
+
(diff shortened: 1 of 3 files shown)


package-readme.md
```diff
@@ -0,0 +1,16 @@
+# Swashbuckle.AspNetCore
+
+[Swagger](https://swagger.io) tooling for APIs built with ASP.NET Core.
+
+Generate beautiful API documentation, including a UI to explore and test operations, directly from your routes, controllers and models.
+
+[![Build status](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/actions/workflows/build.yml/badge.svg?branch=master&event=push)](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/actions?query=workflow%3Abuild+branch%3Amaster+event%3Apush) [![Code coverage](https://codecov.io/gh/domaindrivendev/Swashbuckle.AspNetCore/branch/master/graph/badge.svg)](https://codecov.io/gh/domaindrivendev/Swashbuckle.AspNetCore)
```
dnperfors, line 7: What is the added value to add Build status and Code coverage badges to the package readme of the master branch?
When I look for a package and see the readme, I don't need to know if the master branch failed to build. It could be interesting to know what the code coverage is for the specific package version, but it wouldn't influence my decision to use a package very much (except perhaps if the coverage is 0%...)

martincostello: Just because it's already in the README and it still renders correctly in NuGet's Content Security Policy. I've essentially kept the "top" of the repo README, but with everything their CSP policy forbids, such as the OpenSSF Scorecard badge.

martincostello: The coverage also wouldn't be "a" version. It would just be "now". Aligning that as burned-in content in the package is too much engineering effort.

dnperfors: > The coverage also wouldn't be "a" version. It would just be "now". Aligning that as burned-in content in the package is too much engineering effort.

Possibly it might be done using the /tags/{versionnumber} url, but indeed the effort doesn't really warrant that change.

Anyways, I just wanted to raise the question of the necessity for these badges in the package readme.

dnperfors (approved):

domaindrivendev (approved):

PR #2824, opened 2024-04-23
remove redundant any check in InferRequestContentTypes

SimonCropp:


SimonCropp:
side effect of avoid an extra enumeration over the SupportedRequestFormats

martincostello (approved):

PR #2838, opened 2024-04-25
Fix flaky tests by locking on the statup type

### LINKED ISSUE #2836: Fix flaky CI test on Windows
The tests on Windows are being a bit flaky due to some file copy conflicts during the build of the test projects.

Need to do _something_ to resolve that.

remcolam:
@martincostello  I think this will contribute to #2836

### DIFF
--- a/test/Swashbuckle.AspNetCore.IntegrationTests/SwaggerIntegrationTests.cs
+++ b/test/Swashbuckle.AspNetCore.IntegrationTests/SwaggerIntegrationTests.cs
@@ -31,45 +31,39 @@ namespace Swashbuckle.AspNetCore.IntegrationTests
             string swaggerRequestUri)
         {
             var testSite = new TestSite(startupType);
-            using (var client = testSite.BuildClient())
-            {
+            using var client = testSite.BuildClient();
 
-                var swaggerResponse = await client.GetAsync(swaggerRequestUri);
+            using var swaggerResponse = await client.GetAsync(swaggerRequestUri);
 
-                swaggerResponse.EnsureSuccessStatusCode();
-                var contentStream = await swaggerResponse.Content.ReadAsStreamAsync();
-                new OpenApiStreamReader().Read(contentStream, out OpenApiDiagnostic diagnostic);
-                Assert.Empty(diagnostic.Errors);
-            }
+            swaggerResponse.EnsureSuccessStatusCode();
+            using var contentStream = await swaggerResponse.Content.ReadAsStreamAsync();
+            new OpenApiStreamReader().Read(contentStream, out OpenApiDiagnostic diagnostic);
+            Assert.Empty(diagnostic.Errors);
         }
 
         [Fact]
         public async Task SwaggerEndpoint_ReturnsNotFound_IfUnknownSwaggerDocument()
         {
             var testSite = new TestSite(typeof(Basic.Startup));
-            using (var client = testSite.BuildClient())
-            {
+            using var client = testSite.BuildClient();
 
-                var swaggerResponse = await client.GetAsync("/swagger/v2/swagger.json");
+            using var swaggerResponse = await client.GetAsync("/swagger/v2/swagger.json");
 
-                Assert.Equal(System.Net.HttpStatusCode.NotFound, swaggerResponse.StatusCode);
-            }
+            Assert.Equal(System.Net.HttpStatusCode.NotFound, swaggerResponse.StatusCode);
         }
 
         [Fact]
         public async Task SwaggerEndpoint_DoesNotReturnByteOrderMark()
         {
             var testSite = new TestSite(typeof(Basic.Startup));
-            using (var client = testSite.BuildClient())
-            {
+            using var client = testSite.BuildClient();
 
-                var swaggerResponse = await client.GetAsync("/swagger/v1/swagger.json");
+            using var swaggerResponse = await client.GetAsync("/swagger/v1/swagger.json");
 
-                swaggerResponse.EnsureSuccessStatusCode();
-                var contentBytes = await swaggerResponse.Content.ReadAsByteArrayAsync();
-                var bomBytes = Encoding.UTF8.GetPreamble();
-                Assert.NotEqual(bomBytes, contentBytes.Take(bomBytes.Length));
-            }
+            swaggerResponse.EnsureSuccessStatusCode();
+            var contentBytes = await swaggerResponse.Content.ReadAsByteArrayAsync();
+            var bomBytes = Encoding.UTF8.GetPreamble();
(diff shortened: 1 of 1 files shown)


test/Swashbuckle.AspNetCore.IntegrationTests/SwaggerIntegrationTests.cs
```diff
@@ -30,8 +30,14 @@ public class SwaggerIntegrationTests
             Type startupType,
             string swaggerRequestUri)
         {
+            System.Net.Http.HttpClient client;
             var testSite = new TestSite(startupType);
-            var client = testSite.BuildClient();
+
+            // Make sure not to build the same client in parallel
+            lock(startupType)
```
martincostello, line 37:
```suggestion
            lock (startupType)
```

martincostello (approved):

martincostello:
Thanks - I'll run the build a few times and if it passes every time I'll merge it 🙇‍♂️

remcolam:
@martincostello Actually, thinking about it, I think the lock needs to be larger, encapsulating the whole test. Because while the client is running, you are not able to compile it.

martincostello:
There's also a failure after the tests end where there's a write conflict on the coverage report:

```console
2024-04-25T10:44:39: Error during rendering summary report (Report type: 'Html'): The process cannot access the file 'D:\a\Swashbuckle.AspNetCore\Swashbuckle.AspNetCore\artifacts\coverage\index.html' because it is being used by another process.
  Basic -> D:\a\Swashbuckle.AspNetCore\Swashbuckle.AspNetCore\artifacts\bin\Basic\release_netcoreapp3.1\Basic.dll
  Out-File: The process cannot access the file 'D:\a\_temp\_runner_file_commands\step_summary_23f9f5a4-591b-4ba4-ac37-76846145df0c' because it is being used by another process.
D:\a\Swashbuckle.AspNetCore\Swashbuckle.AspNetCore\Directory.Build.targets(25,5): error MSB3073: The command "pwsh -Command "('<details><summary>:chart_with_upwards_trend: <b>Swashbuckle.AspNetCore.Annotations.Test Code Coverage report</b></summary>' + [System.Environment]::NewLine + [System.Environment]::NewLine + (Get-Content D:\a\Swashbuckle.AspNetCore\Swashbuckle.AspNetCore\artifacts\coverage\SummaryGithub.md | Out-String) + [System.Environment]::NewLine + [System.Environment]::NewLine + '</details>') >> D:\a\_temp\_runner_file_commands\step_summary_23f9f5a4-591b-4ba4-ac37-76846145df0c"" exited with code 1. [D:\a\Swashbuckle.AspNetCore\Swashbuckle.AspNetCore\test\Swashbuckle.AspNetCore.Annotations.Test\Swashbuckle.AspNetCore.Annotations.Test.csproj]
  Swashbuckle.AspNetCore.SwaggerGen.Test -> D:\a\Swashbuckle.AspNetCore\Swashbuckle.AspNetCore\artifacts\bin\Swashbuckle.AspNetCore.SwaggerGen.Test\release_net6.0\Swashbuckle.AspNetCore.SwaggerGen.Test.dll
```

remcolam:
I already realized it won't work... the lock is not long enough and putting the whole method in it will not allow await

remcolam:
This reworked version has run 50 times in parallel for me (however, I was only running the integratin tests, not other tests)

martincostello:
> whole method in it will not allow await

Await is possible within `lock` in recent versions of C# IIRC.

remcolam:
Noth that I could find, MS documentation also states it is not possible: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/statements/lock?devlangs=csharp&f1url=%3FappId%3DDev17IDEF1%26l%3DEN-US%26k%3Dk(lock_CSharpKeyword)%3Bk(SolutionItemsProject)%3Bk(DevLang-csharp)%26rd%3Dtrue

martincostello:
Maybe I've confused it with await support for `finally`

test/Swashbuckle.AspNetCore.IntegrationTests/SwaggerIntegrationTests.cs
```diff
@@ -26,25 +28,31 @@ public class SwaggerIntegrationTests
...
+                    var contentStream = swaggerResponse.Content.ReadAsStream();
+                    new OpenApiStreamReader().Read(contentStream, out OpenApiDiagnostic diagnostic);
+                    Assert.Empty(diagnostic.Errors);
+                }
             }
 
-            var swaggerResponse = await client.GetAsync(swaggerRequestUri);
+            HttpResponseMessage GetSwaggerResponse(HttpClient client)
+            {
+                return client.GetAsync(swaggerRequestUri).Result;
```
martincostello, line 53: This could potentially create deadlocks.

Making this class implement [IAsyncLifetime](https://www.danclarke.com/cleaner-tests-with-iasynclifetime) with a class fixture might be a way to create a serialized way to do this work ahead of the individual tests running.

remcolam:
@martincostello I was reading up on xunit, tests inside a class are run sequentially, so I think the issue is actually that the client wasn't being disposed, so depending on the garbage collector running you would run into the issue. I added a using to all the methods

martincostello:
Ah cool - I had noticed that myself in a few tests when doing work on them, but I hadn't gone through all the specific usages to ensure everything is disposed of.

test/Swashbuckle.AspNetCore.IntegrationTests/SwaggerIntegrationTests.cs
```diff
@@ -31,39 +31,45 @@ public class SwaggerIntegrationTests
             string swaggerRequestUri)
         {
             var testSite = new TestSite(startupType);
-            var client = testSite.BuildClient();
+            using (var client = testSite.BuildClient())
```
martincostello, line 34: Can you change these all to use `using var client = testSite.BuildClient();` instead please?

It's newer syntax that avoids all these indentation changes where the scope ends at the end of the method anyway, so it's the same.

At the same time, all the responses can also be made `using var` as `HttpResponseMessage` implements `IDisposable` too.

remcolam:
Hopefully this now fixes the locking issues... I haven't run into it, hopefully now it solves it for you as well

martincostello:
I'll run the CI a few times and see how it goes. I've updated the PR description to not close the associated issue as I don't think this change will fix the case where it fails due to conflicts when writing the code coverage report(s): https://github.com/domaindrivendev/Swashbuckle.AspNetCore/pull/2838#issuecomment-2076893700

remcolam:
Happy to help

remcolam:
I think you could call me doing this in dutch Woggen (tranlates to english as Wabbing). Making a verb from the acronym for Werk Ontwijkend Gedrag/Work Avoiding Behavior)

Issue #2386, opened 2022-03-24
Content-type specified in `ProducesResponseTypeAttribute` is not in generated openapi document


magiino:
I'm using:
- .NET 6
- Swahbuckle.AspNetCore 6.3.0

I wonder how can I add to open api information about `content-type` which is different for each response.
- response  200 - content-type: application/zip
- response 500 - content-type: application/json

### I have prepared example:
1. See following action in controller and its open api document. `content-type` is documented with `ProducesResponseTypeAttribute`.

```
/// <summary>
/// Gets zip file.
/// </summary>
/// <response code="200">Zip file.</response>
/// <response code="500">Detailed internal server error.</response>
[ProducesResponseType(typeof(FileContentResult), StatusCodes.Status200OK, MediaTypeNames.Application.Zip)]
[ProducesResponseType(typeof(WebApiError), StatusCodes.Status500InternalServerError, MediaTypeNames.Application.Json)]
[HttpGet("zipfile")]
public ActionResult GetZipFile()
{
    try
    {
        var (zip, fileName) = service.GetZip();
        return File(zip, MediaTypeNames.Application.Zip, fileName);
    }
    catch (Exception ex)
    {
        return StatusCode(500, new WebApiError("test", "Test"));
    }
}
```

```
{
  "openapi": "3.0.1",
  "info": {
    "title": "Test Web API",
    "description": "",
    "version": "0.1.0"
  },
  "paths": {
    "/diagnostic/zipfile": {
      "get": {
        "tags": [
          "Diagnostic"
        ],
        "summary": "Gets zip file.",
        "operationId": "GetZipFile",
        "responses": {
          "200": {
            "description": "Zip file."
          },
          "500": {
            "description": "Detailed internal server error."
          }
        }
      }
    },
```
As you can see there are missing response details for 200 and 500 code in open api document.
Is this bug, can someone explain why is this not working?

2. If I add `ProducesAttribute` the responses are documented, but both repsonses have both `content-type`s (which is wrong).

```
/// <summary>
/// Gets zip file.
/// </summary>
/// <response code="200">Zip file.</response>
/// <response code="500">Detailed internal server error.</response>
[ProducesResponseType(typeof(FileContentResult), StatusCodes.Status200OK)]
[ProducesResponseType(typeof(WebApiError), StatusCodes.Status500InternalServerError)]
[Produces(MediaTypeNames.Application.Zip, MediaTypeNames.Application.Json)]
[...]

PR #2841, opened 2024-04-26
Fix handling of FileResult's with content types

### LINKED ISSUE #2320: Support content type from ProducesResponseTypeAttribute
Using version 6.2.3

I have been trying to set the response type of one of my endpoints to "application/pdf".
It works, sort of, when using the ProducesAttribute, like in the documentation, but adding another response type like 400 = application/json is not easy.

I would like to be able to use ProducesResponseTypeAttribute instead and it looks as if it follows the features as SwaggerResponseAttribute.

However, it seems the content-type from ProducesResponseTypeAttribute is not used in the schema:

```csharp
// Works! content-type set to application/pdf
[SwaggerResponse(200, null, typeof(FileResult), "application/pdf")]
// Doesnt work content-type not set to application/pdf
[ProducesResponseType(typeof(FileResult), StatusCodes.Status200OK, "application/pdf")]
```

IGx89:
When ProducesResponseType is used with FileResult subclasses and explicit content types, ApiExplorer sets ModelMetadata to null. We were then erroneously treating it as returning no content.

Resolves #2320, #2386

### DIFF
--- a/test/Swashbuckle.AspNetCore.SwaggerGen.Test/SwaggerGenerator/SwaggerGeneratorTests.cs
+++ b/test/Swashbuckle.AspNetCore.SwaggerGen.Test/SwaggerGenerator/SwaggerGeneratorTests.cs
@@ -5,6 +5,7 @@ using System.Text.Json;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Authentication;
 using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.Mvc.Abstractions;
 using Microsoft.AspNetCore.Mvc.ApiExplorer;
 using Microsoft.AspNetCore.Mvc.ModelBinding;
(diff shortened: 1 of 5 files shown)


test/WebSites/Basic/Controllers/FilesController.cs
```diff
@@ -27,8 +27,12 @@ public IActionResult PostFormWithFile([FromForm]FormWithFile formWithFile)
         }
 
         [HttpGet("{name}")]
-        [Produces("application/octet-stream", Type = typeof(FileResult))]
```
martincostello, line 30: Do we still want a test for `FileResult` though?

test/Swashbuckle.AspNetCore.SwaggerGen.Test/SwaggerGenerator/SwaggerGeneratorTests.cs
```diff
@@ -854,6 +855,40 @@ public void GetSwagger_GeneratesResponses_ForSupportedResponseTypes()
...
+                {
+                    new ApiResponseType
+                    {
+                        ApiResponseFormats = new [] { new ApiResponseFormat { MediaType = "application/zip" } },
+                        StatusCode = 200,
+                        Type = typeof(FileContentResult)
+                    }
+                });
+
+            // ASP.NET Core sets ModelMetadata to null for FileResult's
```
martincostello, line 876:
```suggestion
            // ASP.NET Core sets ModelMetadata to null for FileResults
```

test/WebSites/Basic/Controllers/FilesController.cs
IGx89: Good question. I removed it because UseOneOfForPolymorphism is enabled and using the FileResult base type consequently produces a oneOf with four identical schemas. Testing polymorphism didn't seem the goal of this particular method and its related test cases, and the result could be confusing for others running this, so I switched it to a specific subclass instead. I'm open to switching it back, but it'll muddy things a bit and make my asserts more complicated. Just LMK what you'd prefer!

martincostello: I think it's more that people might still use the patent class like mentioned in the original issue, and I think it's better we have coverage for that as well, rather than change the existing test to be more specific to your changes.

IGx89: Done!

test/Swashbuckle.AspNetCore.SwaggerGen.Test/SwaggerGenerator/SwaggerGeneratorTests.cs
```diff
@@ -5,6 +5,7 @@
 using System.Threading.Tasks;
 using System.Reflection;
 using Microsoft.AspNetCore.Authentication;
+using Microsoft.AspNetCore.Http;
```
martincostello, line 8: The duplicate on line 14 needs removing.

IGx89: 🤦 fixed

martincostello (approved):

PR #2868, opened 2024-05-12
Bump some NuGet packages

martincostello:
- Bump xunit to 2.8.0.
- Bump to ReportGenerator 5.2.5.
- Bump Microsoft.AspNetCore.Mvc.Testing and Microsoft.AspNetCore.TestHost to their latest versions for .NET 6 and 8.
- Fix new xunit analyser warnings and two IDE suggestions in the same file.
- Remove unused test project for .NET Core 2.1.

Updates applied manually after I found that Dependabot is failing to update our NuGet packages: https://github.com/dependabot/dependabot-core/issues/8578#issuecomment-2106151653


mesMess (approved):
NuGet package bump
Associated hashes bumped
Swap for TASK (Top-level async return for void)
Project files updated
Extraneous files deleted
Build configurations cleaned up

PR #2870, opened 2024-05-13
Only attest packages

martincostello:
Turns out there's a [64 file limit on attestations](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/actions/runs/9058739831/job/24884934967#step:9:66), and the binary/TFM combinations go beyond that.


mesMess (approved):
Consolidated artifacts directory (upper limit of 64)

mesMess:
@martincostello it would be nice if we could drop the error/notification on the upper limit for attestations here :)

martincostello:
It's linked to in the PR description.

mesMess:
Yep, I see it, thanks!

PR #2872, opened 2024-05-13
Add descriptions for more HTTP status codes

### LINKED ISSUE #2871: [Feature request]: Add descriptions for more HTTP status codes

I've noticed that the HTTP status code 422 has the description "Client Error" by default. I'd expect it to be "Unprocessable Entity".
Seems like "Client Error" is the default description for all 4xx HTTP status codes that don't have a better description.

Please update the default HTTP status code description list. A good source seems to be https://github.com/dotnet/aspnetcore/blob/main/src/Http/WebUtilities/src/ReasonPhrases.cs

### Additional context

_No response_

cremor:
# Pull Request

Fixes #2871

As suggested in the issue, I've mostly added the status codes missing when compared with https://www.rfc-editor.org/rfc/rfc9110.html#name-status-codes

The following status codes are not in that RFC but I've added them because I think they make sense. You can read details about them here: https://httpstatuses.io

423 Locked
424 Failed Dependency
428 Precondition Required
429 Too Many Requests (was already there)
431 Request Header Fields Too Large
451 Unavailable For Legal Reasons

I hope the readme changes for 200 OK and 500 Internal Server Error are ok. If not, please tell if I should only revert the readme changes, or if I should revert the whole 200 OK and 500 Internal Server Error lines (so that they fall back to the generic description like before).


src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
```diff
@@ -667,31 +667,64 @@ private static bool IsFromFormAttributeUsedWithIFormFile(ApiParameterDescription
...
             new KeyValuePair<string, string>("404", "Not Found"),
             new KeyValuePair<string, string>("405", "Method Not Allowed"),
             new KeyValuePair<string, string>("406", "Not Acceptable"),
+            new KeyValuePair<string, string>("407", "Proxy Authentication Required"),
             new KeyValuePair<string, string>("408", "Request Timeout"),
             new KeyValuePair<string, string>("409", "Conflict"),
+            new KeyValuePair<string, string>("410", "Gone"),
+            new KeyValuePair<string, string>("411", "Length Required"),
+            new KeyValuePair<string, string>("412", "Precondition Failed"),
+            new KeyValuePair<string, string>("413", "Payload Too Large"),
```
martincostello, line 706:
```suggestion
            new KeyValuePair<string, string>("413", "Content Too Large"),
```

martincostello (approved):

martincostello:
Looks like something in the tests is broken or needs updating.

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
cremor: Thanks, I missed that in RFC9110. Should we do the same for 422 Unprocessable Content? (I've added it as Unprocessable Entity.)

martincostello: Good spot - I missed that one.

cremor:
Tests should be fixed.

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
cremor: Fixed that too.

Issue #2877, opened 2024-05-14
[Bug]: Required and nullable properties are not nullable in the OpenAPI JSON


cremor:
### Describe the bug

Properties with the `required` modifier are always generated as not nullable (and in the case of strings also with `minLength: 1`), even if they are nullable.

### Expected behavior

Nullable properties should be nullable in the OpenAPI JSON.

Example:
![grafik](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/assets/116314/13893894-cad9-487c-a367-c4d30836d67a)

### Actual behavior

Nullable properties are always not-nullable in the OpenAPI JSON if they are also `required`.

Example:
![grafik](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/assets/116314/4731daa1-cd5c-4e6c-8f4f-5c4b407bfede)

### Steps to reproduce

1. Create a new .NET 8 ASP.NET Core Web API project.
2. Change the `WeatherForecast` model like shown below.
3. Change the controller GET method to `return null;` so the project compiles.
4. Check the generated OpenAPI JSON or Swagger UI.

```cs
    public class WeatherForecast
    {
        public string? StringNullable { get; set; }
        public string StringNotNullable { get; set; }
        public required string? StringRequiredNullable { get; set; }
        public required string StringRequiredNotNullable { get; set; }
        public int? IntNullable { get; set; }
        public int IntNotNullable { get; set; }
        public required int? IntRequiredNullable { get; set; }
        public required int IntRequiredNotNullable { get; set; }
        public Other? OtherNullable { get; set; }
        public Other OtherNotNullable { get; set; }
        public required Other? OtherRequiredNullable { get; set; }
        public required Other OtherRequiredNotNullable { get; set; }
    }

    public class Other
    {
    }
```

### Swashbuckle.AspNetCore version

6.6.1

### .NET Version

8.0.204

### Anything else?

This is a new problem in Swashbuckle 6.6.1 (6.6.0). In 6.5.0 the `required` keyword didn't have any effect.
I assume this was caused by @keahpeters in #2810

`required` doesn't mean that a not-null value is required. It just means that any value (which can be `null`) is required. In other words: It just means that the request body needs to contain the property.

Related to #2036, but that issue has so many comments already that I'm not sure if it requests something more than I do here.
Should be added to #2793

cremor:
[...]

PR #2879, opened 2024-05-14
Fix to make required and nullable properties nullable in schema

keahpeters:
Resolves https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/2877

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/SchemaGenerator.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/SchemaGenerator.cs
@@ -68,24 +68,13 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
                 {
                     var requiredAttribute = customAttributes.OfType<RequiredAttribute>().FirstOrDefault();
 
+                    schema.Nullable = _generatorOptions.SupportNonNullableReferenceTypes
+                        ? dataProperty.IsNullable && requiredAttribute == null && !memberInfo.IsNonNullableReferenceType()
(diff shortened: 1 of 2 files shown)


test/Swashbuckle.AspNetCore.SwaggerGen.Test/SchemaGenerator/JsonSerializerSchemaGeneratorTests.cs
```diff
@@ -367,9 +367,10 @@ public void GenerateSchema_SetsReadOnlyAndWriteOnlyFlags_IfPropertyIsRestricted(
         }
 
 #if NET7_0_OR_GREATER
-        public class TypeWithRequiredProperty
+        public class TypeWithRequiredProperties
```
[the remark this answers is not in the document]
keahpeters: Done.

src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/SchemaGenerator.cs
martincostello, line 1: Is this change just reverting back completely to the old behaviour from 6.5.0 and undoing the support for `required`? Or am I misinterpreting the diff?

keahpeters: It's only reverting the way we set `Nullable` and `MinLength`. This didn't need to be changed since the property can still be null or an empty string with the required `keyword`. The `required` keyword will still be supported with this change.

martincostello (approved):

PR #2880, opened 2024-05-14
Do not run IHostedService implementations

### LINKED ISSUE #2712: CLI is running all registered IHostedService

martincostello:
- Remove `IHostedService` implementations so the CLI doesn't run hosted services (and then hang/fail).
- Refactor CLI tests to reduce duplication.
- Use newer C# syntax as suggested by Visual Studio.

Supersedes #2713.

Resolves #2712.

/cc @desjoerd

### DIFF
--- a/src/Swashbuckle.AspNetCore.Cli/HostingApplication.cs
+++ b/src/Swashbuckle.AspNetCore.Cli/HostingApplication.cs
@@ -4,7 +4,9 @@ using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Hosting.Server;
 using Microsoft.AspNetCore.Http.Features;
+#if NETCOREAPP3_0_OR_GREATER
 using Microsoft.Extensions.DependencyInjection;
+#endif
 using Microsoft.Extensions.Hosting;
 
 namespace Swashbuckle.AspNetCore.Cli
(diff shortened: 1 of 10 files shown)


src/Swashbuckle.AspNetCore.Cli/HostingApplication.cs
```diff
@@ -31,6 +33,18 @@ void ConfigureHostBuilder(object hostBuilder)
                 {
                     services.AddSingleton<IServer, NoopServer>();
                     services.AddSingleton<IHostLifetime, NoopHostLifetime>();
+
+                    for (var i = services.Count - 1; i >= 0; i--)
+                    {
+                        // exclude all implementations of IHostedService
+                        // except Microsoft.AspNetCore.Hosting.GenericWebHostService because that one will build/configure
+                        // the WebApplication/Middleware pipeline in the case of the GenericWebHostBuilder.
+                        if (typeof(IHostedService).IsAssignableFrom(services[i].ServiceType)
```
desjoerd, line 42:
```suggestion
                        if (services[i].ServiceType == typeof(IHostedService)
```

This is taken from https://github.com/RicoSuter/NSwag/pull/4756

desjoerd (changes requested):
I've added the suggestion of the fix I wanted to port :)

src/Swashbuckle.AspNetCore.Cli/HostingApplication.cs
martincostello: Thanks! Do you have a specific example of a service that shouldn't be removed so I can augment the test?

mesMess (approved):
Stop running Hosted Service interfaces (some) to reduce hanging
Syntax cleanup and changes (Visual Studio Syntax Guide)

PR #2882, opened 2024-05-14
Add support for the `[Length]`attribute.

### LINKED ISSUE #2756: [Length] Attribute isn't supported
This issue is a feature request for a support of the `[Length]` attribute. Judging by the example in the [Contribution guide](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/8f363f7359cb1cb8fa5de5195ec6d97aefaa16b3/CONTRIBUTING.md?plain=1#L16), this issue falls into the **Feature Requests** category.

The `System.ComponentModel.DataAnnotations` namespace has 3 (actually even more, but it is not within the scope of this Issue) Attributes responsible for the string length validation:
1. `[MinLength(length)]` - [MS Docs](https://learn.microsoft.com/en-us/dotnet/api/system.componentmodel.dataannotations.minlengthattribute)
2. `[MaxLength(length)]` - [MS Docs](https://learn.microsoft.com/en-us/dotnet/api/system.componentmodel.dataannotations.maxlengthattribute)
3. `[Length(minimumLength, maximumLength)]` - [MS Docs](https://learn.microsoft.com/en-us/dotnet/api/system.componentmodel.dataannotations.lengthattribute)

And while the first two are supported, the latter isn't.
Suppose, we have the following DTO definition:
```csharp
class Dto
{
    [MinLength(2), MaxLength(10)]
    public string Prop1 { get; set; }

    [Length(2, 10)]
    public string Prop2 { get; set; }
}
```
Having such model SwaggerGen produces the following `OpenApi` specification:
```json
"Dto": {
  "type": "object",
  "properties": {
    "prop1": {
      "maxLength": 10,  
      "minLength": 2,
      "type": "string",
      "nullable": true
    },
    "prop2": {
      "type": "string",
      "nullable": true
    }
  },
  "additionalProperties": false
}
```
Where the `prop1` field has both `maxLength` and `minLength` values specified, but the `prop2` field doesn't have them at all.

satma0745:
Resolves #2756

### DIFF
--- a/test/Swashbuckle.AspNetCore.TestSupport/Fixtures/TypeWithValidationAttributesViaMetadataType.cs
+++ b/test/Swashbuckle.AspNetCore.TestSupport/Fixtures/TypeWithValidationAttributesViaMetadataType.cs
@@ -12,6 +12,14 @@ namespace Swashbuckle.AspNetCore.TestSupport
 
         public string[] ArrayWithMinMaxLength { get; set; }
 
+#if NET8_0_OR_GREATER
+
+        public string StringWithLength { get; set; }
+
+        public string[] ArrayWithLength { get; set; }
+
+#endif
+
         public int IntWithRange { get; set; }
 
         public string StringWithRegularExpression { get; set; }
@@ -33,6 +41,16 @@ namespace Swashbuckle.AspNetCore.TestSupport
 
         [MinLength(1), MaxLength(3)]
         public string[] ArrayWithMinMaxLength { get; set; }
+
+#if NET8_0_OR_GREATER
+
+        [Length(1, 3)]
+        public string StringWithLength { get; set; }
+
+        [Length(1, 3)]
+        public string[] ArrayWithLength { get; set; }
+
+#endif
 
         [Range(1, 10)]
         public int IntWithRange { get; set; }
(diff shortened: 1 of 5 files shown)


test/Swashbuckle.AspNetCore.TestSupport/Fixtures/TypeWithValidationAttributesViaMetadataType.cs
```diff
@@ -12,6 +12,14 @@ public class TypeWithValidationAttributesViaMetadataType
 
         public string[] ArrayWithMinMaxLength { get; set; }
 
+#if NET8_0_OR_GREATER
+
+        public string StringWithLength { get; set; }
+
+        public string[] ArrayWithLength { get; set; }
```
martincostello, line 19: Should these be called `*WithoutLength`?

satma0745:
What else can be covered?

martincostello:
Codecov seems to think that the majority of your changes aren't actually covered by the tests you added:

image

cremor:
How are the builds deciding which test run (so which target framework) is taken for the coverage report? Maybe they are taking the coverage from a non .NET 8 run?

martincostello:
> How are the builds deciding which test run (so which target framework) is taken for the coverage report? Maybe they are taking the coverage from a non .NET 8 run?

Good question - I'm not sure, but they _should_ be being combined together.

cremor:
If you download the coverage artifact from the build you can see that something is merged. But I'd expect at least 3 coverage files (for the 3 target frameworks used in tests), and it only shows 2:
grafik

And according to that report the new method isn't covered at all, which supports my suspected cause:
grafik

It's also weird that it shows each method twice in the method list:
grafik

martincostello (approved):

martincostello:
I'm happy that it looks like it's covered/tested.

The yes/no coverage report is likely related to #2836 and the fact that the number of projects and TFMs we have causes lots of file write conflicts.

PR #2883, opened 2024-05-15
Update Swagger UI section in README

cremor:
Update a few outdated links to the Swagger UI repository and add missing extension methods on `SwaggerUIOptions`.

I also thought about adding mentions to the properties directly on `SwaggerUIOptions` (e.g. `HeadContent`) but decided against it for now. Those would need some additinal warnings because otherwise you might undo your previous modifications done by the extension methods again.

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,7 +166,7 @@ Additionally, there's add-on packages (CLI tools, [an alternate UI](https://gith
 |Package|Description|
 |---------|-----------|
 |Swashbuckle.AspNetCore.Swagger|Exposes Swagger JSON endpoints. It expects an implementation of `ISwaggerProvider` to be registered in the DI container, which it queries to retrieve `OpenAPIDocument(s)` that are then exposed as serialized JSON|
-|Swashbuckle.AspNetCore.SwaggerGen|Injects an implementation of `ISwaggerProvider` that can be used by the above component. This particular implementation generates `OpenApiDocument(s)` from your routes, controllers and models|
+|Swashbuckle.AspNetCore.SwaggerGen|s an implementation of `ISwaggerProvider` that can be used by the above component. This particular implementation generates `OpenApiDocument(s)` from your routes, controllers and models|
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1292,6 +1295,18 @@ app.UseSwaggerUI(c =>
...
 ```
 
+### Inject Custom JavaScript ###
+
+To tweak the behavior, you can inject additional JavaScript files by adding them to your `wwwroot` folder and specifying the relative paths in the middleware options:
+
+```csharp
+app.UseSwaggerUI(c =>
+{
+    ...
```
martincostello, line 1305: As otherwise it breaks the C# syntax highlighting

```suggestion
```

martincostello (approved):

martincostello:
Thanks. I wouldn't do any more PRs to add further new content (tweaks/fixes are fine) - we think the README has gotten too big at this point and want to rework it. We just haven't worked out exactly what we want to do instead yet.

README.md
cremor: For me it looks the same as in the existing "Inject Custom CSS" section. But if you want I can remove it. The question is then, should it also be removed in the existing section(s)?

martincostello: Well I only see what's in the diff 😄

I would remove it from all so all the snippets are valid.

cremor: Ok, done. At least for the Swagger UI code snippets, because I assumed that's what you meant. ("All" would mean over 40 occurences of "...".)

You can prevew it here: https://github.com/cremor/Swashbuckle.AspNetCore/tree/patch-1?tab=readme-ov-file#inject-custom-javascript

README.md
```diff
@@ -226,7 +226,7 @@ The steps described above will get you up and running with minimal setup. Howeve
     * [Change Document Title](#change-document-title)
     * [List Multiple Swagger Documents](#list-multiple-swagger-documents)
     * [Apply swagger-ui Parameters](#apply-swagger-ui-parameters)
-    * [Inject Custom JavaScript](#inject-custom-javascript)
+    * [Inject Custom ](#inject-custom-)
```
martincostello:
```suggestion
    * [Inject Custom JavaScript](#inject-custom-javascript)
```

PR #2677, opened 2023-06-29
Add customized document serialization support

remcolam:
Added support to inject a ISwaggerDocumentSerializer implementation, which can be called to manually serialize an OpenAPIDocument object.

- Resolves #2668
- Resolves #2700
- Resolves #2105

PR #2886, opened 2024-05-15
Existing output file should be deleted/overwritten

patrikwlund:
The CLI currently uses `File.OpenWrite(path)` (added in https://github.com/domaindrivendev/Swashbuckle.AspNetCore/pull/2677) to create an output stream for the resulting swagger document. This does not delete the existing contents, but instead starts writing from position 0, causing issues if the output is smaller than the previous file content.

I changed it to `File.Create(path)` to make sure any previous content is discarded.

### DIFF
--- a/src/Swashbuckle.AspNetCore.Cli/Program.cs
+++ b/src/Swashbuckle.AspNetCore.Cli/Program.cs
@@ -101,7 +101,7 @@ namespace Swashbuckle.AspNetCore.Cli
                         ? Path.Combine(Directory.GetCurrentDirectory(), arg1)
                         : null;
 
-                    using (Stream stream = outputPath != null ? File.OpenWrite(outputPath) : Console.OpenStandardOutput())
+                    using (Stream stream = outputPath != null ? File.Create(outputPath) : Console.OpenStandardOutput())
                     using (var streamWriter = new FormattingStreamWriter(stream, CultureInfo.InvariantCulture))
                     {
                         IOpenApiWriter writer;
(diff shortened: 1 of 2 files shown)


martincostello (approved):

patrikwlund:
> Thanks for the fix - it looks like this is creating some sort of race condition though.

After investigating, I don't think that is related to this PR, as I can reproduce that on the `master` branch as well. Rebuilding the `test/WebSites/CliExample` project often (but not always) causes the "`being used by another process`" exception.

It may be because it's building net6, net7, and net8 concurrently, and all of them try to output to the same `test/WebSites/CliExample/wwwroot/swagger/v1/swagger.json` file in that project. It happens regardless of `File.OpenWrite()` or `File.Create()`.

martincostello:
I've re-run the CI for a third time now, but if it keeps failing it will need fixing in order to be merged.

martincostello:
It's possible/probable that while it's an existing issue, the presence of the additional test is crossing a threshold that makes it more likely to happen.

If you can't easily fix it, we could temporarily skip the new test (I'm happy conceptually it's covering the issue and the bug is fixed) for now, and re-enable it later.

patrikwlund:
I'll try removing the new test and at least see if it makes a difference in CI.

martincostello:
> I'll try removing the new test and at least see if it makes a difference in CI.

I suspect it will because dependabot just ran and there were no failures.

martincostello:
If it does "fix" it, can you put it back but just mark it as skipped please?

martincostello:
Resolves #2887.

martincostello:
Thanks - I've created #2893 to track the underlying flakiness and we can un-skip the test when we resolve that issue.

Issue #2893, opened 2024-05-15
Fix CI flakiness


martincostello:
Despite some previous attempts (see #2836), there's still some instability in CI, such as #2886.

We need to come up with good lasting solution to avoid these write conflicts in build and test.

martincostello:
When resolved, un-skip this test:

https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/d793865353eaf311eba6ebf32e9b34b7610c937c/test/Swashbuckle.AspNetCore.Cli.Test/ToolTests.cs#L31-L32

EvgeniyZ:
@martincostello it seems you can close this PR as resolved, I'm not seeing any `[Fact(Skip = ...` tests in the codebase

martincostello:
Thanks - I'll re-open this if we get anymore flakiness.

PR #2894, opened 2024-05-15
Disable parallel build

martincostello:
Disable parallel build to see if it fixes file-write conflicts. Relates to #2893.


martincostello:
3 consecutive builds and no failures.

Might not be the best permanent fix, but it seems to be an improvement with no meaningful increase to the build duration.

PR #2897, opened 2024-05-16
Avoid competing swagger generation during test build

patrikwlund:
The previous implementation caused the concurrent builds for various frameworks to compete in writing to the output `swagger.json` file, causing exceptions during builds (since it happened in csproj).

In my local development I often failed to rebuild the full solution because some projects failed to build for this reason. It has also happened in CI (https://github.com/domaindrivendev/Swashbuckle.AspNetCore/pull/2886).

Note that the NSwag source code generator in `NswagClientExample` still outputs everything in one spot for all frameworks, but I haven't noticed failures from that. Maybe it's just less likely to write concurrently. I left it as is.


martincostello:
Do you think this will fix that flaky test from yesterday? If so, can you unskip it?

martincostello:
Just re-running the last build a few times to be sure...

martincostello (approved):
Thanks!

PR #2901, opened 2024-05-16
Fix schema generation for c#9 positional record with param tag without example property creating unexpected empty example

### LINKED ISSUE #2900: [Bug]: C#9 positional Record with xmldoc containing param-tag without example property, generates unexpected 'example=""' for types with "string" json type
### Describe the bug

We recently tried updating from 6.5.1 to 6.6.1 but had to put it off because we started seeing a bunch of these warnings in the OpenApiDiagnostics object being returned when generating the swagger doc
image

The problematic part of the schema was reported to be the generated 'examples' taken from the xmldoc of the start|end property of the record:
image

We could see that it happened for Records using positional syntax with **DateTime** fields and an xmldoc with a **param** tag _without_ the **example** property, like so:
```c#
/// <summary>
/// Some Request DTO
/// </summary>
/// <param name="IncludeSomething">Include a thing</param>
/// <param name="UserIds">Ids of users</param>
/// <param name="Start">Start date of something</param>
/// <param name="End">End date of something</param>
public sealed record SomeRequestDto(
    bool IncludeSomething,
    List<string> UserIds,
    DateTime Start,
    DateTime End);
```

The same behaviour is present for `Bool`, `Guid` and `string`.

### Expected behavior

The swagger doc should not contain empty schema 'examples' when the source records contains param tags without an example property.

### Actual behavior

The swagger doc contains empty schema 'examples' even though the source records do not contain the example property in their param tags.

### Steps to reproduce

Add an endpoint accepting a Record like this:

```c#
/// <summary>
/// Some Request DTO
/// </summary>
/// <param name="IncludeSomething">Include a thing</param>
/// <param name="UserIds">Ids of users</param>
/// <param name="Start">Start date of something</param>
/// <param name="End">End date of something</param>
public sealed record SomeRequestDto(
    bool IncludeSomething,
    List<string> UserIds,
    DateTime Start,
    DateTime End);
```

Call the Swagger endpoint and check the Warnings in the OpenApiDiagnostics of the result.
Check the generated Schema and observe there being empty examples.

### Exception(s) (if any)

_No response_

6.6.1

### .NET Version

net8.0

### Anything else?

I've tracked the issue down to [this line](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/828b7df2ac697d4b3b513d42a17cad152ad9f2e6/src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsSchemaFilter.cs#L52) which is using [this method](https://learn.microsoft.com/en-us/dotnet/api/system.xml.xpath.xpathnavigator.getattribute?view=net-8.0#returns).
image It returns `string.Empty` when the example propery isn't found, which causes the [`TrySetExample`](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/828b7df2ac697d4b3b513d42a17cad152ad9f2e6/src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsSchemaFilter.cs#L67-L73) method to set the empty string as the example.

I can't think of a situation where it is ever desired for this method to allow an empty string anyway, so it should be easily fixed by replacing the `example == null` check with `string.IsNullOrEmpty(example)`

I can push a fix along with a test covering the problem, if desired :)

stb-co:
# Pull Request

Fixes #2900 

I added a check for the case where the example string would be emtpy due to not being found.
I also added a test for when the example property is not provided.


test/Swashbuckle.AspNetCore.SwaggerGen.Test/XmlComments/XmlCommentsSchemaFilterTests.cs
```diff
@@ -62,6 +62,7 @@ public void Apply_SetsDescription_FromFieldSummaryTag()
         [InlineData(typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.LongProperty), "integer", "4294967295")]
         [InlineData(typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.FloatProperty), "number", "1.2")]
         [InlineData(typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.DoubleProperty), "number", "1.25")]
+        [InlineData(typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.DateTimeProperty), "string", "6/22/2022 12:00:00 AM")]
```
martincostello, line 65:
```suggestion
        [InlineData(typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.DateTimeProperty), "string", "\"6/22/2022 12:00:00 AM\"")]
```

test/Swashbuckle.AspNetCore.SwaggerGen.Test/XmlComments/XmlCommentsSchemaFilterTests.cs
```diff
@@ -73,6 +74,7 @@ public void Apply_SetsDescription_FromFieldSummaryTag()
         [InlineData(typeof(XmlAnnotatedRecord), nameof(XmlAnnotatedRecord.LongProperty), "integer", "4294967295")]
         [InlineData(typeof(XmlAnnotatedRecord), nameof(XmlAnnotatedRecord.FloatProperty), "number", "1.2")]
         [InlineData(typeof(XmlAnnotatedRecord), nameof(XmlAnnotatedRecord.DoubleProperty), "number", "1.25")]
+        [InlineData(typeof(XmlAnnotatedRecord), nameof(XmlAnnotatedRecord.DateTimeProperty), "string", "6/22/2022 12:00:00 AM")]
```
martincostello, line 77:
```suggestion
        [InlineData(typeof(XmlAnnotatedRecord), nameof(XmlAnnotatedRecord.DateTimeProperty), "string", "\"6/22/2022 12:00:00 AM\"")]
```

martincostello (approved):

PR #2907, opened 2024-05-16
Fix interceptor JSON casing

### LINKED ISSUE #2906: [Bug]: Request/Response Interceptors Broken with .NET8 Upgrade / `6.6.1`
### Describe the bug

After the upgrade to `6.6.1`, Request and Response interceptors are output to the document with the wrong property names (incorrect casing). They are expected to be written with initial uppercase letters, `RequestInterceptorFunction` and `ResponseInterceptorFunction` but are now being written as `requestInterceptorFunction` and `responseInterceptorFunction`. This causes them to longer function as expect (or rather, at all).

We rely on this functionality and missed this during the upgrade process as we were more focused on the document generation. We will need to roll back unless there's a workaround I'm not aware of

### Expected behavior

The serialized functions should be output using the correct property names (or the expected property name in the index.html should be adjusted to reflect the new names) and should be invoked in response to requests/responses.

### Actual behavior

Functions are serialized with incorrect names and no longer execute.

### Steps to reproduce

options.

previously
```js
var interceptors = JSON.parse('{"RequestInterceptorFunction":"function (req) { }","ResponseInterceptorFunction":null}');
if (interceptors.RequestInterceptorFunction)
    configObject.requestInterceptor = parseFunction(interceptors.RequestInterceptorFunction);
if (interceptors.ResponseInterceptorFunction)
    configObject.responseInterceptor = parseFunction(interceptors.ResponseInterceptorFunction);
```

with latest version (6.6.1)
```js
var interceptors = JSON.parse('{"requestInterceptorFunction":"function (req) { }"}');
if (interceptors.RequestInterceptorFunction)
    configObject.requestInterceptor = parseFunction(interceptors.RequestInterceptorFunction);
if (interceptors.ResponseInterceptorFunction)
    configObject.responseInterceptor = parseFunction(interceptors.ResponseInterceptorFunction);
```

### Exception(s) (if any)

_No response_

6.6.1

### .NET Version

net8.0

### Anything else?

_No response_

martincostello:
Fix property casing of Swagger UI interceptor function properties.

Resolves #2906.

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerUI/SwaggerUIOptions.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerUI/SwaggerUIOptions.cs
@@ -1,10 +1,10 @@
 ﻿using System;
-using System.IO;
-using System.Reflection;
 using System.Collections.Generic;
+using System.IO;
 using System.Linq;
-using System.Text.Json.Serialization;
+using System.Reflection;
 using System.Text.Json;
+using System.Text.Json.Serialization;
 
 namespace Swashbuckle.AspNetCore.SwaggerUI
 {
@@ -233,7 +233,7 @@ namespace Swashbuckle.AspNetCore.SwaggerUI
         /// <summary>
         /// String array of initially selected oauth scopes, default is empty array
         /// </summary>
-        public IEnumerable<string> Scopes { get; set; } = new string[] { };
+        public IEnumerable<string> Scopes { get; set; } = [];
 
         /// <summary>
         /// Additional query parameters added to authorizationUrl and tokenUrl
@@ -262,6 +262,7 @@ namespace Swashbuckle.AspNetCore.SwaggerUI
         /// Accepts one argument requestInterceptor(request) and must return the modified request, or a Promise that resolves to the modified request.
         /// Ex: "function (req) { req.headers['MyCustomHeader'] = 'CustomValue'; return req; }"
         /// </summary>
+        [JsonPropertyName("RequestInterceptorFunction")]
         public string RequestInterceptorFunction { get; set; }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


cremor (approved):

src/Swashbuckle.AspNetCore.SwaggerUI/SwaggerUIOptions.cs
```diff
@@ -262,6 +262,7 @@ public class InterceptorFunctions
         /// Accepts one argument requestInterceptor(request) and must return the modified request, or a Promise that resolves to the modified request.
         /// Ex: "function (req) { req.headers['MyCustomHeader'] = 'CustomValue'; return req; }"
         /// </summary>
+        [JsonPropertyName("RequestInterceptorFunction")]
```
domaindrivendev, line 265: Given that the rest of the config objects are serialized to camel case, I wonder would it be better to maintain that consistency and just update the assignment in `index.html`?

https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/432c417ee5f7d1598de837495e4de97853e8bb4d/src/Swashbuckle.AspNetCore.SwaggerUI/index.html#L99-L103

domaindrivendev: Just a thought but not a showstopper for me other than that looks good 👍

domaindrivendev (approved):

src/Swashbuckle.AspNetCore.SwaggerUI/SwaggerUIOptions.cs
domaindrivendev: doh didn't realize we had auto-merge - oh well I guess we leave it as is

PR #2926, opened 2024-05-29
Fix package validation

martincostello:
Fix the package validation job not failing if there are any issues.


mesMess (approved):
Action changes to support version checks and proper validation.

PR #2938, opened 2024-06-06
Add support for async filters

mauve:
Adds async document, operation, parameter and request body filters.

Fixes: #1661


src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/IDocumentFilter.cs
```diff
@@ -9,6 +10,11 @@ public interface IDocumentFilter
         void Apply(OpenApiDocument swaggerDoc, DocumentFilterContext context);
     }
 
+    public interface IDocumentAsyncFilter
+    {
+        Task Apply(OpenApiDocument swaggerDoc, DocumentFilterContext context);
```
martincostello, line 15:
```suggestion
        Task ApplyAsync(OpenApiDocument swaggerDoc, DocumentFilterContext context);
```

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/IOperationFilter.cs
```diff
@@ -9,6 +10,11 @@ public interface IOperationFilter
         void Apply(OpenApiOperation operation, OperationFilterContext context);
     }
 
+    public interface IOperationAsyncFilter
+    {
+        Task Apply(OpenApiOperation operation, OperationFilterContext context);
```
martincostello, line 15:
```suggestion
        Task ApplyAsync(OpenApiOperation operation, OperationFilterContext context);
```

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/IParameterFilter.cs
```diff
@@ -9,6 +10,11 @@ public interface IParameterFilter
         void Apply(OpenApiParameter parameter, ParameterFilterContext context);
     }
 
+    public interface IParameterAsyncFilter
+    {
+        Task Apply(OpenApiParameter parameter, ParameterFilterContext context);
```
martincostello, line 15:
```suggestion
        Task ApplyAsync(OpenApiParameter parameter, ParameterFilterContext context);
```

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/IRequestBodyFilter.cs
```diff
@@ -9,6 +10,11 @@ public interface IRequestBodyFilter
         void Apply(OpenApiRequestBody requestBody, RequestBodyFilterContext context);
     }
 
+    public interface IRequestBodyAsyncFilter
+    {
+        Task Apply(OpenApiRequestBody requestBody, RequestBodyFilterContext context);
```
martincostello, line 15:
```suggestion
        Task ApplyAsync(OpenApiRequestBody requestBody, RequestBodyFilterContext context);
```

mauve:
Hi @martincostello, Thanks for taking over maintainership of the project ❤️.

Is there a way for me to trigger the checks?

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
```diff
@@ -58,14 +59,25 @@ public async Task<OpenApiDocument> GetSwaggerAsync(string documentName, string h
...
+
             swaggerDoc.Components.Schemas = new SortedDictionary<string, OpenApiSchema>(swaggerDoc.Components.Schemas, _options.SchemaComparer);
 
             return swaggerDoc;
         }
 
         public OpenApiDocument GetSwagger(string documentName, string host = null, string basePath = null)
         {
-            var (applicableApiDescriptions, swaggerDoc, schemaRepository) = GetSwaggerDocumentWithoutFilters(documentName, host, basePath);
+            if (_options.HasAsyncFilters)
```
martincostello, line 74: I don't think there's a need to do this - if someone needs a filter that's asynchronous and has others that don't need it, they should be able to implement it with both interfaces, not force them to refactor all their others to make everything match.

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
```diff
@@ -113,6 +125,38 @@ public OpenApiDocument GetSwagger(string documentName, string host = null, strin
             return (applicableApiDescriptions, swaggerDoc, schemaRepository);
         }
 
+        private async Task<(IEnumerable<ApiDescription>, OpenApiDocument, SchemaRepository)> GetSwaggerDocumentWithoutDocumentFiltersAsync(string documentName, string host = null, string basePath = null)
```
martincostello, line 128: This seems to introduce a lot of duplication between the two implementations. Can it be refactored to just make the existing methods async and then iterate over both sets?

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGeneratorOptions.cs
```diff
@@ -61,12 +65,22 @@ public SwaggerGeneratorOptions()
...
+
         public List<IOperationFilter> OperationFilters { get; set; }
 
+        public List<IOperationAsyncFilter> OperationAsyncFilters { get; set; }
+
         public IList<IDocumentFilter> DocumentFilters { get; set; }
 
+        public IList<IDocumentAsyncFilter> DocumentAsyncFilters { get; set; }
+
+        public bool HasAsyncFilters => (ParameterAsyncFilters?.Any() ?? false) || (RequestBodyAsyncFilters?.Any() ?? false) || (OperationAsyncFilters?.Any() ?? false) || (DocumentAsyncFilters?.Any() ?? false);
```
martincostello, line 82: I don't we need this - they shouldn't be mutually exclusive.

test/Swashbuckle.AspNetCore.SwaggerGen.Test/SwaggerGenerator/SwaggerGeneratorAsyncTests.cs
```diff
@@ -0,0 +1,298 @@
...
+using Microsoft.AspNetCore.Mvc.ApiExplorer;
+using Microsoft.AspNetCore.Mvc.ModelBinding;
+using Microsoft.OpenApi.Any;
+using Microsoft.OpenApi.Models;
+using Swashbuckle.AspNetCore.TestSupport;
+using Xunit;
+
+namespace Swashbuckle.AspNetCore.SwaggerGen.Test
+{
+    public class SwaggerGeneratorAsyncTests
```
martincostello, line 15: Just add these tests to the existing test class?

martincostello:
> Is there a way for me to trigger the checks?

No, it's a built-in feature of GitHub Actions for first time contributors to a repo - someone with write access needs to approve them to run until your first contribution is merged.

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
mauve: I was thinking about something like that, but I am not sure how I can call these async methods in the existing non-async code-path?

Don't we need to support `ISwaggerProvider.GetSwagger()` still?

test/Swashbuckle.AspNetCore.SwaggerGen.Test/SwaggerGenerator/SwaggerGeneratorAsyncTests.cs
mauve: Will do

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
martincostello: I only looked at the diff, I didn't go digging through all the code. Can you point me at the bit where you're looking?

We would need to keep the existing methods/support though, yes.

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGeneratorOptions.cs
martincostello: Ideally people should be able to use both if we can design it that way.

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
mauve: I don't think that is what this code will do, at least what I meant that it does is that it fails if you use an async filter with the sync `GetSwagger()`.

I am not 100% sure when `GetSwagger()` is called instead of `GetSwaggerAsync()` the middleware seems to prefer the async version of the sync version.

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGeneratorOptions.cs
mauve: I don't feel super-strongly about this, just let me know which option you prefer:

1. GetSwagger() silently ignores async filters
2. GetSwagger() throws error if async filters are used

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
mauve: I figured out a way to remove the duplication.

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGeneratorOptions.cs
mauve: > Ideally people should be able to use both if we can design it that way.

Sorry, I missed this comment before my last answer. What are you referring to with `both`? `GetSwagger()` and `GetSwaggerAsync()` or async and non-async filters?

martincostello: Just the filters - if there's an async middleware, then it could use both implementations, where as a sync middleware/interface/whatever would only be able to use the sync ones - in that case it should probably just ignore the async ones entirely.

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
martincostello: When I have time to clone the code locally and dig through it in an IDE I can take a better look at what's the best way to go about this to try and make it "just work" for the async case.

If we were to make breaking changes in the future we'd probably remove the sync versions if we also have async.

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGeneratorOptions.cs
mauve: Ok, I will implement that.

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
mauve: I have now removed this check (and the corresponding property on SwaggerGeneratorOptions)

mauve:
@martincostello The tests are now green :) Let me know if you want any further changes.

martincostello:
Thanks - I'll take a look at this in more detail tomorrow 👍

mauve:
@martincostello Hi, have you had chance to look at it?

martincostello:
Sorry, not yet, but it's on my TODO list.

martincostello:
Hi @mauve - I checked out your pull request locally and had a look around the code. I've done some refactoring in this commit that aims to reduce the duplication in `SwaggerGenerator`: https://github.com/martincostello/Swashbuckle.AspNetCore/commit/01cf4d585f5416649da9dc2fe3b6eb97ca6e3594

What are your thoughts on the approach?

mauve:
Great! Unsure what the process should be but I just cherry-picked your commit to the top of my PR.

martincostello:
Thanks - I'll take a proper look at this again either later this afternoon, or more likely, tomorrow 👍

src/Swashbuckle.AspNetCore.Swagger/IAsyncSwaggerProvider.cs
martincostello, line 1: Sorry, my bad here - I thought this interface was new as part of your PR, not something we already had 😅

This would be a breaking change, so the `cancellationToken` parameter needs to be removed.

That _might_ then make the other `CancellationToken` additions to the new interfaces pointless, unless the `HttpContext` happens to be already present in what's available within `SwaggerGenerator`, in which case it can pass through `HttpContext.RequestAborted`.

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGeneratorOptions.cs
```diff
@@ -61,12 +65,20 @@ public SwaggerGeneratorOptions()
 
         public IList<IParameterFilter> ParameterFilters { get; set; }
 
+        public IList<IParameterAsyncFilter> ParameterAsyncFilters { get; set; }
+
         public List<IRequestBodyFilter> RequestBodyFilters { get; set; }
 
+        public List<IRequestBodyAsyncFilter> RequestBodyAsyncFilters { get; set; }
```
martincostello, line 72:
```suggestion
        public IList<IRequestBodyAsyncFilter> RequestBodyAsyncFilters { get; set; }
```

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGeneratorOptions.cs
```diff
@@ -61,12 +65,20 @@ public SwaggerGeneratorOptions()
...
+
         public List<IOperationFilter> OperationFilters { get; set; }
```
martincostello, line 74:
```suggestion
        public IList<IOperationFilter> OperationFilters { get; set; }
```

mauve: I have changed this aswell, however this is an existing property, isn't that technically an API break?

mauve:
@martincostello I have removed the CancellationToken from GetSwaggerAsync, however I left them in the newly added interfaces so that we don't need to break that interface when we introduce CancellationToken again.

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGeneratorOptions.cs
martincostello: I must have commented on the wrong line - the new properties should use `IList<T>` but anything already existing should be left alone.

src/Swashbuckle.AspNetCore.Swagger/IAsyncSwaggerProvider.cs
```diff
@@ -1,3 +1,4 @@
+using System.Threading;
```
martincostello, line 1: This can be removed.

```suggestion
```

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGeneratorOptions.cs
```diff
@@ -61,12 +65,20 @@ public SwaggerGeneratorOptions()
...
 
-        public List<IRequestBodyFilter> RequestBodyFilters { get; set; }
+        public IList<IParameterAsyncFilter> ParameterAsyncFilters { get; set; }
 
-        public List<IOperationFilter> OperationFilters { get; set; }
+        public IList<IRequestBodyFilter> RequestBodyFilters { get; set; }
+
+        public IList<IRequestBodyAsyncFilter> RequestBodyAsyncFilters { get; set; }
+
+        public IList<IOperationFilter> OperationFilters { get; set; }
```
martincostello, line 74:
```suggestion
        public List<IOperationFilter> OperationFilters { get; set; }
```

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGeneratorOptions.cs
```diff
@@ -61,12 +65,20 @@ public SwaggerGeneratorOptions()
 
         public IList<IParameterFilter> ParameterFilters { get; set; }
 
-        public List<IRequestBodyFilter> RequestBodyFilters { get; set; }
+        public IList<IParameterAsyncFilter> ParameterAsyncFilters { get; set; }
```
martincostello, line 68:
```suggestion
        public List<IParameterAsyncFilter> ParameterAsyncFilters { get; set; }
```

mauve:
I accepted all your suggestions @martincostello

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGeneratorOptions.cs
```diff
@@ -61,12 +65,20 @@ public SwaggerGeneratorOptions()
 
         public IList<IParameterFilter> ParameterFilters { get; set; }
 
-        public List<IRequestBodyFilter> RequestBodyFilters { get; set; }
+        public List<IParameterAsyncFilter> ParameterAsyncFilters { get; set; }
+
+        public IList<IRequestBodyFilter> RequestBodyFilters { get; set; }
```
martincostello, line 70: These two are the wrong way around:

```suggestion
        public IList<IParameterAsyncFilter> ParameterAsyncFilters { get; set; }

        public List<IRequestBodyFilter> RequestBodyFilters { get; set; }
```

martincostello (approved):

mauve:
@martincostello Thanks 🎉 I am very happy have made my first contribution :)

mauve:
Hi @martincostello do you have any plans to create a new release? Is there anything I can assist with in that case?

martincostello:
There's been a drip-feed of changes happening, so I've been waiting for things to quieten down.

Issue #2052, opened 2021-03-10
SwaggerDiscriminator Attribute: discriminator field moved from paths to components


danstur:
After updating from 5.6.3 to 6.x (tested both 6.1.0 and 6.0.7) when using the SwaggerDiscriminator attribute the "discriminator" field moved from `paths/<path>/responses/<..>/content/schema` into `/components/schemas/BaseType`. 

I.e. in the old code we had:
```[json]
"responses": {
  "200": {
	"content": {
	  "application/json": {
		"schema": {
		  "oneOf": [
			{
			  "$ref": "#/components/schemas/Foo"
			},
			{
			  "$ref": "#/components/schemas/Bar"
			}
		  ],
		  "discriminator": {
			"propertyName": "type",
			"mapping": {
			  "Foo": "#/components/schemas/Foo",
			  "Bar": "#/components/schemas/Bar"
			}
		  }
		}
	  }
	}
  }
}
```
now this has become
```[json]
"responses": {
  "200": {
	"content": {
	  "application/json": {
		"schema": {
		  "oneOf": [
			{
			  "$ref": "#/components/schemas/Foo"
			},
			{
			  "$ref": "#/components/schemas/Bar"
			}
		  ]
		}
	  }
	}
  }
}
// components
  "FooBase": {
	"required": [
	  "type"
	],
	"type": "object",
	"properties": {
	  "type": {
		"type": "string",
		"nullable": true,
		"readOnly": true
	  }
	},
	"additionalProperties": false,
	"discriminator": {
	  "propertyName": "type",
	  "mapping": {
		"Foo": "#/components/schemas/Foo",
		"Bar": "#/components/schemas/Bar"
	  }
	}
  }
```

and the discriminator and mapping can only be found under components of the base type. I assume this is an intentional change? 

I can see the point of not repeating the discriminator mappings repeatedly, but it seems rather cumbersome to find the discriminator and mapping with the new structure (basically find the common allOf() parent of all the types referenced in oneOf I think? Not sure this would work for more complicated inheritance hierarchies though). 

The sample shown at [swagger.io](https://swagger.io/docs/specification/data-models/inheritance-and-polymorphism) has the discriminator under the response part as it was done in the old version though.

## Reproduction

Using the basic ASP.NET Core 3.1 template with the following controller:
```[csharp]
    [ApiController]
    [Route("[controller]")]
    public class LogsController : ControllerBase
    {
        [HttpGet]
[...]

PR #2683, opened 2023-07-15
Update SchemaGenerator to fix #2514 and #2052

danielcrabtree:
This pull request fixes (reverts) the placement of the discriminator to the polymorphic oneOf element. Fixes #2052.

This pull request also fixes #2514 and the bug where enabling UseOneOfForPolymorphism forces UseAllOfForInheritance=true (when UseAllOfForInheritance=false).

PR #2940, opened 2024-06-07
Update SchemaGenerator to fix #2514 and #2052 (Rebased)

### LINKED ISSUE #2514: Inheritance hierarchy not flattened by default
According to the docs:
> By default, Swashbuckle flattens inheritance hierarchies. That is, for derived models, the inherited properties are combined and listed alongside the declared properties.

However, when using the Shape-Rectange-Circle hierarchy (as in the [tests](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/master/test/WebSites/Basic/Controllers/PolymorphicTypesController.cs)), the schema generated with `Swashbuckle.AspNetCore.Swagger v6.4.0` still includes the base (parent) fields via allOf.

**Expected**: `name` and `typeName` to end up under `Circle` -> properties.
**Actual**: `allOf` that references `Shape` is under `Circle`, despite **not** calling `c.UseAllOfForInheritance()` in the code.

```
openapi: 3.0.1
info:
  title: 'WebApplication1, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null'
  version: '1.0'
paths:
  /Shapes:
    get:
      tags:
        - Shapes
      operationId: GetShape
      responses:
        '200':
          description: Success
          content:
            text/plain:
              schema:
                oneOf:
                  - $ref: '#/components/schemas/Rectangle'
                  - $ref: '#/components/schemas/Circle'
            application/json:
              schema:
                oneOf:
                  - $ref: '#/components/schemas/Rectangle'
                  - $ref: '#/components/schemas/Circle'
            text/json:
              schema:
                oneOf:
                  - $ref: '#/components/schemas/Rectangle'
                  - $ref: '#/components/schemas/Circle'
components:
  schemas:
    Circle:
      type: object
      allOf:
        - $ref: '#/components/schemas/Shape'
      properties:
        radius:
          type: integer
          format: int32
      additionalProperties: false
    Rectangle:
      type: object
      allOf:
        - $ref: '#/components/schemas/Shape'
      properties:
        height:
          type: integer
          format: int32
        width:
          type: integer
          format: int32
      additionalProperties: false
    Shape:
      required:
        - typeName
      type: object
      properties:
        typeName:
          type: string
        name:
          type: string
          nullable: true
      additionalProperties: false
      discriminator:
        propertyName: typeName
        mapping:
          Rectangle: '#/components/schemas/Rectangle'
          Circle: '#/components/schemas/Circle'
```

Here is an (almost) minimal repro with .NET 6 (Program.cs):
```
using Microsoft.AspNetCore.Mvc;

var builder = WebApplication.CreateBuilder(args);

builder.Services.AddControllers();
builder.Services.AddSwaggerGen(c =>
{
    c.UseOneOfForPolymorphism();

    c.SelectDiscriminatorNameUsing((baseType) => "typeName");
    c.SelectDiscriminatorValueUsing((subType) => subType.Name);
});

var app = builder.Build();

app.UseSwagger();
app.UseSwaggerUI();

app.UseHttpsRedirection();
app.UseAuthorization();
app.MapControllers();

app.Run();

[ApiController]
[Route("[controller]")]
public class ShapesController : ControllerBase
{
    [HttpGet(Name = "GetShape")]
    public Shape GetShape() => new Rectangle {Height = 10, Width = 5};


    public abstract class Shape
    {
    }

    public class Rectangle : Shape
    {
        public int Height { get; set; }

        public int Width { get; set; }
    }

    public class Circle : Shape
    {
        public int Radius { get; set; }
    }
}
```

And here is the `WebApplication1.csproj` file:
```
<Project Sdk="Microsoft.NET.Sdk.Web">
    <PropertyGroup>
        <TargetFramework>net6.0</TargetFramework>
        <Nullable>enable</Nullable>
        <ImplicitUsings>enable</ImplicitUsings>
    </PropertyGroup>
    <ItemGroup>
        <PackageReference Include="Newtonsoft.Json" Version="12.0.2" />
        <PackageReference Include="Swashbuckle.AspNetCore" Version="6.4.0" />
        <PackageReference Include="Swashbuckle.AspNetCore.Swagger" Version="6.4.0" />
    </ItemGroup>
</Project>
```

leth:
Rebases #2683

#2052 #2514

> This pull request fixes (reverts) the placement of the discriminator to the polymorphic oneOf element. Fixes https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/2052.
>
> This pull request also fixes https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/2514 and the bug where enabling UseOneOfForPolymorphism forces UseAllOfForInheritance=true (when UseAllOfForInheritance=false).

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/SchemaGenerator.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/SchemaGenerator.cs
@@ -174,6 +174,7 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
             knownTypesDataContracts = null;
 
             if (dataContract.DataType != DataType.Object) return false;
+            if (dataContract.UnderlyingType.IsAssignableToOneOf(BinaryStringTypes)) return false;
 
             var subTypes = _generatorOptions.SubTypesSelector(dataContract.UnderlyingType);
 
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/SchemaGenerator.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/SchemaGenerator.cs
@@ -387,21 +387,18 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
                         .Where(dataProperty => dataProperty.MemberInfo.DeclaringType == dataContract.UnderlyingType);
                 }
 
-                if (_generatorOptions.UseOneOfForPolymorphism)
+                if (IsBaseTypeWithKnownTypesDefined(dataContract, out var knownTypesDataContracts))
                 {
-                    if (IsBaseTypeWithKnownTypesDefined(dataContract, out var knownTypesDataContracts))
+                    foreach (var knownTypeDataContract in knownTypesDataContracts)
                     {
-                        foreach(var knownTypeDataContract in knownTypesDataContracts)
-                        {
-                            // Ensure schema is generated for all known types
-                            GenerateConcreteSchema(knownTypeDataContract, schemaRepository);
-                        }
+                        // Ensure schema is generated for all known types
+                        GenerateConcreteSchema(knownTypeDataContract, schemaRepository);
+                    }
 
-                        if (TryGetDiscriminatorFor(dataContract, schemaRepository, knownTypesDataContracts, out var discriminator))
-                        {
-                            schema.Properties.Add(discriminator.PropertyName, new OpenApiSchema { Type = "string" });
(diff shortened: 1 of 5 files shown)


leth:
I've posted my current progress, but I think I'm stuck now. The code compiles and work, but the unit tests fail.
I don't know how to fix them because I don't understand the different use cases well enough.

Specifically,
1. What is the interplay between UseAllOfForInheritance and UseOneOfForPolymorphism?
2. Does it (ever?) make sense for the BaseType to keep a discriminator entry?
3. What works well/at all with other projects like code generators/Json deserialisers?

Issue #2920, opened 2024-05-28
[Bug]: The JSON value could not be converted to System.DateOnly


Bidou44:
### Describe the bug

Everything was working fine with v6.5.0. After migrating to v6.6.1 or v6.6.2, I got the following exception

>  System.Text.Json.JsonException: The JSON value could not be converted to System.DateOnly.

I didn't change any line of code, just migrated the package.

### Expected behavior

the json definition should load correctly. Instead, I got an error message.

### Actual behavior

This is what I get

```
{
    "type": "https://tools.ietf.org/html/rfc9110#section-15.6.1",
    "title": "Internal Server Error",
    "status": 500,
    "detail": "An unhandled exception occurred",
    "instance": "/api/doc/v1.json",
    "traceId": "f0b3b9ac7ca1e1711f1ebfbb291f8385",
[...]

Issue #2937, opened 2024-06-06
[Bug]: DefaultValue(null) doesn't work with nullable int


PaulKamps:
### Describe the bug

I have a few properties that are as following:

[DefaultValue(null)]
public int? Id { get; set; }

### Expected behavior

I expect it to work as it did in 6.5.0 and before.
Now it crashes when loading the swagger page with the following error: `The JSON value could not be converted to System.Int32.`

### Actual behavior

`System.Text.Json.JsonException: The JSON value could not be converted to System.Int32.
   at System.Text.Json.ThrowHelper.ThrowJsonException_DeserializeUnableToConvertValue(Type propertyType)
   at System.Text.Json.JsonSerializer.UnboxOnWrite[T](Object value)
   at System.Text.Json.Serialization.Metadata.JsonTypeInfo`1.SerializeAsObject(Utf8JsonWriter writer, Object rootValue)
   at System.Text.Json.JsonSerializer.WriteStringAsObject(Object value, JsonTypeInfo jsonTypeInfo)
   at Swashbuckle.AspNetCore.SwaggerGen.JsonSerializerDataContractResolver.JsonConverterFunc(Object value, Type type)
   at Swashbuckle.AspNetCore.SwaggerGen.JsonSerializerDataContractResolver.<>c__DisplayClass2_0.<GetDataContractForType>b__1(Object value)
   at Swashbuckle.AspNetCore.SwaggerGen.SchemaGenerator.GenerateDefaultValue(DataContract dataContract, Type modelType, Object defaultValue)
   at Swashbuckle.AspNetCore.SwaggerGen.SchemaGenerator.GenerateSchemaForMember(Type modelType, SchemaRepository schemaRepository, MemberInfo memberInfo, DataProperty dataProperty)
   at Swashbuckle.AspNetCore.SwaggerGen.SchemaGenerator.CreateObjectSchema(DataContract dataContract, SchemaRepository schemaRepository)
   at Swashbuckle.AspNetCore.SwaggerGen.SchemaGenerator.<>c__DisplayClass11_0.<GenerateConcreteSchema>b__3()
   at Swashbuckle.AspNetCore.SwaggerGen.SchemaGenerator.GenerateReferencedSchema(DataContract dataContract, SchemaRepository schemaRepository, Func`1 definitionFactory)
[...]

PR #2941, opened 2024-06-07
default null value on nullable types caused errors

jgarciadelanoceda:
# DefaultValue Null on Nullable types caused errors

This PR fixes the issues #2920 and  #2937

When serializing the default  value on a Nullable type on System.Text.Json the library was trying to serialize it as the not nullable type, this was not possible on STJ because it gave an exception.


martincostello (approved):

cremor:
@martincostello The two issues which are referencec in this PR are still open. Can they be closed now that this PR is merged?

Issue #2939, opened 2024-06-07
[Bug]: Authentication problem with 6.6.2 


kenwonders:
### Describe the bug

Upgrading from 6.50 to 6.6.2 caused our authentication system to stop working. We have a separate identity server using client credentials and with 6.6.2 we are always getting the error:

![swashbuckle error](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/assets/5612157/a91188f3-8f56-4fe7-8219-804cbca64d8e)

Reverting just Swashbuckle.AspNetCore back to 6.5.0 resolves the issue.

Our .AddSwaggerGen code:

```
                c.AddSecurityDefinition("Bearer", new OpenApiSecurityScheme
                {
                    Type = SecuritySchemeType.OAuth2,
                    Flows = new OpenApiOAuthFlows
                    {
                        ClientCredentials = new OpenApiOAuthFlow
                        {
                            TokenUrl = new Uri(Configuration["Identity:Authority"] + "/connect/token"),
                            Scopes = new Dictionary<string, string>
                            {
                                { "***.User", "" },
                                { "***.Admin", "" }
                            },
                            AuthorizationUrl = new Uri(Configuration["Identity:Authority"] + "/oauth2/authorize"),
                        }
                    }
                });
                c.AddSecurityRequirement(new OpenApiSecurityRequirement
                {
                    {
                        new OpenApiSecurityScheme
                        {
                            Reference = new OpenApiReference
                            {
                                Id = "Bearer",
                                Type = ReferenceType.SecurityScheme
                            }
                        },
                        new List<string>()
                    }
                });

            services.AddAuthentication(JwtBearerDefaults.AuthenticationScheme)
                .AddJwtBearer(o =>
                {                    
                    o.Authority = Configuration["Identity:Authority"];
                    o.RequireHttpsMetadata = bool.Parse(Configuration["Identity:RequireHttpsMetadata"]);
                    o.Audience = Configuration["Identity:Audience"];
                    o.TokenValidationParameters = new TokenValidationParameters { ValidateAudience = bool.Parse(Configuration["Identity:ValidateAudience"]) };                                        
                });

            services.AddAuthorization(options =>
            {
                options.AddPolicy("***", policy => policy.RequireClaim("client_Role", "***", "***"));
                options.AddPolicy("***", policy => policy.RequireClaim("client_Role", "***"));
            });

```

In our UseSwaggerUI:

```
[...]

PR #2942, opened 2024-06-09
Swagger UI refactoring

martincostello:
Refactoring for a few things I spotted while investigating #2939:

- Add explicit `[JsonPropertyName]` attributes to the Swagger UI configuration objects to ensure we don't accidentally break things with refactoring on the C# side (property name changes etc.)
- Apply some IDE refactoring suggestions (collection expressions, redundant usings).
- Use HTTPS with Kestrel for the OAuth 2 sample.
- Fix broken operation filter for the OAuth 2 sample.
- Update old documentation URL.


mesMess (approved):
Syntax cleanup
Using newer C# syntax
Uncommenting useful code
Slick LINQ call: GetPolicy w/o param

Issue #1685, opened 2020-05-29
Can't generate enum item descriptions


Mazas17:
According to OpenAPI documentation, it is possible to add description to single enum item:
```
      parameters:
        - in: query
          name: sort
          schema:
            type: string
            enum: [asc, desc]
          description: >
            Sort order:
             * `asc` - Ascending, from A to Z
             * `desc` - Descending, from Z to A
```
https://swagger.io/docs/specification/data-models/enums/

How to do that with swashbuckle? Can't find any documentation

jgarciadelanoceda:
Right now if you combine that with a DocumentFilter you will have everything.

```c#
public class EnumDocumentFilter : IDocumentFilter
{
	public void Apply(OpenApiDocument swaggerDoc, DocumentFilterContext context)
	{
		foreach (var path in swaggerDoc.Paths.Values)
		{
			foreach (var operation in path.Operations.Values)
			{
				foreach (var parameter in operation.Parameters)
				{
					var schematypeId = parameter.Schema.Reference?.Id;

					if (!string.IsNullOrEmpty(schematypeId) && context.SchemaRepository.Schemas.TryGetValue(schematypeId, out var schema))
					{
						if ((schema.Enum?.Count ?? 0) > 0)
						{
							parameter.Description = schema.Description;
						}
					}
					if (!string.IsNullOrWhiteSpace(parameter.Schema.Description))
					{
						parameter.Description = parameter.Schema.Description;
					}
				}
			}
		}
	}
}
```

jgarciadelanoceda:
The IDocumentFilter could be replaced by a IParameterFilter that it's much simpler:

```c#
public class ParameterFilter : IParameterFilter
{
	public void Apply(OpenApiParameter parameter, ParameterFilterContext context)
	{
		if (string.IsNullOrWhiteSpace(parameter.Description))
		{
			parameter.Description ??= parameter.Schema.Description;
			if (string.IsNullOrEmpty(parameter.Description)
				&& !string.IsNullOrEmpty(parameter.Schema?.Reference?.Id)
				&& context.SchemaRepository.Schemas.TryGetValue(parameter.Schema.Reference.Id, out var value))
			{
				parameter.Description = value.Description;
			}
		}
	}
}
```

PR #2943, opened 2024-06-11
Apply only a SchemaFilter to create the Description on the SwaggerUi

jgarciadelanoceda:
#Apply only a SchemaFilter to create the Description on the SwaggerUi

This PR fixes the issue #1685.
I do not know if it's a SwaggerUi Bug, because this PR only copies the Description that was obtained in the SchemaFilter to the Parameters section. If the description is not showed under every parameter of the operation then the SwaggerUi does not show it.

### DIFF
--- a/test/Swashbuckle.AspNetCore.SwaggerGen.Test/SwaggerGenerator/SwaggerGeneratorTests.cs
+++ b/test/Swashbuckle.AspNetCore.SwaggerGen.Test/SwaggerGenerator/SwaggerGeneratorTests.cs
@@ -1723,15 +1723,69 @@ namespace Swashbuckle.AspNetCore.SwaggerGen.Test
             Assert.Equal(new[] { nameof(SwaggerIngoreAnnotatedType.NotIgnoredString) }, mediaType.Encoding.Keys);
         }
 
+        [Fact]
+        public void GetSwagger_Copies_Description_From_GeneratedSchema()
+        {
+            var propertyEnum = typeof(TypeWithDefaultAttributeOnEnum).GetProperty(nameof(TypeWithDefaultAttributeOnEnum.EnumWithDefault));
+            var modelMetadataForEnum = new DefaultModelMetadata(
+                                    new DefaultModelMetadataProvider(new FakeICompositeMetadataDetailsProvider()),
+                                    new FakeICompositeMetadataDetailsProvider(),
+                                    new DefaultMetadataDetails(ModelMetadataIdentity.ForProperty(propertyEnum, typeof(IntEnum), typeof(TypeWithDefaultAttributeOnEnum)), ModelAttributes.GetAttributesForProperty(typeof(TypeWithDefaultAttributeOnEnum), propertyEnum)));
+
+            var propertyEnumArray = typeof(TypeWithDefaultAttributeOnEnum).GetProperty(nameof(TypeWithDefaultAttributeOnEnum.EnumArrayWithDefault));
+            var modelMetadataForEnumArray = new DefaultModelMetadata(
+                                    new DefaultModelMetadataProvider(new FakeICompositeMetadataDetailsProvider()),
+                                    new FakeICompositeMetadataDetailsProvider(),
+                                    new DefaultMetadataDetails(ModelMetadataIdentity.ForProperty(propertyEnum, typeof(IntEnum), typeof(TypeWithDefaultAttributeOnEnum)), ModelAttributes.GetAttributesForProperty(typeof(TypeWithDefaultAttributeOnEnum), propertyEnum)));
+            var subject = Subject(
+               apiDescriptions:
+               [
+                   ApiDescriptionFactory.Create<FakeController>(
+                        c => nameof(c.ActionHavingFromFormAttributeWithSwaggerIgnore),
+                        groupName: "v1",
+                        httpMethod: "POST",
+                        relativePath: "resource",
+                        parameterDescriptions: new[]
+                        {
+                            new ApiParameterDescription
+                            {
+                                Name = nameof(TypeWithDefaultAttributeOnEnum.EnumWithDefault),
+                                Source = BindingSource.Query,
+                                Type = typeof(IntEnum),
+                                ModelMetadata = modelMetadataForEnum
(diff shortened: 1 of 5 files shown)


test/Swashbuckle.AspNetCore.SwaggerGen.Test/SwaggerGenerator/SwaggerGeneratorTests.cs
```diff
@@ -1723,15 +1723,69 @@ public void GetSwagger_Works_As_Expected_When_FromForm_Attribute_With_SwaggerIgn
...
+           );
+            var document = subject.GetSwagger("v1");
+
+            var operation = document.Paths["/resource"].Operations[OperationType.Post];
+            Assert.NotEmpty(operation.Parameters);
+            Assert.Equal(nameof(TypeWithDefaultAttributeOnEnum.EnumWithDefault), operation.Parameters[0].Name);
+            Assert.NotEmpty(operation.Parameters[0].Description);
+            Assert.Equal(nameof(TypeWithDefaultAttributeOnEnum.EnumArrayWithDefault), operation.Parameters[1].Name);
+            Assert.NotEmpty(operation.Parameters[1].Description);
+
```
martincostello, line 1776:
```suggestion
```

martincostello (commented):
Are you able to add an integration test for this scenario too?

test/Swashbuckle.AspNetCore.SwaggerGen.Test/SwaggerGenerator/SwaggerGeneratorTests.cs
jgarciadelanoceda: I am not very familiar with IntegrationTest. If I add a controller in the Basic webSite and the EnumSchemaFilter then all is OK right?

martincostello: If you take a look at the recently added verification tests, some thing like that.

jgarciadelanoceda: I think right now is done :). I change the test to the Verify part :)

jgarciadelanoceda:
Can I relaunch the builts?. I do not know why it had errors

jgarciadelanoceda:
> Are you able to add an integration test for this scenario too?

Finally I gave up, because the Verify test was giving different response in the linux build

martincostello:
Do you use an editor that supports EditorConfig?

The [.editorconfig](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/a97c53fd4dfe1a2ca2e2607959cf6c002b02c2e0/.editorconfig#L35-L43) and [.gitattributes](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/a97c53fd4dfe1a2ca2e2607959cf6c002b02c2e0/.gitattributes) are both configured so that line endings _shouldn't_ be an issue, unless I've misunderstood how it works. Alternatively, does it have a setting that allows us to tell it to ignore line ending differences?

jgarciadelanoceda:
As soon as I insert the Verify again the build fail, I do not see why it failed now. Also I have changed the condition from NotEmpty to be equal to the schema description (This could be the reason why you suggested to create the test using verify)

test/Swashbuckle.AspNetCore.SwaggerGen.Test/SwaggerGeneratorVerifyTests/SwaggerGeneratorVerifyTests.GetSwagger_Copies_Description_From_GeneratedSchema.verified.txt
martincostello, line 1: If you deleted the verify test then this can be deleted too?

jgarciadelanoceda: So true!. I am seeing that there is a PR to move the test to Verify and it's having the same problems as I had. Maybe with that PR they discover what is failing. At first was the Linux test because I did not save the file in Visual Studio.. But then it failed the Windows built

martincostello:
Could you rebase onto the latest changes please? I just merged #2950 which added a bunch of new verification tests and this change might impact them and need regenerating.

test/Swashbuckle.AspNetCore.SwaggerGen.Test/SwaggerGeneratorVerifyTests/SwaggerGeneratorVerifyTests.cs
```diff
@@ -224,14 +224,15 @@ public Task ActionWithRequiredMember()
...
+            IEnumerable<ApiDescription> apiDescriptions,
+            SwaggerGeneratorOptions options = null,
+            IEnumerable<AuthenticationScheme> authenticationSchemes = null,
+            List<ISchemaFilter> schemaFilters = null)
     {
         return new SwaggerGenerator(
             options ?? DefaultOptions,
             new FakeApiDescriptionGroupCollectionProvider(apiDescriptions),
-            new SchemaGenerator(new SchemaGeneratorOptions(), new JsonSerializerDataContractResolver(new JsonSerializerOptions())),
+            new SchemaGenerator(new SchemaGeneratorOptions() { SchemaFilters = schemaFilters ?? [] }, new JsonSerializerDataContractResolver(new JsonSerializerOptions())),
```
jgarciadelanoceda, line 235: @martincostello  I have not removed this since it's a good addition here, if we change the test to Verify ones then this does not need to be done

jgarciadelanoceda: Finally I have managed to get the build working

martincostello (approved):

PR #2958, opened 2024-06-20
Adding support for .Net8 Model State Attributes

### LINKED ISSUE #2957: [Feature request]: Support new .NET 8 Model State attributes

Support for `[Length]` was recently added in #2882. There are more Model State attributes in .NET 8 that Swashbuckle could potentially support:

| ASP.NET | .NET Types | OpenAPI |
| --- | --- | --- |
| `[Base64String]` | `string` | `"format": "byte"` |
| `[Range(..., MinimumIsExclusive = true, MaximumIsExclusive = true)]` | `int`, `double`, `IComparable` | `"exclusiveMinimum": true, "exclusiveMaximum": true` |
| `[AllowedValues]` | any (uses `object.Equals`) | `anyOf` maybe? |
| `[DeniedValues]` | any (uses `object.Equals`) |`pattern` maybe? (tricky to get the escaping right) |

Please take the entries in the table above into consideration.

### Additional context

_No response_

jgarciadelanoceda:
Fixes #2957

It adds support for Base64String and MinimumIsExclusive and MaximumIsExclusive  properties for RangeAttribute.
The SwaggerUi does not seem to be looking at neither of these properties when validating it

### DIFF
--- a/test/Swashbuckle.AspNetCore.Newtonsoft.Test/SchemaGenerator/NewtonsoftSchemaGeneratorTests.cs
+++ b/test/Swashbuckle.AspNetCore.Newtonsoft.Test/SchemaGenerator/NewtonsoftSchemaGeneratorTests.cs
@@ -322,6 +322,9 @@ namespace Swashbuckle.AspNetCore.Newtonsoft.Test
             Assert.Equal(3, schema.Properties["StringWithLength"].MaxLength);
             Assert.Equal(1, schema.Properties["ArrayWithLength"].MinItems);
             Assert.Equal(3, schema.Properties["ArrayWithLength"].MaxItems);
+            Assert.Equal(true, schema.Properties["IntWithRange"].ExclusiveMinimum);
+            Assert.Equal(true, schema.Properties["IntWithRange"].ExclusiveMaximum);
+            Assert.Equal("byte", schema.Properties["StringWithBase64"].Format);
 #endif
             Assert.Equal(1, schema.Properties["IntWithRange"].Minimum);
             Assert.Equal(10, schema.Properties["IntWithRange"].Maximum);
--- a/test/Swashbuckle.AspNetCore.Newtonsoft.Test/SchemaGenerator/NewtonsoftSchemaGeneratorTests.cs
+++ b/test/Swashbuckle.AspNetCore.Newtonsoft.Test/SchemaGenerator/NewtonsoftSchemaGeneratorTests.cs
@@ -325,6 +325,7 @@ namespace Swashbuckle.AspNetCore.Newtonsoft.Test
             Assert.Equal(true, schema.Properties["IntWithRange"].ExclusiveMinimum);
             Assert.Equal(true, schema.Properties["IntWithRange"].ExclusiveMaximum);
             Assert.Equal("byte", schema.Properties["StringWithBase64"].Format);
+            Assert.Equal("string", schema.Properties["StringWithBase64"].Type);
 #endif
             Assert.Equal(1, schema.Properties["IntWithRange"].Minimum);
             Assert.Equal(10, schema.Properties["IntWithRange"].Maximum);
--- a/test/Swashbuckle.AspNetCore.SwaggerGen.Test/SchemaGenerator/JsonSerializerSchemaGeneratorTests.cs
+++ b/test/Swashbuckle.AspNetCore.SwaggerGen.Test/SchemaGenerator/JsonSerializerSchemaGeneratorTests.cs
@@ -348,6 +348,7 @@ namespace Swashbuckle.AspNetCore.SwaggerGen.Test
             Assert.Equal(true, schema.Properties["IntWithRange"].ExclusiveMinimum);
             Assert.Equal(true, schema.Properties["IntWithRange"].ExclusiveMaximum);
             Assert.Equal("byte", schema.Properties["StringWithBase64"].Format);
+            Assert.Equal("string", schema.Properties["StringWithBase64"].Type);
 #endif
             Assert.Equal(1, schema.Properties["IntWithRange"].Minimum);
             Assert.Equal(10, schema.Properties["IntWithRange"].Maximum);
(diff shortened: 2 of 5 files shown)


test/Swashbuckle.AspNetCore.TestSupport/Fixtures/TypeWithValidationAttributesViaMetadataType.cs
```diff
@@ -17,6 +17,7 @@ public class TypeWithValidationAttributesViaMetadataType
         public string StringWithLength { get; set; }
 
         public string[] ArrayWithLength { get; set; }
+        public string StringWithBase64 { get; set; }
```
martincostello, line 20: Please keep the spacing consistent:

```suggestion

        public string StringWithBase64 { get; set; }
```

test/Swashbuckle.AspNetCore.TestSupport/Fixtures/TypeWithValidationAttributesViaMetadataType.cs
```diff
@@ -46,14 +47,18 @@ public class MetadataType
...
-
+        [Range(1, 10, MinimumIsExclusive = true, MaximumIsExclusive = true)]
+        public int IntWithRange { get; set; }
+        [Base64String]
+        public string StringWithBase64 { get; set; }
+#else
         [Range(1, 10)]
         public int IntWithRange { get; set; }
+#endif
+
```
martincostello, line 61:
```suggestion
```

test/Swashbuckle.AspNetCore.TestSupport/Fixtures/TypeWithValidationAttributesViaMetadataType.cs
```diff
@@ -46,14 +47,18 @@ public class MetadataType
...
         public string StringWithLength { get; set; }
-
         [Length(1, 3)]
         public string[] ArrayWithLength { get; set; }
 
-#endif
-
+        [Range(1, 10, MinimumIsExclusive = true, MaximumIsExclusive = true)]
+        public int IntWithRange { get; set; }
+        [Base64String]
```
martincostello, line 55:
```suggestion

        [Base64String]
```

test/Swashbuckle.AspNetCore.TestSupport/Fixtures/TypeWithValidationAttributesViaMetadataType.cs
```diff
@@ -46,14 +47,18 @@ public class MetadataType
 
         [Length(1, 3)]
         public string StringWithLength { get; set; }
-
```
martincostello, line 49: Put this back.

test/Swashbuckle.AspNetCore.TestSupport/Fixtures/TypeWithValidationAttributesViaMetadataType.cs
jgarciadelanoceda: Done!

jgarciadelanoceda:
@martincostello can I rerun the validation?. I think that the error is not a real one

test/Swashbuckle.AspNetCore.Newtonsoft.Test/SchemaGenerator/NewtonsoftSchemaGeneratorTests.cs
```diff
@@ -322,6 +322,9 @@ public void GenerateSchema_SetsValidationProperties_IfComplexTypeHasValidationAt
             Assert.Equal(3, schema.Properties["StringWithLength"].MaxLength);
             Assert.Equal(1, schema.Properties["ArrayWithLength"].MinItems);
             Assert.Equal(3, schema.Properties["ArrayWithLength"].MaxItems);
+            Assert.Equal(true, schema.Properties["IntWithRange"].ExclusiveMinimum);
+            Assert.Equal(true, schema.Properties["IntWithRange"].ExclusiveMaximum);
+            Assert.Equal("byte", schema.Properties["StringWithBase64"].Format);
```
martincostello, line 327: Feels like we should be asserting that the `type` is `string` as well.

test/Swashbuckle.AspNetCore.SwaggerGen.Test/SchemaGenerator/JsonSerializerSchemaGeneratorTests.cs
```diff
@@ -345,6 +345,9 @@ public void GenerateSchema_SetsValidationProperties_IfComplexTypeHasValidationAt
             Assert.Equal(3, schema.Properties["StringWithLength"].MaxLength);
             Assert.Equal(1, schema.Properties["ArrayWithLength"].MinItems);
             Assert.Equal(3, schema.Properties["ArrayWithLength"].MaxItems);
+            Assert.Equal(true, schema.Properties["IntWithRange"].ExclusiveMinimum);
+            Assert.Equal(true, schema.Properties["IntWithRange"].ExclusiveMaximum);
+            Assert.Equal("byte", schema.Properties["StringWithBase64"].Format);
```
martincostello, line 350: Same here.

test/Swashbuckle.AspNetCore.Newtonsoft.Test/SchemaGenerator/NewtonsoftSchemaGeneratorTests.cs
jgarciadelanoceda: I do not see the point, because right now if you put the Base64StringAttribute over a thing that it's not a string it will generate the format as well.

I doubt about it as you can see here: https://github.com/domaindrivendev/Swashbuckle.AspNetCore/pull/2958/commits/454168b213feb3fb1421ff9416c088dabecb4b12

But I decided to apply the same policy as RangeAttribute, if you put the RangeAttribute over a thing that is not a number it will put in the Swagger the values. For example if you put the RangeAttribute over a DateTimeOnly you will see the Swagger generated with Minimum and Maximum values.

If you think that we should apply the Base64StringAttribute over a property that is a string or the RangeAttribute over a property that is a number I think that should be a different issue

martincostello: I still think it's worth asserting on as that's the intended outcome of the schema. It might not be related to the fact it's added on _any_ type, but this test is specifically testing base64 strings so I think we should assert that fact.

martincostello: It's also one line of code and trivial to add.

jgarciadelanoceda: Ok, as soon as I have access to the computer I'll add it

martincostello (approved):

PR #2960, opened 2024-06-21
Only set Exclusive Range when they are, do not set them when they are…

### LINKED ISSUE #2959: [Bug]: Breaking: Writes default values for exclusiveMaximum/exclusiveMinimum to openapi.json
### Describe the bug

@jgarciadelanoceda Thanks for implementing #2958! I gave it a try today on our codebase using the MyGet feed.

Unfortunately, I'm getting test failures on `openapi.json` files that don't use `[Range]`. This is because `exclusiveMinimum` and `exclusiveMaximum` are being written even when set to `false`, or not set at all. In those cases, they should be omitted from the output `openapi.json`, because the default is already `false` in OpenAPI (and also in .NET).

### Expected behavior

No `"exclusiveMaximum": false, "exclusiveMinimum": false` when not using `[Range]` or the Exclusive* properties on `[Range]` are set to `false`.

### Actual behavior

The fragment `"exclusiveMaximum": false, "exclusiveMinimum": false` is written to output `openapi.json`, causing our assertions on the file contents to fail.

### Steps to reproduce

_No response_

### Exception(s) (if any)

_No response_

latest master (2db53de)

### .NET Version

.NET 8

### Anything else?

_No response_

jgarciadelanoceda:
Fixes #2959, it was a little breaking change that was affecting a client implementation.
There is no need to set the Inclusive/Exclusive parameter when they are not set to true, to have the same OpenApi document with dotnet7 and dotnet8

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/OpenApiSchemaExtensions.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/OpenApiSchemaExtensions.cs
@@ -183,12 +183,12 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
         {
 #if NET8_0_OR_GREATER
 
-            if(rangeAttribute.MinimumIsExclusive)
+            if (rangeAttribute.MinimumIsExclusive)
             {
                 schema.ExclusiveMinimum = true;
             }
 
-            if(rangeAttribute.MaximumIsExclusive)
+            if (rangeAttribute.MaximumIsExclusive)
             {
                 schema.ExclusiveMaximum = true;
             }
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/OpenApiSchemaExtensions.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/OpenApiSchemaExtensions.cs
@@ -183,8 +183,15 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
         {
 #if NET8_0_OR_GREATER
 
-            schema.ExclusiveMinimum = rangeAttribute.MinimumIsExclusive;
-            schema.ExclusiveMaximum = rangeAttribute.MaximumIsExclusive;
+            if(rangeAttribute.MinimumIsExclusive)
+            {
+                schema.ExclusiveMinimum = true;
+            }
+
+            if(rangeAttribute.MaximumIsExclusive)
+            {
+                schema.ExclusiveMaximum = true;
+            }
 
 #endif
 
--- a/test/Swashbuckle.AspNetCore.Newtonsoft.Test/SchemaGenerator/NewtonsoftSchemaGeneratorTests.cs
+++ b/test/Swashbuckle.AspNetCore.Newtonsoft.Test/SchemaGenerator/NewtonsoftSchemaGeneratorTests.cs
@@ -322,10 +322,15 @@ namespace Swashbuckle.AspNetCore.Newtonsoft.Test
             Assert.Equal(3, schema.Properties["StringWithLength"].MaxLength);
             Assert.Equal(1, schema.Properties["ArrayWithLength"].MinItems);
             Assert.Equal(3, schema.Properties["ArrayWithLength"].MaxItems);
-            Assert.Equal(true, schema.Properties["IntWithRange"].ExclusiveMinimum);
-            Assert.Equal(true, schema.Properties["IntWithRange"].ExclusiveMaximum);
             Assert.Equal("byte", schema.Properties["StringWithBase64"].Format);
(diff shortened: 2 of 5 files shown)


test/Swashbuckle.AspNetCore.Newtonsoft.Test/SchemaGenerator/NewtonsoftSchemaGeneratorTests.cs
```diff
@@ -322,10 +322,15 @@ public void GenerateSchema_SetsValidationProperties_IfComplexTypeHasValidationAt
...
-            Assert.Equal(true, schema.Properties["IntWithRange"].ExclusiveMinimum);
-            Assert.Equal(true, schema.Properties["IntWithRange"].ExclusiveMaximum);
             Assert.Equal("byte", schema.Properties["StringWithBase64"].Format);
             Assert.Equal("string", schema.Properties["StringWithBase64"].Type);
+            Assert.Null(schema.Properties["IntWithRange"].ExclusiveMinimum);
+            Assert.Null(schema.Properties["IntWithRange"].ExclusiveMaximum);
+            Assert.Equal(true, schema.Properties["IntWithExclusiveRange"].ExclusiveMinimum);
+            Assert.Equal(true, schema.Properties["IntWithExclusiveRange"].ExclusiveMaximum);
+#else
+            Assert.Null(schema.Properties["IntWithRange"].ExclusiveMinimum);
```
martincostello, line 332: Are these two assertions in both branches? If so, just have them after the `#endif`

test/Swashbuckle.AspNetCore.SwaggerGen.Test/SchemaGenerator/JsonSerializerSchemaGeneratorTests.cs
```diff
@@ -345,10 +345,15 @@ public void GenerateSchema_SetsValidationProperties_IfComplexTypeHasValidationAt
             Assert.Equal(3, schema.Properties["StringWithLength"].MaxLength);
             Assert.Equal(1, schema.Properties["ArrayWithLength"].MinItems);
             Assert.Equal(3, schema.Properties["ArrayWithLength"].MaxItems);
-            Assert.Equal(true, schema.Properties["IntWithRange"].ExclusiveMinimum);
-            Assert.Equal(true, schema.Properties["IntWithRange"].ExclusiveMaximum);
             Assert.Equal("byte", schema.Properties["StringWithBase64"].Format);
             Assert.Equal("string", schema.Properties["StringWithBase64"].Type);
+            Assert.Null(schema.Properties["IntWithRange"].ExclusiveMinimum);
```
martincostello, line 350: Same here.

src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/OpenApiSchemaExtensions.cs
```diff
@@ -183,8 +183,15 @@ private static void ApplyRangeAttribute(OpenApiSchema schema, RangeAttribute ran
         {
 #if NET8_0_OR_GREATER
 
-            schema.ExclusiveMinimum = rangeAttribute.MinimumIsExclusive;
-            schema.ExclusiveMaximum = rangeAttribute.MaximumIsExclusive;
+            if(rangeAttribute.MinimumIsExclusive)
```
martincostello, line 186:
```suggestion
            if (rangeAttribute.MinimumIsExclusive)
```

src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/OpenApiSchemaExtensions.cs
```diff
@@ -183,8 +183,15 @@ private static void ApplyRangeAttribute(OpenApiSchema schema, RangeAttribute ran
...
+            {
+                schema.ExclusiveMinimum = true;
+            }
+
+            if(rangeAttribute.MaximumIsExclusive)
```
martincostello, line 191:
```suggestion
            if (rangeAttribute.MaximumIsExclusive)
```

test/Swashbuckle.AspNetCore.SwaggerGen.Test/SchemaGenerator/JsonSerializerSchemaGeneratorTests.cs
jgarciadelanoceda: true!

src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/OpenApiSchemaExtensions.cs
jgarciadelanoceda: I do sorry!. Usually VS formats the code, but it does not seem to format it inside Compile conditions

martincostello (approved):

Issue #2912, opened 2024-05-21
[Bug]: "InvalidOperationException: Sequence contains more than one matching element" in Swagger when using multiple [FromForm] parameters in Minimal API


marcominerva:
### Describe the bug

I have originally posted the issue in the ASP.NET Core repo: https://github.com/dotnet/aspnetcore/issues/53831. See in particular this comment: https://github.com/dotnet/aspnetcore/issues/53831#issuecomment-1930720704

### Expected behavior

Swagger should not throw an exception.

### Actual behavior

An `InvalidOperationException` is thrown.

### Steps to reproduce

Minimal repro here: https://github.com/marcominerva/FromFormBindingIssue

### Exception(s) (if any)

InvalidOperationException: Sequence contains more than one matching element
System.Linq.ThrowHelper.ThrowMoreThanOneMatchException()
System.Linq.Enumerable.TryGetSingle(IEnumerable source, Func<TSource, bool> predicate, out bool found)
System.Linq.Enumerable.SingleOrDefault(IEnumerable source, Func<TSource, bool> predicate)
Swashbuckle.AspNetCore.SwaggerGen.SwaggerGenerator.GenerateOpenApiOperationFromMetadata(ApiDescription apiDescription, SchemaRepository schemaRepository)
Swashbuckle.AspNetCore.SwaggerGen.SwaggerGenerator.GenerateOperation(ApiDescription apiDescription, SchemaRepository schemaRepository)
Swashbuckle.AspNetCore.SwaggerGen.SwaggerGenerator.GenerateOperations(IEnumerable apiDescriptions, SchemaRepository schemaRepository)
Swashbuckle.AspNetCore.SwaggerGen.SwaggerGenerator.GeneratePaths(IEnumerable apiDescriptions, SchemaRepository schemaRepository)
Swashbuckle.AspNetCore.SwaggerGen.SwaggerGenerator.GetSwaggerDocumentWithoutFilters(string documentName, string host, string basePath)
Swashbuckle.AspNetCore.SwaggerGen.SwaggerGenerator.GetSwaggerAsync(string documentName, string host, string basePath)
Swashbuckle.AspNetCore.Swagger.SwaggerMiddleware.Invoke(HttpContext httpContext, ISwaggerProvider swaggerProvider)
Microsoft.AspNetCore.Authorization.AuthorizationMiddleware.Invoke(HttpContext context)
[...]

PR #2962, opened 2024-06-22
AsParameters throwing error on cast when showing the Description with EnableAnnotations

jgarciadelanoceda:
This fixes an issue shown when investigating #2658. Not sure if this fixes it all.

Tested with the project WebApi, if you do not apply the GetCustomAttributes to GetCustomAttribute change then you will see an error on the SwaggerUi based on a Cast.

As the SwaggerParameter is an attribute that only can be set once per property there is no issue on the change
I have studied if inserting a test in Swashbuckle.AspNetCore.Annotations.Test, but the existing test are hitting this change, so I think there is no need to add it(In fact I have not seen a test using MinimalApi approach)

PR #2963, opened 2024-06-24
Fix RequestBodyFilterAnnotation and MultipleFromForm for MinimalApi 

### LINKED ISSUE #2662: SwaggerRequestBodyAttribute Description have no effect in Minimal Api
**Minimal API**

I'm using Swashbuckle.AspNetCore 6.5.0 with Swashbuckle.AspNetCore.Annotations 6.5.0 and have added a SwaggerRequestBodyAttribute to a response class thus:

```c#
app.MapPost("/todos", ([FromBody, SwaggerRequestBody("Todo object to be created", Required = true)] Todo todo) =>
    {
        todos.Add(todo with { Id = Guid.NewGuid() });
        return TypedResults.Ok(todo);
    })
    .WithOpenApi(operation => new(operation)
    {
        Summary = "Create a todo",
        OperationId = "createTodo",
        Tags = new List<OpenApiTag> { new() { Name = "Todos" } },
        //Parameters = new List<OpenApiParameter>
        //{
        //    new OpenApiParameter{ Description ="Todo object to be created" }
        //}
    });
```

The project docs say I can use this to "to enrich the corresponding Schema metadata that's generated by Swashbuckle", but the generated swagger.json doesn't contain "MyTitle" or "MyDescription" at all, only:

```json
"post": {
    "tags": [
      "Todos"
    ],
    "summary": "Create a todo",
    "operationId": "createTodo",
    "requestBody": {
      "content": {
        "application/json": {
          "schema": {
            "$ref": "#/components/schemas/Todo"
          }
        }
      },
      "required": true
    },
    "responses": {
      "200": {
        "description": "OK",
        "content": {
          "application/json": {
            "schema": {
              "$ref": "#/components/schemas/Todo"
            }
          }
        }
      }
    }
  }
```

My configurations

```c#
builder.Services.AddSwaggerGen(options =>
{
    options.SwaggerDoc("v1", new OpenApiInfo { Title = "ChatGPT Plugin", Version = "v1" });
    options.DocInclusionPredicate((docName, apiDesc) => true);
    options.EnableAnnotations();
});
```

jgarciadelanoceda:
Fixes #2662. Very similar to #2962 PR. When the RequestBody had a SwaggerRequestBody attribute the attribute was not applying and it provoked an exception

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
@@ -422,9 +422,8 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
                 {
                     var requestParameters = apiDescription.ParameterDescriptions.Where(desc => desc.IsFromBody() || desc.IsFromForm());
                     var countOfParameters = requestParameters.Count();
-                    if (countOfParameters >= 1)
+                    if (countOfParameters > 0)
                     {
-                        _ = requestParameters.SingleOrDefault(desc => desc.IsFromBody());
                         if (countOfParameters == 1)
                         {
                             var requestParameter = requestParameters.First();
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
@@ -421,16 +421,18 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
                 foreach (var content in requestContentTypes)
                 {
                     var requestParameters = apiDescription.ParameterDescriptions.Where(desc => desc.IsFromBody() || desc.IsFromForm());
-                    if (requestParameters.Any())
+                    var countOfParameters = requestParameters.Count();
+                    if (countOfParameters >= 1)
                     {
                         _ = requestParameters.SingleOrDefault(desc => desc.IsFromBody());
-                        if (requestParameters.Count() == 1)
+                        if (countOfParameters == 1)
                         {
+                            var requestParameter = requestParameters.First();
                             content.Schema = GenerateSchema(
-                                requestParameters.First().ModelMetadata.ModelType,
+                                requestParameter.ModelMetadata.ModelType,
                                 schemaRepository,
-                                requestParameters.First().PropertyInfo(),
-                                requestParameters.First().ParameterInfo());
+                                requestParameter.PropertyInfo(),
+                                requestParameter.ParameterInfo());
                         }
                         else
                         {
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
@@ -421,8 +421,9 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
                 foreach (var content in requestContentTypes)
                 {
                     var requestParameters = apiDescription.ParameterDescriptions.Where(desc => desc.IsFromBody() || desc.IsFromForm());
-                    if (requestParameters is not null)
+                    if (requestParameters.Any())
                     {
+                        _ = requestParameters.SingleOrDefault(desc => desc.IsFromBody());
                         if (requestParameters.Count() == 1)
                         {
                             content.Schema = GenerateSchema(
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
@@ -420,14 +420,31 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
             {
                 foreach (var content in requestContentTypes)
                 {
-                    var requestParameter = apiDescription.ParameterDescriptions.SingleOrDefault(desc => desc.IsFromBody() || desc.IsFromForm());
-                    if (requestParameter is not null)
+                    var requestParameters = apiDescription.ParameterDescriptions.Where(desc => desc.IsFromBody() || desc.IsFromForm());
+                    if (requestParameters is not null)
                     {
-                        content.Schema = GenerateSchema(
-                            requestParameter.ModelMetadata.ModelType,
-                            schemaRepository,
-                            requestParameter.PropertyInfo(),
-                            requestParameter.ParameterInfo());
+                        if (requestParameters.Count() == 1)
+                        {
+                            content.Schema = GenerateSchema(
+                                requestParameters.First().ModelMetadata.ModelType,
+                                schemaRepository,
+                                requestParameters.First().PropertyInfo(),
+                                requestParameters.First().ParameterInfo());
+                        }
+                        else
+                        {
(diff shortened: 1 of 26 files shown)


martincostello (commented):
Could you add a test that shows the description in the attribute is rendered in the generated document please?

The test change here only checks it doesn't cause generation to fail.

jgarciadelanoceda:
It's tested in the test Apply_EnrichesRequestBodyMetadata_IfControllerParameterDecoratedWithSwaggerRequestBodyAttribute of the file AnnotationsRequestBodyFilterTests

martincostello:
I feel we're still missing _something_ in our coverage here, as if we were validating the generated content for a Minimal API using this we would have found this bug before now.

jgarciadelanoceda:
Yeah.. I do agree, altough there is no test for any MinimalApi (Apart from the IntegrationTests). I do not know how to fit this missing tests in the library(You cannot do a MapGet if there is no app. Could we create an issue to test the different apis using MinimalApi

martincostello:
A verification test from the websites should fit this sort of niche.

If we had some in this area already, your addition would have caused a test to fail as the description would have changed, even if the original bug wasn't there.

That suggests to me adding such a test would give us the coverage that the document is what we expect and draw attention to future changes to ensure they are intentional.

jgarciadelanoceda:
Ok, I will add Verify test to the webSites and try to fit the current MinimalApi issues I found over Swashbuckle.

jgarciadelanoceda:
This last commit also fixes #2912.
Added test for the diferent MinimalApi method documentation

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
```diff
@@ -420,14 +420,31 @@ private OpenApiOperation GenerateOpenApiOperationFromMetadata(ApiDescription api
             {
                 foreach (var content in requestContentTypes)
                 {
-                    var requestParameter = apiDescription.ParameterDescriptions.SingleOrDefault(desc => desc.IsFromBody() || desc.IsFromForm());
-                    if (requestParameter is not null)
+                    var requestParameters = apiDescription.ParameterDescriptions.Where(desc => desc.IsFromBody() || desc.IsFromForm());
+                    if (requestParameters is not null)
```
martincostello, line 424: `Where()` will never return `null`.

test/WebSites/WebApi/Program.cs
```diff
@@ -1,60 +1,43 @@
...
+    c.IncludeXmlComments(Assembly.GetExecutingAssembly());
     c.SwaggerDoc("v1", new() { Title = "WebApi", Version = "v1" });
 });
-
 var app = builder.Build();
-
 app.UseSwagger();
 app.UseSwaggerUI();
 
+
```
martincostello, line 22:
```suggestion
```

test/WebSites/WebApi/EndPoints/OpenApiEndpoints.cs
```diff
@@ -0,0 +1,57 @@
...
+            group.MapPost("/api/people-minimalapi", ([FromForm] Person person, [FromForm] Address address) =>
+            {
+                TypedResults.NoContent();
+            })
+            .WithOpenApi()
+            .DisableAntiforgery();
+
+
+
+            //group.MapGet("DateTimeKind", (DateTimeKind d) => d);
```
martincostello, line 45:
```suggestion
```

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
```diff
@@ -421,8 +421,9 @@ private OpenApiOperation GenerateOpenApiOperationFromMetadata(ApiDescription api
                 foreach (var content in requestContentTypes)
                 {
                     var requestParameters = apiDescription.ParameterDescriptions.Where(desc => desc.IsFromBody() || desc.IsFromForm());
-                    if (requestParameters is not null)
```
martincostello: This method is inefficient the way it's using LINQ.

First there's a check for "is there at least one", then there's a check for "is there exactly one" that is discarded, then they're counted, then there's either a `ToList()` if there isn't exactly one or `First()` is called three times to get the same value.

This can be refactored to be smarter.

jgarciadelanoceda: Done, I am not sure if I can remove the discard (To have the same behaviour as before)

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
```diff
@@ -421,16 +421,18 @@ private OpenApiOperation GenerateOpenApiOperationFromMetadata(ApiDescription api
                 foreach (var content in requestContentTypes)
                 {
                     var requestParameters = apiDescription.ParameterDescriptions.Where(desc => desc.IsFromBody() || desc.IsFromForm());
-                    if (requestParameters.Any())
+                    var countOfParameters = requestParameters.Count();
+                    if (countOfParameters >= 1)
                     {
                         _ = requestParameters.SingleOrDefault(desc => desc.IsFromBody());
```
martincostello, line 427: What does this achieve given the result is discarded?

jgarciadelanoceda: It will throw when there is more than one Body document, as it was doing previously. It does not make sense to have more than 1 Body element

jgarciadelanoceda:
The build server seems to have a file corrupted :(

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
martincostello: Computing it as a side-effect to throw a not-very-helpful exception message doesn't seem to be desirable behaviour to me. If we want to stop this if it occurs, then we should explicitly check for it and then throw a useful exception, rather than a vague "Sequence contains more than one element" exception.

jgarciadelanoceda: Finally I tried several ways to introduce several FromBody (Passing AsParameters and then FromBody or putting FromBody in just one element and a exception is thrown before reaching Swashbucle code (A binding error)

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
```diff
@@ -420,14 +420,33 @@ private OpenApiOperation GenerateOpenApiOperationFromMetadata(ApiDescription api
...
+                                AllOf = requestParameters.Select(s =>
+                                    GenerateSchema(
+                                    s.ModelMetadata.ModelType,
+                                    schemaRepository,
+                                    s.PropertyInfo(),
+                                    s.ParameterInfo()))
+                                .ToList()
+                            };
+                        }
+
```
martincostello, line 449:
```suggestion
```

martincostello (approved):

jgarciadelanoceda:
@martincostello  I do not know how the release notes are going to be generated (For the version 6.7.0) but the description of [Fix RequestBodyFilterAnnotation for Minimal API) is missing the multiple FromForms part

martincostello:
They're auto-generated by GitHub and then I'll manually edit anything missing/incorrect/

PR #2980, opened 2024-07-10
Use apiParameter.Type when it does not lie

### LINKED ISSUE #2733: [AsParameter] attribute generates a string query parameter when it should generate a decimal
**Issue**
The `[AsParameters]` attribute is not generating the correct type of `decimal` when using Swashbuckle to generate a WebApi. It generates a `string`.

When a decimal type is used on the object that is annotated with the `[AsParameters]` attribute the type generated is always a `string` rather than a decimal.

**Reproduction**
I have attached a zip file with a net7.0 solution with a basic replication of the problem. It contains a net7 minimal web API and you need to run it to see the type generated on the `/swagger` URL page is a string, not a decimal as expected.

[TestWebApp.zip](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/files/13265006/TestWebApp.zip)

**Expected Behaviour**
Generate a decimal when the client runs the build step.

jgarciadelanoceda:
This PR fixes #2733.
It's a continuation of #2532 but with the VerifyTest running over IntegrationTest.

The issue is that the type for MvcWithNullable was coming as string(And its a LogLevel)
And the modelMetadata.ModelType that is coming from SwaggerAnnotationsEndpoints was coming as string.

So.. The best of both worlds

### DIFF
--- a/test/WebSites/WebApi/EndPoints/SwaggerAnnotationsEndpoints.cs
+++ b/test/WebSites/WebApi/EndPoints/SwaggerAnnotationsEndpoints.cs
@@ -16,6 +16,11 @@ namespace WebApi.EndPoints
 
             group.MapPost("/fruit/{id}", CreateFruit);
 
+            group.MapGet("/AsParameters", ([AsParameters] AsParametersRecord record) =>
+            {
+                return record;
+            });
+
             return app;
         }
 
@@ -29,4 +34,11 @@ namespace WebApi.EndPoints
 
     [SwaggerSchema("Description for Schema")]
     record Fruit(string Name);
+
+    record class AsParametersRecord([SwaggerParameter(Description = "Description")] Guid? paramOne, Guid paramTwo,
+        DateTime? paramThree, DateTime paramFour,
+        DateOnly? paramFive, DateOnly paramSix,
+        TimeOnly? paramSeven, TimeOnly paramEight,
+        DateTimeKind? paramNine, DateTimeKind paramTen,
+        decimal? paramEleven, decimal paramTwelve);
 }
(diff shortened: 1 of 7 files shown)


jgarciadelanoceda:
Draft

jgarciadelanoceda:
I was checking if the Parameters of the Controller when the type is Nullable<T>, we put nullable on the property, and as we are not doing it (For whatever reason). For MinimalApis it has the same behaviour

test/WebSites/WebApi/EndPoints/SwaggerAnnotationsEndpoints.cs
```diff
@@ -29,4 +34,11 @@ record struct CreateFruitModel
 
     [SwaggerSchema("Description for Schema")]
     record Fruit(string Name);
+
+    record class AsParametersRecord([SwaggerParameter(Description = "Description")] Guid? paramOne, Guid paramTwo,
+        DateTime? paramThree, DateTime paramFour,
+        DateOnly? paramFive, DateOnly paramSix,
+        TimeOnly? paramSeven, TimeOnly paramEight,
+        DateTimeKind? paramNine, DateTimeKind paramTen,
+        decimal? paramEleven, decimal paramTwelve);
```
martincostello, line 43:
```suggestion
    record class AsParametersRecord(
        [SwaggerParameter(Description = "Description")] Guid? paramOne,
        Guid paramTwo,
        DateTime? paramThree,
        DateTime paramFour,
        DateOnly? paramFive,
        DateOnly paramSix,
        TimeOnly? paramSeven,
        TimeOnly paramEight,
        DateTimeKind? paramNine,
        DateTimeKind paramTen,
        decimal? paramEleven,
        decimal paramTwelve);
```

martincostello (approved):

Issue #2655, opened 2023-05-27
Doesn't generate summary doc when the controller class has overload methods.


codelovercc:
Swashbuckle.AspNetCore 6.5.0
.net 7.0

Project.csproj
```
        <GenerateDocumentationFile>true</GenerateDocumentationFile>
```

Controller class:

```csharp
[ApiController]
[Route("api/[controller]")]
public class MyController : MyBaseController<MyModel>
{
    public MyController(MyBuzService service):base(service){}
   
    /// <summary>
    /// Find by single id
    /// </summary>
    /// <param name="id">ID</param>
    /// <returns></returns>
    [HttpGet]
    [Route("Find/{id}")]
   public async Task<IActionResult> Find([FromRoute] string id) => await Find(new[] { id });

    /// <summary>
    /// Find by multi id
    /// </summary>
    /// <param name="id">ID</param>
    /// <returns></returns>
    [HttpPost]
    [Route("Find")]
   public async Task<IActionResult> Find([FromQuery] [FromBody] string[] id)
   {
        return Ok(await Service.GetModelAsync(id));
   }
}

public abstract class MyBaseController<TModel>
{
    private MyBuzService Service;
    public MyController(MyBuzService service)
   {
        Service = service;
   }
   

    /// <summary>
    /// Add some models
    /// </summary>
    /// <param name="models">Models to add</param>
    /// <returns></returns>
   [HttpPost]
    [Route("AddRange")]
   public async Task<IActionResult> AddRange([FromBody] TModel[] models)
   {
       return Ok(await Service.AddRange(models));
   }
}
```

I've checked the generated XML documentation file, it's generated correctly. 
But summary document are missing in the swagger.json.
I've created a class that implements `IOperationFilter` interface, when I'm debugging, I found `OpenApiOperation.Summary` property with value `null`.

update issue:
The method `public async Task<IActionResult> AddRange([FromBody] TModel[] models)`  takes one parameter of TModel[] type which is array type doesn't generate summary doc too!.

Naxaliav:
@codelovercc have you configured to read .xml file? :) 
![image](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/assets/33034058/2f6bff61-519f-4443-88e2-1b32611e9440)

codelovercc:
[...]

PR #2982, opened 2024-07-14
Document arrays of GenericParameters with XmlComments

jgarciadelanoceda:
It fixes an issue found while analyzing #2655 

I am hesitant to fix the other issue found (When there are 2 methods on a Controller with the same number of parameters, because you will have to make assumptions here. The base class has to has the firstParameter as TResource and do not have more parameters.. Because I tried to do the following:
image

But it seems extremelly over complicated and I do not like it very much, the issue is that the XML only has the comments for the baseClass and when searching for the Method, SwashBuckle is searching by the Name, and the number of parameters inside the file MethodInfoExtensions

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsNodeNameHelper.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsNodeNameHelper.cs
@@ -50,7 +50,11 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
         private static string QualifiedNameFor(Type type, bool expandGenericArgs = false)
         {
             if (type.IsArray)
-                return $"{QualifiedNameFor(type.GetElementType(), expandGenericArgs)}[]";
+            {
+                var elementType = type.GetElementType();
+                return elementType.IsGenericParameter ? $"`{elementType.GenericParameterPosition}[]" : $"{QualifiedNameFor(type.GetElementType(), expandGenericArgs)}[]";
+            }
+
 
             var builder = new StringBuilder();
 
--- a/test/Swashbuckle.AspNetCore.IntegrationTests/SwaggerVerifyIntegrationTest.SwaggerEndpoint_ReturnsValidSwaggerJson_startupType=GenericControllers.Startup_swaggerRequestUri=v1.verified.txt
+++ b/test/Swashbuckle.AspNetCore.IntegrationTests/SwaggerVerifyIntegrationTest.SwaggerEndpoint_ReturnsValidSwaggerJson_startupType=GenericControllers.Startup_swaggerRequestUri=v1.verified.txt
@@ -57,6 +57,80 @@
             }
           }
         }
+      },
+      "delete": {
+        "tags": [
+          "Orders"
+        ],
+        "summary": "Delete by Ids",
+        "parameters": [
+          {
+            "name": "tennantId",
+            "in": "path",
+            "required": true,
+            "schema": {
+              "type": "string"
+            }
+          }
+        ],
+        "requestBody": {
+          "description": "deleting Ids",
+          "content": {
+            "application/json": {
+              "schema": {
+                "type": "array",
+                "items": {
+                  "$ref": "#/components/schemas/Order"
+                }
+              }
+            },
+            "text/json": {
+              "schema": {
+                "type": "array",
+                "items": {
+                  "$ref": "#/components/schemas/Order"
+                }
+              }
+            },
+            "application/*+json": {
+              "schema": {
+                "type": "array",
+                "items": {
+                  "$ref": "#/components/schemas/Order"
+                }
+              }
+            }
+          },
+          "required": true
+        },
+        "responses": {
+          "200": {
+            "description": "Deleted",
+            "content": {
+              "text/plain": {
+                "schema": {
+                  "type": "integer",
+                  "format": "int32"
+                }
+              },
+              "application/json": {
+                "schema": {
+                  "type": "integer",
+                  "format": "int32"
+                }
+              },
+              "text/json": {
+                "schema": {
+                  "type": "integer",
+                  "format": "int32"
+                }
+              }
+            }
+          },
+          "404": {
+            "description": "Failed"
+          }
+        }
       }
     },
     "/{tennantId}/products": {
@@ -109,6 +183,80 @@
                 }
               }
             }
+          }
+        }
+      },
+      "delete": {
+        "tags": [
+          "Products"
+        ],
+        "summary": "Delete by Ids",
+        "parameters": [
+          {
+            "name": "tennantId",
+            "in": "path",
+            "required": true,
+            "schema": {
+              "type": "string"
+            }
+          }
+        ],
+        "requestBody": {
+          "description": "deleting Ids",
+          "content": {
+            "application/json": {
+              "schema": {
+                "type": "array",
+                "items": {
+                  "$ref": "#/components/schemas/Product"
+                }
+              }
+            },
+            "text/json": {
+              "schema": {
+                "type": "array",
+                "items": {
+                  "$ref": "#/components/schemas/Product"
+                }
+              }
+            },
+            "application/*+json": {
+              "schema": {
+                "type": "array",
+                "items": {
+                  "$ref": "#/components/schemas/Product"
+                }
+              }
+            }
+          },
+          "required": true
+        },
+        "responses": {
+          "200": {
+            "description": "Deleted",
+            "content": {
+              "text/plain": {
+                "schema": {
+                  "type": "integer",
+                  "format": "int32"
+                }
+              },
+              "application/json": {
+                "schema": {
+                  "type": "integer",
+                  "format": "int32"
+                }
+              },
+              "text/json": {
+                "schema": {
+                  "type": "integer",
+                  "format": "int32"
+                }
+              }
+            }
+          },
+          "404": {
+            "description": "Failed"
           }
         }
       }
--- a/test/WebSites/GenericControllers/Controllers/GenericResourceController.cs
+++ b/test/WebSites/GenericControllers/Controllers/GenericResourceController.cs
@@ -1,6 +1,5 @@
-﻿using System.Collections.Generic;
-using System.ComponentModel.DataAnnotations;
-using Microsoft.AspNetCore.Http;
+﻿using System.ComponentModel.DataAnnotations;
+using System.Threading;
 using Microsoft.AspNetCore.Mvc;
 
 namespace GenericControllers.Controllers
@@ -11,11 +10,26 @@ namespace GenericControllers.Controllers
         /// Creates a resource
         /// </summary>
         /// <param name="resource">The resource</param>
+        /// <param name="cancellationToken"></param>
         /// <returns></returns>
         [HttpPost]
         [ProducesResponseType(201)]
         [Consumes("application/json")]
-        public int Create([FromBody, Required]TResource resource)
+        public int Create([FromBody, Required] TResource resource, CancellationToken cancellationToken)
+        {
+            return 1;
+        }
+
+        /// <summary>
+        /// Delete by Ids
+        /// </summary>
+        /// <param name="ids">deleting Ids</param>
+        /// <param name="cancellationToken"></param>
+        /// <returns></returns>
+        /// <response code="200">Deleted</response>
+        /// <response code="404">Failed</response>
+        [HttpDelete("")]
+        public virtual int Delete([Required, FromBody] TResource[] ids, CancellationToken cancellationToken)
         {
             return 1;
         }


jgarciadelanoceda:
Finally I added support for Overload methods and test it with the IntegrationTests

src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/MethodInfoExtensions.cs
```diff
@@ -9,13 +9,65 @@ public static MethodInfo GetUnderlyingGenericTypeMethod(this MethodInfo construc
...
+            var genericArguments = constructedType.GenericTypeArguments;
 
             // Retrieve list of candidate methods that match name and parameter count
             var candidateMethods = genericTypeDefinition.GetMethods()
                 .Where(m =>
                 {
-                    return (m.Name == constructedTypeMethod.Name)
-                        && (m.GetParameters().Length == constructedTypeMethod.GetParameters().Length);
+                    var genericTypeDefinitionParameters = m.GetParameters();
+                    var constructedTypeParameters = constructedTypeMethod.GetParameters();
```
martincostello, line 19: `constructedTypeParameters` can be hoisted outside of the LINQ method so it isn't computed more than once.

src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/MethodInfoExtensions.cs
```diff
@@ -9,13 +9,65 @@ public static MethodInfo GetUnderlyingGenericTypeMethod(this MethodInfo construc
...
+                            {
+                                var genericTypeDefinitionArguments = genericTypeDefinitionParameters[i].ParameterType.GetGenericArguments();
+                                var constructedDefinitionArguments = constructedTypeParameters[i].ParameterType.GetGenericArguments();
+                                if (genericTypeDefinitionArguments.Length != constructedDefinitionArguments.Length)
+                                {
+                                    return false;
+                                }
+                                for (var j = 0; j < genericTypeDefinitionArguments.Length; j++)
+                                {
+
```
martincostello, line 47:
```suggestion
```

src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/MethodInfoExtensions.cs
```diff
@@ -9,13 +9,65 @@ public static MethodInfo GetUnderlyingGenericTypeMethod(this MethodInfo construc
         {
             var constructedType = constructedTypeMethod.DeclaringType;
             var genericTypeDefinition = constructedType.GetGenericTypeDefinition();
+            var genericArguments = constructedType.GenericTypeArguments;
 
             // Retrieve list of candidate methods that match name and parameter count
             var candidateMethods = genericTypeDefinition.GetMethods()
                 .Where(m =>
```
martincostello, line 16: This method has gotten pretty complicated - feels like it could use with some dedicated unit tests, rather than just relying on the (non-exhaustive) verification tests.

jgarciadelanoceda: Yeah.. If you see the logic, it's pretty much the logic that was previously with the exception that right now you distinguish between two methods with the same number of parameters, but different kind. I'll do a UnitTest, but even for this simple logic the UnitTest is going to be hard

jgarciadelanoceda: Done!

jgarciadelanoceda:
The requested test were created

test/Swashbuckle.AspNetCore.SwaggerGen.Test/Fixtures/FakeGenericResourceController.cs
```diff
@@ -0,0 +1,24 @@
...
+        where T : class
+    {
+        public virtual int DifferentMethodsSignatures([FromBody, Required] T t, CancellationToken cancellationToken) => 1;
+        public virtual int DifferentMethodsSignatures([FromBody, Required] string key, T t) => 1;
+        public virtual int DifferentMethodsSignatures([FromBody, Required] T[] ts) => 1;
+        public virtual int DifferentMethodsSignatures([FromBody, Required] string[] strings) => 1;
+        public virtual int DifferentMethodsSignatures([FromBody, Required] List<T> ts) => 1;
+        public virtual int DifferentMethodsSignatures([FromBody, Required] List<string> strings) => 1;
+        public virtual int DifferentMethodsSignatures([FromBody, Required] Dictionary<int, T> ts) => 1;
+        public virtual int DifferentMethodsSignatures([FromBody, Required] IEnumerable<T> ts) => 1;
```
martincostello, line 18: Why virtual?

test/Swashbuckle.AspNetCore.SwaggerGen.Test/Fixtures/FakeGenericResourceController.cs
```diff
@@ -0,0 +1,24 @@
...
+    }
+    internal class NonGenericResourceController : FakeGenericResourceController<XmlAnnotatedRecord>
+    {
+
+    }
```
martincostello, line 23:
```suggestion

    internal class NonGenericResourceController : FakeGenericResourceController<XmlAnnotatedRecord>;
```

test/Swashbuckle.AspNetCore.SwaggerGen.Test/Fixtures/FakeGenericResourceController.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
+using System.Threading;
+using Microsoft.AspNetCore.Mvc;
+
+namespace Swashbuckle.AspNetCore.SwaggerGen.Test.Fixtures
```
martincostello, line 6: For new files you've added, could use use file-scoped namespaces?

test/Swashbuckle.AspNetCore.SwaggerGen.Test/XmlComments/XmlMethodInfoExtensionsTests.cs
```diff
@@ -0,0 +1,23 @@
+﻿using System.Linq;
+using System.Reflection;
+using Swashbuckle.AspNetCore.SwaggerGen.Test.Fixtures;
+using Xunit;
+
+namespace Swashbuckle.AspNetCore.SwaggerGen.Test
+{
+    public class XmlMethodInfoExtensionsTests
+    {
+        [Theory, MemberData(nameof(DifferentMethodsSignatures))]
```
martincostello, line 10:
```suggestion
        [Theory]
        [MemberData(nameof(DifferentMethodsSignatures))]
```

test/Swashbuckle.AspNetCore.SwaggerGen.Test/XmlComments/XmlMethodInfoExtensionsTests.cs
```diff
@@ -0,0 +1,23 @@
...
+        public void DifferentMethodsSignatures_ShouldNotBeNull(MethodInfo methodInfo)
+        {
+            Assert.NotNull(methodInfo.GetUnderlyingGenericTypeMethod());
```
martincostello, line 13: Is there anything meaningful to assert on the result other than it doesn't throw and isn't null? Feels like this could return the wrong thing and still pass.

jgarciadelanoceda: Applied a strategy to change the parameterNames of every method and the result types to be different

test/Swashbuckle.AspNetCore.SwaggerGen.Test/Fixtures/FakeGenericResourceController.cs
jgarciadelanoceda: No need

test/Swashbuckle.AspNetCore.SwaggerGen.Test/Fixtures/FakeGenericResourceController.cs
```diff
@@ -0,0 +1,20 @@
...
+    public byte DifferentMethodsSignatures([FromBody, Required] T t, CancellationToken cancellationToken) => default;
+    public sbyte DifferentMethodsSignatures([FromBody, Required] string key, T t) => default;
+    public short DifferentMethodsSignatures([FromBody, Required] T[] arrayOfTs) => default;
+    public ushort DifferentMethodsSignatures([FromBody, Required] string[] arrayOfStrings) => default;
+    public int DifferentMethodsSignatures([FromBody, Required] List<T> listOfTs) => default;
+    public uint DifferentMethodsSignatures([FromBody, Required] List<string> listOfStrings) => default;
+    public long DifferentMethodsSignatures([FromBody, Required] Dictionary<int, T> dictionaryOfTs) => default;
+    public ulong DifferentMethodsSignatures([FromBody, Required] IEnumerable<T> iEnumerableOfTs) => default;
+}
+internal class NonGenericResourceController : FakeGenericResourceController<XmlAnnotatedRecord>;
```
martincostello, line 20:
```suggestion

internal class NonGenericResourceController : FakeGenericResourceController<XmlAnnotatedRecord>;
```

test/Swashbuckle.AspNetCore.SwaggerGen.Test/XmlComments/XmlMethodInfoExtensionsTests.cs
```diff
@@ -0,0 +1,28 @@
...
+        Assert.Equal(methodInfo.GetParameters().Select(s => s.Name), underlyingGenericMethodParameters.Select(s => s.Name));
+        Assert.Equal(methodInfo.ReturnParameter.ParameterType, underlyingGenericMethod.ReturnParameter.ParameterType);
+    }
+
+    public static TheoryData<MethodInfo> DifferentMethodsSignatures =>
+        new(typeof(NonGenericResourceController).GetMethods()
+            .Where(s =>
+            {
+                return s.Name == nameof(NonGenericResourceController.DifferentMethodsSignatures);
+            }));
```
martincostello, line 27:
```suggestion
            .Where(s => s.Name == nameof(NonGenericResourceController.DifferentMethodsSignatures)));
```

test/Swashbuckle.AspNetCore.SwaggerGen.Test/XmlComments/XmlMethodInfoExtensionsTests.cs
```diff
@@ -0,0 +1,28 @@
...
+    [Theory]
+    [MemberData(nameof(DifferentMethodsSignatures))]
+    public void DifferentMethodsSignatures_ShouldBeExpected(MethodInfo methodInfo)
+    {
+        var underlyingGenericMethod = methodInfo.GetUnderlyingGenericTypeMethod();
+        Assert.NotNull(underlyingGenericMethod);
+        var underlyingGenericMethodParameters = underlyingGenericMethod.GetParameters();
+        Assert.NotEmpty(underlyingGenericMethodParameters);
+        Assert.Equal(methodInfo.GetParameters().Select(s => s.Name), underlyingGenericMethodParameters.Select(s => s.Name));
+        Assert.Equal(methodInfo.ReturnParameter.ParameterType, underlyingGenericMethod.ReturnParameter.ParameterType);
```
martincostello, line 19: I think it would be better to explicitly return the expected names and types as part of the theory data. Otherwise you're effectively using the same logic as the code you're testing to compute what the expected results should be.

jgarciadelanoceda: The parameters names are not checked in the GetUnderlyingGenericTypeMethod, I am checking that the parameters are the expetect(That are the ones that come from the methodInfo).
I can only check the parametersNames in this case, because the parameterTypes in the case of the underlyingGenericMethod are generic

jgarciadelanoceda: @martincostello I keep it open in case you need a change here

test/Swashbuckle.AspNetCore.SwaggerGen.Test/Fixtures/FakeGenericResourceController.cs
jgarciadelanoceda: @martincostello, we could apply an editorConfig setting in order to get this automatically:
csharp_style_namespace_declarations = file_scoped:suggestion

test/Swashbuckle.AspNetCore.SwaggerGen.Test/XmlComments/XmlMethodInfoExtensionsTests.cs
martincostello: What I'm getting at is that you're dynamically computing what you expect the result to be.

Instead of would be better for you to explicitly reference the methods to get the `MethodInfo`s and then hard-code what you expect the names and types to be, so for example:

```csharp
[firstMethodInfo, ["t", "cancellationToken"], typeof(byte)],
[secondMethodInfo, ["key", "t"], typeof(sbyte)],
// etc.
```

This would then go with asserts like:

```csharp
Assert.Equal(methodInfo.GetParameters().Select(s => s.Name), expectedParameterNames);
Assert.Equal(methodInfo.ReturnParameter.ParameterType, expectedReturnParameterType);
```

That's the general gist anyway.

test/Swashbuckle.AspNetCore.SwaggerGen.Test/XmlComments/XmlMethodInfoExtensionsTests.cs
```diff
@@ -9,20 +11,20 @@ public class XmlMethodInfoExtensionsTests
 {
     [Theory]
     [MemberData(nameof(DifferentMethodsSignatures))]
-    public void DifferentMethodsSignatures_ShouldBeExpected(MethodInfo methodInfo)
+    public void DifferentMethodsSignatures_ShouldBeExpected((MethodInfo MethodInfo, IEnumerable<string> ExpectedParameterNames, Type ExpectedReturnParameterType) theoryData)
```
martincostello, line 14: This doesn't need to be a tuple, `TheoryData` has variants with more type parameters, so it can be `TheoryData<MethodInfo, string[], Type>`.

test/Swashbuckle.AspNetCore.SwaggerGen.Test/XmlComments/XmlMethodInfoExtensionsTests.cs
```diff
@@ -10,21 +10,31 @@ namespace Swashbuckle.AspNetCore.SwaggerGen.Test;
...
+        Assert.NotNull(underlyingGenericMethod.ReturnType);
+        Assert.Equal(expectedReturnType, underlyingGenericMethod.ReturnType);
+    }
+
+    public class DifferentMethodsSignaturesData : TheoryData<MethodInfo, IEnumerable<string>, Type>
+    {
+        public DifferentMethodsSignaturesData()
+        {
+            var methods = typeof(NonGenericResourceController).GetMethods()
+            .Where(s => s.Name == nameof(NonGenericResourceController.DifferentMethodsSignatures));
```
martincostello, line 31:
```suggestion
            var methods = typeof(NonGenericResourceController)
                .GetMethods()
                .Where(s => s.Name == nameof(NonGenericResourceController.DifferentMethodsSignatures));
```

test/Swashbuckle.AspNetCore.SwaggerGen.Test/XmlComments/XmlMethodInfoExtensionsTests.cs
```diff
@@ -10,21 +10,31 @@ namespace Swashbuckle.AspNetCore.SwaggerGen.Test;
...
+
+            foreach (var method in methods)
+            {
+                Add(method, method.GetParameters().Select(p => p.Name), method.ReturnType);
```
martincostello, line 35:
```suggestion
                Add(method, method.GetParameters().Select(p => p.Name).ToArray(), method.ReturnType);
```

test/Swashbuckle.AspNetCore.SwaggerGen.Test/XmlComments/XmlMethodInfoExtensionsTests.cs
```diff
@@ -10,21 +10,31 @@ namespace Swashbuckle.AspNetCore.SwaggerGen.Test;
...
         Assert.NotEmpty(underlyingGenericMethodParameters);
-        Assert.Equal(theoryData.ExpectedParameterNames, underlyingGenericMethodParameters.Select(s => s.Name));
-        Assert.NotNull(underlyingGenericMethod.ReturnParameter);
-        Assert.Equal(theoryData.ExpectedReturnParameterType, underlyingGenericMethod.ReturnParameter.ParameterType);
+        Assert.Equal(expectedParameterNames, underlyingGenericMethodParameters.Select(s => s.Name));
+        Assert.NotNull(underlyingGenericMethod.ReturnType);
+        Assert.Equal(expectedReturnType, underlyingGenericMethod.ReturnType);
+    }
+
+    public class DifferentMethodsSignaturesData : TheoryData<MethodInfo, IEnumerable<string>, Type>
```
martincostello, line 26:
```suggestion
    public class DifferentMethodsSignaturesData : TheoryData<MethodInfo, string[], Type>
```

test/Swashbuckle.AspNetCore.SwaggerGen.Test/XmlComments/XmlMethodInfoExtensionsTests.cs
```diff
@@ -10,21 +10,31 @@ namespace Swashbuckle.AspNetCore.SwaggerGen.Test;
 public class XmlMethodInfoExtensionsTests
 {
     [Theory]
-    [MemberData(nameof(DifferentMethodsSignatures))]
-    public void DifferentMethodsSignatures_ShouldBeExpected((MethodInfo MethodInfo, IEnumerable<string> ExpectedParameterNames, Type ExpectedReturnParameterType) theoryData)
+    [ClassData(typeof(DifferentMethodsSignaturesData))]
+    public void DifferentMethodsSignatures_ShouldBeExpected(MethodInfo methodInfo, IEnumerable<string> expectedParameterNames, Type expectedReturnType)
```
martincostello, line 14:
```suggestion
    public void DifferentMethodsSignatures_ShouldBeExpected(MethodInfo methodInfo, string[] expectedParameterNames, Type expectedReturnType)
```

martincostello (approved):

Issue #2985, opened 2024-07-17
[Bug]: Example value in XML comment doesn't work when using source generation for JSON serialization


hackerzhuli:
### Describe the bug

Hello, I can't get example value in XML comment to work in Swagger with this library. My project is an Web API based on Asp.net Core, Native AOT, .net 8 and Minimal API. JSON serializaion is source generation based.

Example:
``` cs
public record What(string Content);

public static class TestEndPoint
{
    /// <summary>
    /// This is swash
    /// </summary>
    /// <param name="name" example="my name">The name</param>
    /// <returns></returns>
    public static Ok<What> Swash(string name)
    {
        return TypedResults.Ok(new What("what"));
    } 
}

```

If I delete ```example="my name"```, swagger works, that is, XML comment is used, for example, summary "This is swash" is shown in Swagger page. With ```example="my name"```, opening the Swagger page throws an exception.

### Expected behavior

Show an example value in corresponding Swagger endpoint parameter. 

### Actual behavior

Throws an exception when I open swagger url.

### Steps to reproduce

reproduce repo: https://github.com/hackerzhuli/TestSwash

### Exception(s) (if any)

``` txt
fail: Microsoft.AspNetCore.Diagnostics.DeveloperExceptionPageMiddleware[1]
      An unhandled exception has occurred while executing the request.
      Swashbuckle.AspNetCore.SwaggerGen.SwaggerGeneratorException: Failed to generate Operation for action - HTTP: GET /test/swash => Swash. See inner exception
       ---> System.InvalidOperationException: Reflection-based serialization has been disabled for this application. Either use the source generator APIs or explicitly configure the 'JsonSerializerOptions.TypeInfoResolver' property.
         at System.Text.Json.ThrowHelper.ThrowInvalidOperationException_JsonSerializerIsReflectionDisabled()
         at System.Text.Json.JsonSerializerOptions.ConfigureForJsonSerializer()
         at System.Text.Json.JsonSerializer.GetTypeInfo(JsonSerializerOptions options, Type inputType)
         at System.Text.Json.JsonSerializer.GetTypeInfo[T](JsonSerializerOptions options)
         at System.Text.Json.JsonSerializer.Serialize[TValue](TValue value, JsonSerializerOptions options)
         at Swashbuckle.AspNetCore.SwaggerGen.XmlCommentsExampleHelper.Create(SchemaRepository schemaRepository, OpenApiSchema schema, String exampleString)
[...]

PR #2986, opened 2024-07-17
Make XML comment examples AoT compatible

martincostello:
Use the JSON source generator with `net8.0` to make XML comments' example generation compatible with native AoT.

Relates to #2985.


martincostello:
Closing as doesn't appear to be appetite to fix.

Issue #2686, opened 2023-07-25
JsonRequired property from STJ not marked as required in generated OpenApi Schema


aygalinc:
Hello,

I try to use swashbuckle in one of my project and I notice when I mark some property with property: JsonRequired in my record, the corresponding specification does not marke the property as required. Is this the expected behavior or a bug ? 

A reproducer : 

public sealed record OrganizationCustomExchangeRatesDto(
    [property: JsonRequired] TemporalHeaderDto ColumnDates,
   [property: JsonRequired] FrequencyDto Frequency,
    [property: JsonRequired] List<OrganizationCurrencyPairRowDto> CurrenciesRates
);

Generates : 

"OrganizationCustomExchangeRatesDto": {
        "type": "object",
        "properties": {
          "columnDates": {
            "$ref": "#/components/schemas/TemporalHeaderDto"
          },
          "frequency": {
            "$ref": "#/components/schemas/FrequencyDto"
          },
          "currenciesRates": {
            "type": "array",
            "items": {
              "$ref": "#/components/schemas/OrganizationCurrencyPairRowDto"
            },
            "nullable": true
          }
        },
        "additionalProperties": false
      },

CurrenciesRates is also mark as nullable despite of my Json required annotation.

Havunen:
This will be fixed in DotSwashbuckle 3.0.3+
https://github.com/Havunen/DotSwashbuckle
https://www.nuget.org/packages/DotSwashbuckle.AspNetCore

Issue #2798, opened 2024-04-14
Drop support for out-of-support .NET versions


martincostello:
The published binaries and code are starting to get bloated with adding support for .NET 8 for #2792, so we should look to remove support for target frameworks that are now _completely unsupported_ by Microsoft.

- `netcoreapp2.2`
- `netcoreapp3.0`
- `netcoreapp3.1`
- `net5.0`
- `net6.0` (from November 2024)
- `net7.0` (from May 2024)

This should be done as a major version bump.

We could either drop them all at once some time after November 2024, or do it in a phased approach:

1. Drop everything not supported today (.NET < 6.0 + 7.0);
2. Drop .NET 6 at a future date.

martincostello:
As part of any breaking changes made when we do this, we should also consider removing any constructor overloads we've added over time to preserve binary compatibility which also ties into #2805.

martincostello:
Will be resolved by #3007.

cremor:
> We could either drop them all at once some time after November 2024, or do it in a phased approach:
>
> 1. Drop everything not supported today (.NET < 6.0 + 7.0);
> 2. Drop .NET 6 at a future date.

Seems like you chose the phased approach based on that #3007 still contains .NET 6.0?

martincostello:
Yes - see #3087.

PR #2988, opened 2024-07-18
Take into account JsonRequired of STJ

jgarciadelanoceda:
Fixes the issue #2686 
JsonRequiredAttribute was introduced in dotnet7

### DIFF
--- a/test/Swashbuckle.AspNetCore.SwaggerGen.Test/SchemaGenerator/JsonSerializerSchemaGeneratorTests.cs
+++ b/test/Swashbuckle.AspNetCore.SwaggerGen.Test/SchemaGenerator/JsonSerializerSchemaGeneratorTests.cs
@@ -885,6 +885,20 @@ namespace Swashbuckle.AspNetCore.SwaggerGen.Test
             Assert.Equal(new[] { "string-with-json-property-name" }, schema.Properties.Keys.ToArray());
         }
 
+#if NET7_0_OR_GREATER
+        [Fact]
+        public void GenerateSchema_HonorsSerializerAttribute_JsonRequired()
+        {
+            var schemaRepository = new SchemaRepository();
+
+            var referenceSchema = Subject().GenerateSchema(typeof(JsonRequiredAnnotatedType), schemaRepository);
+
+            var schema = schemaRepository.Schemas[referenceSchema.Reference.Id];
+            Assert.Equal(new[] { "StringWithJsonRequired" }, schema.Required.ToArray());
+            Assert.True(schema.Properties["StringWithJsonRequired"].Nullable);
+        }
+#endif
+
         [Fact]
         public void GenerateSchema_HonorsSerializerAttribute_JsonExtensionData()
         {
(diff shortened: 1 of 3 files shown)


test/Swashbuckle.AspNetCore.SwaggerGen.Test/SchemaGenerator/JsonSerializerSchemaGeneratorTests.cs
```diff
@@ -885,6 +885,20 @@ public void GenerateSchema_HonorsSerializerAttribute_JsonPropertyName()
...
+#if NET7_0_OR_GREATER
+        [Fact]
+        public void GenerateSchema_HonorsSerializerAttribute_JsonRequired()
+        {
+            var schemaRepository = new SchemaRepository();
+
+            var referenceSchema = Subject().GenerateSchema(typeof(JsonRequiredAnnotatedType), schemaRepository);
+
+            var schema = schemaRepository.Schemas[referenceSchema.Reference.Id];
+            Assert.Equal(new[] { "StringWithJsonRequired" }, schema.Required.ToArray());
```
martincostello, line 897: Can this be simplified with collection expressions?

martincostello (commented):
Could you add (or amend an existing) verification test for this scenario please?

test/Swashbuckle.AspNetCore.SwaggerGen.Test/SchemaGenerator/JsonSerializerSchemaGeneratorTests.cs
jgarciadelanoceda: Call is ambiguous unless ToArray.ToSpan(), as the rest of the methods are in the same way I keep this way

test/WebSites/WebApi/EndPoints/OpenApiEndpoints.cs
```diff
@@ -61,4 +67,6 @@ record WeatherForecast(DateOnly Date, int TemperatureC, string? Summary)
     record class Person(string FirstName, string LastName);
 
     record class Address(string Street, string City, string State, string ZipCode);
+    sealed record OrganizationCustomExchangeRatesDto([property: JsonRequired] CurrenciesRate[] CurrenciesRates);
+    sealed record CurrenciesRate([property: JsonRequired] string currencyFrom, [property: JsonRequired] string currencyTo, [property: JsonRequired] double rate);
```
martincostello, line 71: Could you make one of them _not_ required so we can see the different behaviour for the same model in the rendered contract please?

martincostello (approved):

cremor:
> JsonRequiredAttribute was introduced in dotnet7 altough you can opt in with a nuget reference.. I do not see this an option because previous versions of dotnet are about to get out of support

Do I understand that correctly that you only had to wrap the code in an `#if` because Swashbuckle references STJ 4.6.0, which doesn't contain `JsonRequiredAttribute`?
If so, then why doesn't Swashbuckle update to the current version of STJ? As far as I can see it the current version of STJ (8.0.4) would still support all the target frameworks that Swashbuckle.AspNetCore.SwaggerGen uses.

martincostello:
We will review the dependencies and their versions when we tackle #2798.

Trying to use the latest version with multiple TFMs can introduce a lot of complexity.

PR #3003, opened 2024-08-01
Add API analysers

martincostello:
- Add public API analysers.
- Add public API baselines.
- Enable NuGet package validation.

Resolves #2801.

PR #3007, opened 2024-08-01
Support .NET 9

martincostello:
Add support for .NET 9 and drop support for out-of-support versions of .NET.

Changes:

- Add `net9.0` TFMs.
- Drop TFMs for all out-of-support versions of .NET.
- Bump version to `7.0.0`.
- Only audit direct NuGet dependencies.
- Suppress vulnerabilities in IdentityServer4 used in sample flagged by the .NET 9 SDK.
- Remove some now-redundant pre-processor directives.

Needs rebasing after #3003 is merged.

_Will remain in draft until November 2024 when .NET 9 is officially released, with the intention to release soon afterwards_


test/Swashbuckle.AspNetCore.IntegrationTests/Swashbuckle.AspNetCore.IntegrationTests.csproj
```diff
@@ -38,8 +38,13 @@
   </ItemGroup>
 
   <ItemGroup Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net8.0'))">
-    <PackageReference Update="Microsoft.AspNetCore.Mvc.Testing" VersionOverride="8.0.4" />
-    <PackageReference Update="Microsoft.AspNetCore.TestHost" VersionOverride="8.0.4" />
+    <PackageReference Update="Microsoft.AspNetCore.Mvc.Testing" VersionOverride="8.0.7" />
+    <PackageReference Update="Microsoft.AspNetCore.TestHost" VersionOverride="8.0.7" />
```
martincostello, line 42:
```suggestion
    <PackageReference Update="Microsoft.AspNetCore.Mvc.Testing" VersionOverride="8.0.10" />
    <PackageReference Update="Microsoft.AspNetCore.TestHost" VersionOverride="8.0.10" />
```

Directory.Build.props
```diff
@@ -20,18 +20,12 @@
...
-    TODO Update packages to resolve vulnerability warnings
-    -->
-    <NoWarn>$(NoWarn);NU1902;NU1903</NoWarn>
-    <!--
     TODO Fix warning from Swashbuckle.AspNetCore with dotnet pack
     -->
     <NoWarn>$(NoWarn);NU5128</NoWarn>
+    <!-- TODO Remove once .NET 9.0.0 is released -->
+    <NoWarn>$(NoWarn);NU5104</NoWarn>
+    <NuGetAuditMode>direct</NuGetAuditMode>
```
bkoelman, line 28: I ran into `NuGetAuditMode` as well and did some investigation I'd like to share here, in case it helps. My conclusion is that library projects, such as Swashbuckle, should keep `<NuGetAuditMode>direct</NuGetAuditMode>` in their codebase and _not_ release a patch version each time a dependent package is patched due to vulnerabilities. (It would be nice to update to non-vulnerable versions on a major release though.)

The justification for that is because of https://github.com/dotnet/runtime/pull/107161. It's part of .NET 9, but was also packported and is included in the [.NET October 2024 Updates](https://devblogs.microsoft.com/dotnet/dotnet-and-dotnet-framework-october-2024-servicing-updates/).

What I _think_ this means, is that when an app has a framework or SDK reference, plus a reference to Swashbuckle that depends on a vulnerable version of STJ, no vulnerability warning is logged during `dotnet restore`. There's no point in updating the NuGet reference in the app, because the framework itself provides a newer version, which will be used at runtime. As a result, Swashbuckle doesn't need to release a patch each time STJ is patched, to avoid users getting warnings about vulnerable dependencies.

There is further discussion at https://github.com/dotnet/runtime/issues/105120, which states that Microsoft policy is NOT to ship new versions only to satisfy vulnerable dependencies:

image

src/Swashbuckle.AspNetCore.Annotations/Swashbuckle.AspNetCore.Annotations.csproj
```diff
@@ -8,7 +8,7 @@
     <NoWarn>$(NoWarn);1591</NoWarn>
     <PackageTags>swagger;documentation;discovery;help;webapi;aspnet;aspnetcore;annotations</PackageTags>
     <SignAssembly>true</SignAssembly>
-    <TargetFrameworks>netstandard2.0;netcoreapp3.0;net5.0;net6.0;net7.0;net8.0</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;net6.0;net8.0;net9.0</TargetFrameworks>
```
bkoelman, line 11: From my experience, it's better to put the most recent TFM first. When cloning the repo for the first time and opening a file in Visual Studio, that's the version being activated in the code editor. So when newer APIs are available in .NET 9, IntelliSense will show them, so developers are aware. Of course, then they'll need to use `#if` to fix up older targets, but at least the new APIs are being shown.

bkoelman (commented):
It's great to see preparations being made for .NET 9. I've shared some comments based on my own experience, that may be of help.

.github/workflows/build.yml
```diff
@@ -52,7 +52,7 @@ jobs:
       with:
         dotnet-version: |
           6.0.x
-          7.0.x
+          8.0.x
```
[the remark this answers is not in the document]
martincostello: This isn't needed as there's a separate install step that installs whatever global.json specifies, which is now the .NET 9 SDK.

Directory.Build.props
martincostello: I was going to re-audit the build and any upgrades _just_ for NuGet Audit warnings before merging, so this is a good reminder, but there was no intention to create "churn" by doing new releases if/when there are updates for vulnerabilities. This is just about keeping the build green but also not having to add lots of extra transient version overrides for anything somewhere in the tree with its own warning.

martincostello: Just checked, and there's 46 warnings/errors with this removed.

.github/workflows/build.yml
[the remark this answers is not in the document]
bkoelman: Well, this way you don't need to specify an exact version. But it doesn't matter anymore now, because there won't be any more pre-releases until RTM.

Directory.Build.props
bkoelman: Sorry, forget about this whole conversation. I hadn't had a chance to try it out yet. Turns out it does _not at all_ work as I thought. The October update _only_ helps the runtime team, which can be seen in their updates from a few days ago:

image

So they removed dependencies on packages that are already part of the .NET SDK itself (`System.Text.Json` here). However, it does not help other SDKs, (nor third-party package authors), which explains why the version of `Microsoft.Extensions.Options` (included in the Web SDK) was bumped.

I suspect this could be handled better, and in a way that third-party package authors (like Swashbuckle) would benefit from, which I'm going to propose to the NuGet team.

src/Swashbuckle.AspNetCore.Annotations/Swashbuckle.AspNetCore.Annotations.csproj
bkoelman: I'm unable to resolve this conversation (I don't know why), but it looks like this was addressed and can be closed.

.github/workflows/build.yml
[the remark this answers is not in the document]
martincostello: I prefer to have specific version in global.json for repeatability of the build, so there's no version specified for 9.x in the workflow as it's specified in global.json. That's updated monthly by other automation.

These are just installed so the tests can run under those additional runtimes as well as whichever runtime comes with the 9.0 SDK installed at any given time and without assuming they are installed on the build agent.

test/WebSites/NswagClientExample/NswagClientExample.csproj
```diff
@@ -1,10 +1,13 @@
...
 
   <PropertyGroup>
-    <TargetFrameworks>net6.0;net8.0</TargetFrameworks>
+    <TargetFrameworks>net9.0;net8.0;net6.0</TargetFrameworks>
     <DotNetSwaggerPath>$([System.IO.Path]::Combine('$(ArtifactsPath)', 'bin', 'Swashbuckle.AspNetCore.Cli', '$(Configuration.ToLower())_$(TargetFramework)'))</DotNetSwaggerPath>
     <_NSwagTool>$(NSwagExe_Net60)</_NSwagTool>
     <_NSwagTool Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net8.0'))">$(NSwagExe_Net80)</_NSwagTool>
+    <!--
+    <_NSwagTool Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net9.0'))">$(NSwagExe_Net90)</_NSwagTool>
+    -->
```
martincostello, line 10:
```suggestion
```

test/WebSites/OAuth2Integration/OAuth2Integration.csproj
```diff
@@ -14,4 +14,10 @@
     <ProjectReference Include="..\..\..\src\Swashbuckle.AspNetCore.SwaggerGen\Swashbuckle.AspNetCore.SwaggerGen.csproj" />
     <ProjectReference Include="..\..\..\src\Swashbuckle.AspNetCore.SwaggerUI\Swashbuckle.AspNetCore.SwaggerUI.csproj" />
   </ItemGroup>
+
+  <!-- This is just a sample for testing - updating to Duende.IdentityServer can be done at some point in the future -->
```
martincostello, line 18:
```suggestion
  <!--
    This is just a sample for testing - updating to Duende.IdentityServer can be done at some point in the future.
    See https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/3008.
  -->
```

martincostello:
@mesMess Could I get a review/approval on this PR ahead-of-time please? I'd like to be able to merge this and ship a release on 12th November once the stable `9.0.0` packages are in NuGet.org.

jgarciadelanoceda:
@martincostello the dotnet6 release is going to be out of release the same day that dotnet9 comes, why not deleting also dotnet6?
Also if the nugget of STJ is greater than 8 we can delete a lot of compiled primitives that are present in #2988 have you considered it?

martincostello:
It was considered (see original issue) but I decided against it to not generate cliff on day 1 of support ending. 6 will be removed in a future release.

jgarciadelanoceda:
Yeah I was thinking of .net standard and also framework only, because( I do not know if many are using), but it can give the users of that platforms the ability of using the new STJ attributes

martincostello:
_Maybe_, but sometimes libraries have some features that only light up in newer runtimes due to language features needed, or new APIs or overloads of them. Trying to run a .NET 6 app with a 9 package won't have been tested for, and won't necessarily be able to use any improvements (or worse, fall back to netstandard2.0 and then lose even 5.0 or 6.0 specific features).

Ultimately, it's not worth the complexity/burden to us in terms of support, over just using 6 with 6, 8 with 8 etc.

If people want new features, they can upgrade their apps to the version that introduces them.

mesMess:
> @mesMess Could I get a review/approval on this PR ahead-of-time please? I'd like to be able to merge this and ship a release on 12th November once the stable `9.0.0` packages are in NuGet.org.

Yep, I'll do that today.

mesMess (approved):
Build bump 7 -> 8 (next LTS)
Microsoft OenApi version bump
Bump to the RC for .NET 9
Moving some object to internal from public
Some of the shipped/un-shipped files are empty
Sealing some internal classes (UI)

mesMess (approved):

test/Swashbuckle.AspNetCore.IntegrationTests/Swashbuckle.AspNetCore.IntegrationTests.csproj
```diff
@@ -38,8 +38,13 @@
   </ItemGroup>
 
   <ItemGroup Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net8.0'))">
-    <PackageReference Update="Microsoft.AspNetCore.Mvc.Testing" VersionOverride="8.0.4" />
-    <PackageReference Update="Microsoft.AspNetCore.TestHost" VersionOverride="8.0.4" />
+    <PackageReference Update="Microsoft.AspNetCore.Mvc.Testing" VersionOverride="8.0.10" />
+    <PackageReference Update="Microsoft.AspNetCore.TestHost" VersionOverride="8.0.10" />
```
martincostello, line 42:
```suggestion
    <PackageReference Update="Microsoft.AspNetCore.Mvc.Testing" VersionOverride="8.0.11" />
    <PackageReference Update="Microsoft.AspNetCore.TestHost" VersionOverride="8.0.11" />
```

bkoelman:
That's fast!

PS: Best of luck with your talk at netconf.

coderdnewbie:
Thanks for supporting .Net 9.0, just made my life a lot easier.

PR #3030, opened 2024-08-16
Added dependency injection/easy registration for async filters

tofi92:
Async filters (#2938) could not be registered with the SwaggerGenOptions, so there was no way of using DI with async filters.

I extended the SwaggerGenOptions for all the async filters and also added extension methods similiar to the sync equivalents, so added async filters is now as intuitive as sync filters, e.g. `swaggerGenOptions.DocumentAsyncFilter<MyAsyncFilter>()`.


src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/ConfigureSwaggerGeneratorOptions.cs
```diff
@@ -46,6 +47,19 @@ public void Configure(SwaggerGeneratorOptions options)
             _swaggerGenOptions.DocumentFilterDescriptors.ForEach(
                 filterDescriptor => options.DocumentFilters.Add(GetOrCreateFilter<IDocumentFilter>(filterDescriptor)));
 
+
```
martincostello, line 50:
```suggestion
```

src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenOptionsExtensions.cs
```diff
@@ -353,7 +353,7 @@ public static void SchemaFilter<TFilter>(
                 Type = typeof(TFilter),
                 Arguments = arguments
             });
-        }
+        }
```
martincostello, line 356:
```suggestion
        }
```

src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenOptions.cs
```diff
@@ -23,6 +23,14 @@ public class SwaggerGenOptions
...
 
         public List<FilterDescriptor> SchemaFilterDescriptors { get; set; } = new List<FilterDescriptor>();
+
+        public List<FilterDescriptor> ParameterAsyncFilterDescriptors { get; set; } = new List<FilterDescriptor>();
+
+        public List<FilterDescriptor> RequestBodyAsyncFilterDescriptors { get; set; } = new List<FilterDescriptor>();
+
+        public List<FilterDescriptor> OperationAsyncFilterDescriptors { get; set; } = new List<FilterDescriptor>();
+
+        public List<FilterDescriptor> DocumentAsyncFilterDescriptors { get; set; } = new List<FilterDescriptor>();
```
martincostello, line 33: Have you investigated whether it's possible to extend the existing functionality to support both types of filters so we don't have to double the surface area?

src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/ConfigureSwaggerGeneratorOptions.cs
```diff
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.Linq;
 using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.Mvc.Filters;
```
martincostello, line 5: What's this for?

test/Swashbuckle.AspNetCore.SwaggerGen.Test/ConfigureSwaggerGeneratorOptionsTests.cs
```diff
@@ -216,4 +216,202 @@ public static void AddingDocumentFilterTypes_WhenConfiguringOptions_DifferentIns
         Assert.Equal(2, swaggerGeneratorOptions.DocumentFilters.Count);
         Assert.NotSame(swaggerGeneratorOptions.DocumentFilters.First(), swaggerGeneratorOptions.DocumentFilters.Last());
     }
+
+    //Hier async
+
```
martincostello, line 221:
```suggestion
```

src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/ConfigureSwaggerGeneratorOptions.cs
```diff
@@ -34,17 +35,57 @@ public void Configure(SwaggerGeneratorOptions options)
 
             // Create and add any filters that were specified through the FilterDescriptor lists ...
 
-            _swaggerGenOptions.ParameterFilterDescriptors.ForEach(
-                filterDescriptor => options.ParameterFilters.Add(GetOrCreateFilter<IParameterFilter>(filterDescriptor)));
+            foreach(var filterDescriptor in _swaggerGenOptions.ParameterFilterDescriptors)
```
martincostello, line 38:
```suggestion
            foreach (var filterDescriptor in _swaggerGenOptions.ParameterFilterDescriptors)
```

src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/ConfigureSwaggerGeneratorOptions.cs
```diff
@@ -34,17 +35,57 @@ public void Configure(SwaggerGeneratorOptions options)
...
+            {
+                if (filterDescriptor.Type.IsAssignableTo(typeof(IParameterFilter)))
+                {
+                    options.ParameterFilters.Add(GetOrCreateFilter<IParameterFilter>(filterDescriptor));
+                }
+
+                if (filterDescriptor.Type.IsAssignableTo(typeof(IParameterAsyncFilter)))
+                {
+                    options.ParameterAsyncFilters.Add(GetOrCreateFilter<IParameterAsyncFilter>(filterDescriptor));
+                }
```
martincostello, line 48: If a class implements both interfaces, won't this mean it gets registered as both kinds and runs twice? Should it do that, or should it prefer one interface over the other?

src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenOptions.cs
tofi92: Hi, I pushed a new change that removes those additional lists but that also removes the elegant one-liners with ForEach when creating the filters. I tried pattern matching first, but if a filter implements both sync and async filters, only one will be registered.

src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/ConfigureSwaggerGeneratorOptions.cs
tofi92: Seems my visual studio thought I needed that - I removed it

src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/ConfigureSwaggerGeneratorOptions.cs
tofi92: As a developer if I would implement both I would expect both to be run.

martincostello: But would you expect it to be instantiated twice, once for each interface? I think it would make more sense to instantiate it, then put the same instance into one or both sets.

tofi92: Should this only apply to filters added as instances or also for filters added as types with di? Because registering the same filter twice with e.g. `options.OperationFilter<T>` already creates two different instances, so I would also keep this behaviour even if the class is the same. When using e.g. `options.AddOperationFilterInstance(TFilter instance)` I also agree that only one instance for both implementations should be added.

pinkfloydx33: Same as if you add a class that implements both a schema filter and an operation filter (non async variants). It would instantiate it twice. I think it's less surprising to keep that behavior

pinkfloydx33:
I was very say today when I converted a bunch of filters to be async and then couldn't register them without a bunch of ceremony. I definitely would like this to make it in in some form!

martincostello:
Thanks for your contribution - this looks good.

Could you bump the version to 6.8.0 as we're adding new public API members please? Then this looks good to merge.

https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/7a7230dd54be847ac82982010e00497d9fe5f2cf/Directory.Build.props#L52

tofi92:
Hi @martincostello 
sorry for the delay, some family problems.

Thanks for your input and for accepting :)

martincostello (approved):

Issue #3022, opened 2024-08-12
[Bug]: Nullability problems with dictionaries


ozziepeeps:
### Describe the bug

Two issues: `Dictionary<string, int>?` in a nullable context is incorrectly described with the `int` value being nullable. Additionally the behavior of Swashbuckle changes if my property type is `IDictionary<string, int>?` vs `Dictionary<string, int>?` (e.g. interface vs concrete).

### Expected behavior

If I have a property `nullableMapNonNullableValue` of type `Dictionary<string, int>?` or `IDictionary<string, int>?` on a schema, I would expect the `int` value to be described as NOT nullable in `additionalProperties`, like so:

```json
{
    "nullableMapNonNullableValue": {
        "type": "object",
        "additionalProperties": {
            "type": "integer",
            "format": "int32"
        },
        "nullable": true
    }
}
```

### Actual behavior

The `int` value in `additionalProperties` is incorrectly described as being nullable:

```jsonc
{
    "nullableMapNonNullableValue": {
        "type": "object",
        "additionalProperties": {
            "type": "integer",
            "format": "int32",
            "nullable": true // <-- this is wrong!
        },
        "nullable": true
    }
}
```

If, however, I change the type of the property to be `IDictionary<string, int>?` (e.g. interface instead of concrete) the behavior changes and the nullability of the `int` value is now described correctly in `additionalProperties`:

```json
{
    "nullableMapNonNullableValue": {
        "type": "object",
        "additionalProperties": {
            "type": "integer",
            "format": "int32"
        },
        "nullable": true
    }
}
```

### Steps to reproduce

I'm submitting a PR with unit tests that fixes both issues.

### Swashbuckle.AspNetCore version

6.7.0

### .NET Version

net8.0

PR #3041, opened 2024-08-24
Use NullabilityInfoContext to determine dictionary value nullability

patrikwlund:
A follow-up of #3022 (issue) and #3023 (PR).

The #3023 PR broke the behavior for many of our `IReadOnlyDictionary<string, string>` (and similar) members.

The reason is that the logic for reading and interpreting `NullableAttribute` and `NullableContextAttribute` flags is too simplified and is incorrect for several cases. It just happens to pass the current test cases because of the layout of the test type and its members.

## Too simplified
For example, it ignores the optimization of the flags array into a single byte if they are all the same value. An optimized `Dictionary<string, string>` has a single `[1]` flag, not two or three flags as the current logic expects.

It also does a lot of assumptions related to reference vs value types. An IDictionary itself can be a value type, as can the key, both of which would affect the length of the flag array. It should be checking the number of value types and the number of flags together.

It also doesn't seem to fully respect the `NullableContextAttribute` that's placed on the type containing the member. And the tests did not account for differences in the context value, as the testing type always only has a single flag. So the other flag value is not tested.

## My suggested change
For NET 6 and onwards we should rely on the official `NullabilityInfoContext` to tell us what is and isn't nullable. For earlier versions we can still have custom logic, which we can try to improve on if we want to put in the effort. But it will require a lot of effort to make it fully compliant and flexible.

The tests I added do not pass with the custom flag logic, but they do pass with the `NullabilityInfoContext` code.

The tests could be more flexible or enumerated in other more dynamic ways if you would like, to get the number of lines down, but I tried to follow the existing pattern of laying them all out in-line.

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/MemberInfoExtensions.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/MemberInfoExtensions.cs
@@ -61,6 +61,19 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
 
         public static bool IsDictionaryValueNonNullable(this MemberInfo memberInfo)
         {
+#if NET6_0_OR_GREATER
+            var context = new NullabilityInfoContext();
+            var nullableInfo = memberInfo.MemberType == MemberTypes.Field
+                ? context.Create((FieldInfo)memberInfo)
+                : context.Create((PropertyInfo)memberInfo);
+
+            if (nullableInfo.GenericTypeArguments.Length != 2)
+            {
+                throw new InvalidOperationException("Expected Dictionary to have two generic type arguments.");
+            }
+
+            return nullableInfo.GenericTypeArguments[1].ReadState == NullabilityState.NotNull;
+#else
             var memberType = memberInfo.MemberType == MemberTypes.Field
                 ? ((FieldInfo)memberInfo).FieldType
                 : ((PropertyInfo)memberInfo).PropertyType;
@@ -104,6 +117,7 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
             }
 
             return false;
+#endif
         }
 
         private static object GetNullableAttribute(this MemberInfo memberInfo)
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/MemberInfoExtensions.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/MemberInfoExtensions.cs
@@ -69,7 +69,11 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
 
             if (nullableInfo.GenericTypeArguments.Length != 2)
             {
-                throw new InvalidOperationException("Expected Dictionary to have two generic type arguments.");
+                var length = nullableInfo.GenericTypeArguments.Length;
+                var type = nullableInfo.Type.FullName;
+                var container = memberInfo.DeclaringType.FullName;
+                var member = memberInfo.Name;
+                throw new InvalidOperationException($"Expected Dictionary to have two generic type arguments but it had {length}. Member: {container}.{member} Type: {type}.");
             }
 
             return nullableInfo.GenericTypeArguments[1].ReadState == NullabilityState.NotNull;
--- a/test/Swashbuckle.AspNetCore.SwaggerGen.Test/SchemaGenerator/JsonSerializerSchemaGeneratorTests.cs
+++ b/test/Swashbuckle.AspNetCore.SwaggerGen.Test/SchemaGenerator/JsonSerializerSchemaGeneratorTests.cs
@@ -674,6 +674,11 @@ namespace Swashbuckle.AspNetCore.SwaggerGen.Test
         [Fact]
         public void TypeWithNullableContextAnnotated_IsAnnotated()
         {
+            // This is a sanity check to ensure that TypeWithNullableContextAnnotated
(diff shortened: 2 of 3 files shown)


src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/MemberInfoExtensions.cs
```diff
@@ -61,6 +61,19 @@ public static bool IsNonNullableReferenceType(this MemberInfo memberInfo)
...
         {
+#if NET6_0_OR_GREATER
+            var context = new NullabilityInfoContext();
+            var nullableInfo = memberInfo.MemberType == MemberTypes.Field
+                ? context.Create((FieldInfo)memberInfo)
+                : context.Create((PropertyInfo)memberInfo);
+
+            if (nullableInfo.GenericTypeArguments.Length != 2)
+            {
+                throw new InvalidOperationException("Expected Dictionary to have two generic type arguments.");
```
martincostello, line 72: Could you update the message with a bit more detail, such as the type and the member names? Then if it happened, it would be much easier to debug from the message alone.

test/Swashbuckle.AspNetCore.SwaggerGen.Test/SchemaGenerator/JsonSerializerSchemaGeneratorTests.cs
```diff
@@ -670,15 +671,55 @@ public void GenerateSchema_SupportsOption_UseInlineDefinitionsForEnums()
...
+
+            Assert.NotNull(nullableContext);
+
+            var flag = nullableContext?.GetType().GetField("Flag")?.GetValue(nullableContext);
+
+            Assert.Equal((byte)2, flag);
+        }
+
+        [Fact]
+        public void TypeWithNullableContextNotAnnotated_IsNotAnnotated()
```
martincostello, line 691: Are these two tests just "help you develop" tests? They don't seem to test any part of Swashbuckle, but just assumptions/understanding of nullable annotations.

test/Swashbuckle.AspNetCore.TestSupport/Fixtures/TypeWithNullableContext.cs
```diff
@@ -3,8 +3,41 @@
...
+        public string Dummy11 { get; set; } = default!;
+        public string Dummy12 { get; set; } = default!;
+        public string Dummy13 { get; set; } = default!;
+        public string Dummy14 { get; set; } = default!;
+        public string Dummy15 { get; set; } = default!;
+        public string Dummy16 { get; set; } = default!;
+        public string Dummy17 { get; set; } = default!;
+        public string Dummy18 { get; set; } = default!;
+        public string Dummy19 { get; set; } = default!;
+        public string Dummy20 { get; set; } = default!;
```
martincostello, line 39: Why so many properties?

martincostello (commented):
Thanks for the PR! Just a few questions but otherwise this looks great.

test/Swashbuckle.AspNetCore.SwaggerGen.Test/SchemaGenerator/JsonSerializerSchemaGeneratorTests.cs
patrikwlund: They were intended as sanity checks for the test types, so that we know if our assumptions are suddenly not correct anymore. It can change invisibly and without error or notice if we rewrite the types, or if the compiler behavior changes. 

We have no direct control over how the compiler places its attributes, but for the test suite to be effective we need the attributes to be placed in a certain way.

I should probably add a comment in each test to mention this.

martincostello: Ok cool, that makes sense. Yeah some comments to that effect would be good.

test/Swashbuckle.AspNetCore.TestSupport/Fixtures/TypeWithNullableContext.cs
patrikwlund: I didn't find any other way of coercing the compiler to change the flag in `NullableContextAttribute`. The context flag seems to pick the most common case as an optimization, so that as many members as possible can avoid having `NullableAttribute` at all. But our logic uses the two attributes to figure things out, so the tests need to be able to replicate both nullable and non-nullable cases.

So:
- Many non-nullable -> `NullableContext(1)`
- Many nullable -> `NullableContext(2)`

When we remove all the manual logic related to `Nullable` and `NullableContext` at some point, we can probably simplify and remove a lot of this awkward compiler coercion, and just trust that Microsoft's `NullabilityInfoContext` does a good job.

martincostello (approved):

test/Swashbuckle.AspNetCore.TestSupport/Fixtures/TypeWithNullableContext.cs
martincostello: Thanks for explaining.

martincostello:
Thanks for fixing this - the fix is now available in NuGet as version 6.7.2.

test/Swashbuckle.AspNetCore.TestSupport/Fixtures/TypeWithNullableContext.cs
ozziepeeps: This is interesting. So it's the large number of properties that "tips" the compiler over into doing the "optimized" format? Is the optimized format described somewhere in the Roslyn docs or somewhere else? Thank you for the improved fix!

PR #3044, opened 2024-08-25
Improve IncludeXmlComments performance (2)

### LINKED ISSUE #2164: IncludeXmlComments performance issues

mus65:
# Pull Request

Fixes #2164

This is a rebase and enhancement of #2443  (thanks to @stevendarby, added you as co-author).

This significantly improves the performance of the XML comment filters. The original fix #2443 created a dictionary of the members to avoid iterating the whole XPath for every single item.

I made some further improvements by replacing all usages of `Select` and `SelectSingleNode` with `SelectChildren` (added some helper extensions for this). This completely avoids XPath expressions and [the documentation](https://learn.microsoft.com/en-us/dotnet/standard/data/xml/select-xml-data-using-xpathnavigator#optimized-selection-methods) even mentions `SelectChildren` being faster.

I made the new constructors internal for now (which is also why I changed `ParameterFilter` to `AddParameterFilterInstance` etc.. DI fails otherwise). It felt wrong to me to make these public, but I can make these public if you want.

### Tests

The existing tests imho already cover this code pretty well. I only added a test case for a bug I noticed in #2443  (there was a `return` instead of a `continue` in `XmlCommentsDocumentFilter`).

### Performance

#2443 already mentions significant improvements (25s -> 4s). With another local test case I have with a 1.5MB XML file, generation time decreases from ~1900ms to ~200ms.

[#2443](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/pull/2443#issuecomment-2094261461) also mentions some concern about memory consumption, so I made some manual tests with a very big (30MB) XML document file: the XPathDocument itself already loads the document completely into memory which uses about 60 MB.  Creating the dictionary creates an additional 40MB. So there is some additional memory consumption, but considering that 30MB is pretty extreme and most files should be < 1MB, the memory increase should be negligible for most applications. In my local test case with a 1.5 MB File, memory consumption after GC.Collect increased by about 1MB.

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XPathNavigatorExtensions.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XPathNavigatorExtensions.cs
@@ -1,0 +1,22 @@
+using System.Linq;
+using System.Xml.XPath;
+
+namespace Swashbuckle.AspNetCore.SwaggerGen
+{
+    internal static class XPathNavigatorExtensions
+    {
+        internal static XPathNavigator SelectFirstChild(this XPathNavigator navigator, string name)
+        {
+            return navigator.SelectChildren(name, "")
+                    ?.Cast<XPathNavigator>()
+                    .FirstOrDefault();
+        }
+
+        internal static XPathNavigator SelectFirstChildWithAttribute(this XPathNavigator navigator, string childName, string attributeName, string attributeValue)
+        {
+            return navigator.SelectChildren(childName, "")
+                    ?.Cast<XPathNavigator>()
+                    .FirstOrDefault(n => n.GetAttribute(attributeName, "") == attributeValue);
+        }
+    }
+}
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XPathNavigatorExtensions.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XPathNavigatorExtensions.cs
@@ -1,22 +1,21 @@
 using System.Linq;
 using System.Xml.XPath;
 
-namespace Swashbuckle.AspNetCore.SwaggerGen
-{
-    internal static class XPathNavigatorExtensions
-    {
-        internal static XPathNavigator SelectFirstChild(this XPathNavigator navigator, string name)
-        {
-            return navigator.SelectChildren(name, "")
-                    ?.Cast<XPathNavigator>()
-                    .FirstOrDefault();
-        }
+namespace Swashbuckle.AspNetCore.SwaggerGen;
 
-        internal static XPathNavigator SelectFirstChildWithAttribute(this XPathNavigator navigator, string childName, string attributeName, string attributeValue)
-        {
-            return navigator.SelectChildren(childName, "")
-                    ?.Cast<XPathNavigator>()
-                    .FirstOrDefault(n => n.GetAttribute(attributeName, "") == attributeValue);
-        }
+internal static class XPathNavigatorExtensions
+{
+    internal static XPathNavigator SelectFirstChild(this XPathNavigator navigator, string name)
+    {
+        return navigator.SelectChildren(name, "")
+                ?.Cast<XPathNavigator>()
+                .FirstOrDefault();
+    }
+
+    internal static XPathNavigator SelectFirstChildWithAttribute(this XPathNavigator navigator, string childName, string attributeName, string attributeValue)
+    {
+        return navigator.SelectChildren(childName, "")
+                ?.Cast<XPathNavigator>()
+                .FirstOrDefault(n => n.GetAttribute(attributeName, "") == attributeValue);
     }
 }
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XPathNavigatorExtensions.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XPathNavigatorExtensions.cs
@@ -5,17 +5,29 @@ namespace Swashbuckle.AspNetCore.SwaggerGen;
 
 internal static class XPathNavigatorExtensions
 {
+    private const string EmptyNamespace = "";
+
+    internal static XPathNodeIterator SelectChildren(this XPathNavigator navigator, string name)
+    {
+        return navigator.SelectChildren(name, EmptyNamespace);
+    }
+
+    internal static string GetAttribute(this XPathNavigator navigator, string name)
+    {
+        return navigator.GetAttribute(name, EmptyNamespace);
+    }
+
     internal static XPathNavigator SelectFirstChild(this XPathNavigator navigator, string name)
     {
-        return navigator.SelectChildren(name, "")
+        return navigator.SelectChildren(name, EmptyNamespace)
                 ?.Cast<XPathNavigator>()
                 .FirstOrDefault();
     }
 
     internal static XPathNavigator SelectFirstChildWithAttribute(this XPathNavigator navigator, string childName, string attributeName, string attributeValue)
     {
-        return navigator.SelectChildren(childName, "")
+        return navigator.SelectChildren(childName, EmptyNamespace)
                 ?.Cast<XPathNavigator>()
-                .FirstOrDefault(n => n.GetAttribute(attributeName, "") == attributeValue);
+                .FirstOrDefault(n => n.GetAttribute(attributeName, EmptyNamespace) == attributeValue);
     }
 }
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XPathNavigatorExtensions.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XPathNavigatorExtensions.cs
@@ -20,14 +20,14 @@ internal static class XPathNavigatorExtensions
     internal static XPathNavigator SelectFirstChild(this XPathNavigator navigator, string name)
     {
         return navigator.SelectChildren(name, EmptyNamespace)
-                ?.Cast<XPathNavigator>()
+                ?.OfType<XPathNavigator>()
                 .FirstOrDefault();
     }
 
     internal static XPathNavigator SelectFirstChildWithAttribute(this XPathNavigator navigator, string childName, string attributeName, string attributeValue)
     {
         return navigator.SelectChildren(childName, EmptyNamespace)
-                ?.Cast<XPathNavigator>()
+                ?.OfType<XPathNavigator>()
                 .FirstOrDefault(n => n.GetAttribute(attributeName, EmptyNamespace) == attributeValue);
     }
 }
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsDocumentFilter.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsDocumentFilter.cs
@@ -9,10 +9,9 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
 {
     public class XmlCommentsDocumentFilter : IDocumentFilter
     {
-        private const string MemberXPath = "/doc/members/member[@name='{0}']";
         private const string SummaryTag = "summary";
 
-        private readonly XPathNavigator _xmlNavigator;
+        private readonly IReadOnlyDictionary<string, XPathNavigator> _xmlDocMembers;
         private readonly SwaggerGeneratorOptions _options;
 
         public XmlCommentsDocumentFilter(XPathDocument xmlDoc)
@@ -20,9 +19,13 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
         {
         }
 
-        public XmlCommentsDocumentFilter(XPathDocument xmlDoc, SwaggerGeneratorOptions options)
+        public XmlCommentsDocumentFilter(XPathDocument xmlDoc, SwaggerGeneratorOptions options) : this(XmlCommentsDocumentHelper.GetMemberDictionary(xmlDoc), options)
         {
-            _xmlNavigator = xmlDoc.CreateNavigator();
+        }
+
+        internal XmlCommentsDocumentFilter(IReadOnlyDictionary<string, XPathNavigator> xmlDocMembers, SwaggerGeneratorOptions options)
+        {
+            _xmlDocMembers = xmlDocMembers;
             _options = options;
         }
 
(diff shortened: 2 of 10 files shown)


src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XPathNavigatorExtensions.cs
```diff
@@ -0,0 +1,22 @@
+using System.Linq;
+using System.Xml.XPath;
+
+namespace Swashbuckle.AspNetCore.SwaggerGen
```
martincostello, line 4: For the new files could you use a file-scoped namespace please?

src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XPathNavigatorExtensions.cs
```diff
@@ -0,0 +1,22 @@
...
+{
+    internal static class XPathNavigatorExtensions
+    {
+        internal static XPathNavigator SelectFirstChild(this XPathNavigator navigator, string name)
+        {
+            return navigator.SelectChildren(name, "")
+                    ?.Cast<XPathNavigator>()
```
martincostello, line 11: Are the children here on line 18 guaranteed to always be an `XPathNavigator`?

src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XPathNavigatorExtensions.cs
```diff
@@ -0,0 +1,22 @@
+using System.Linq;
+using System.Xml.XPath;
+
+namespace Swashbuckle.AspNetCore.SwaggerGen
+{
+    internal static class XPathNavigatorExtensions
+    {
+        internal static XPathNavigator SelectFirstChild(this XPathNavigator navigator, string name)
+        {
+            return navigator.SelectChildren(name, "")
```
martincostello, line 10: For readability I think it would be useful to name a constant `const string EmptyNamespace = ""` and then use that in the three places, as otherwise I wasn't sure what it signified just from the diff and had to go and look it up.

src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsDocumentFilter.cs
```diff
@@ -38,22 +41,20 @@ public void Apply(OpenApiDocument swaggerDoc, DocumentFilterContext context)
             foreach (var nameAndType in controllerNamesAndTypes)
             {
                 var memberName = XmlCommentsNodeNameHelper.GetMemberNameForType(nameAndType.Value);
-                var typeNode = _xmlNavigator.SelectSingleNode(string.Format(MemberXPath, memberName));
 
-                if (typeNode != null)
+                if (!_xmlDocMembers.TryGetValue(memberName, out var typeNode)) continue;
```
martincostello, line 45:
```suggestion
                if (!_xmlDocMembers.TryGetValue(memberName, out var typeNode))
                {
                    continue;
                }
```

src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsDocumentFilter.cs
```diff
@@ -38,22 +41,20 @@ public void Apply(OpenApiDocument swaggerDoc, DocumentFilterContext context)
...
+
+                var summaryNode = typeNode.SelectFirstChild(SummaryTag);
+                if (summaryNode != null)
                 {
-                    var summaryNode = typeNode.SelectSingleNode(SummaryTag);
-                    if (summaryNode != null)
+                    if (swaggerDoc.Tags == null)
+                        swaggerDoc.Tags = new List<OpenApiTag>();
```
martincostello, line 51:
```suggestion
                    swaggerDoc.Tags ??= new List<OpenApiTag>();
```

src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsDocumentHelper.cs
```diff
@@ -0,0 +1,25 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using System.Xml.XPath;
+
+namespace Swashbuckle.AspNetCore.SwaggerGen
+{
+    internal static class XmlCommentsDocumentHelper
```
martincostello, line 7: Similar comments here on the empty namespace.

src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsDocumentHelper.cs
```diff
@@ -0,0 +1,25 @@
...
+    {
+        internal static IReadOnlyDictionary<string, XPathNavigator> GetMemberDictionary(XPathDocument xmlDoc)
```
martincostello, line 9: If we're going for performance, this could return the concrete dictionary type to avoid the virtual dispatch over the interface as it's internal and no one _should_ be mutating it.

src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsDocumentHelper.cs
```diff
@@ -0,0 +1,25 @@
...
+                ?.SelectFirstChild("members")
+                ?.SelectChildren("member", "")
+                ?.OfType<XPathNavigator>();
+
+            if (members == null)
+            {
+                return new Dictionary<string, XPathNavigator>();
+            }
+
+            return members.ToDictionary(memberNode => memberNode.GetAttribute("name", ""));
```
martincostello, line 22: Is this guaranteed to never have any key collisions?

src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsOperationFilter.cs
```diff
@@ -32,26 +37,28 @@ public void Apply(OpenApiOperation operation, OperationFilterContext context)
         private void ApplyControllerTags(OpenApiOperation operation, Type controllerType)
         {
             var typeMemberName = XmlCommentsNodeNameHelper.GetMemberNameForType(controllerType);
-            var responseNodes = _xmlNavigator.Select($"/doc/members/member[@name='{typeMemberName}']/response");
+
+            if (!_xmlDocMembers.TryGetValue(typeMemberName, out var methodNode)) return;
+
+            var responseNodes = methodNode.SelectChildren("response", "");
```
martincostello, line 43: Could add an extension method for `SelectChildren(string)` that calls `SelectChildren(value, "")` (via the constant I mentioned) to avoid the repeated need to pass through an empty namespace in the first place?

src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsDocumentHelper.cs
martincostello, line 9: I think `Create` is the better verb here than `Get`, as before I read the definition I assumed it was some sort of static cache.

src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XPathNavigatorExtensions.cs
martincostello: Can you code it in a way that doesn't assume internal implementation details? The runtime could theoretically change the implementation - would `OfType<XPathNavigator>()` do the same but in a way that won't throw with an `InvalidCastException` is that happened?

src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsDocumentHelper.cs
martincostello: If they're documented as unique then it should be fine to use `ToDictionary()` rather than manually loop and overwrite (or not) any conflicts.

src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XPathNavigatorExtensions.cs
mus65: I changed it  to OfType.

mus65:
I created a benchmark in a separate branch [here](https://github.com/mus65/Swashbuckle.AspNetCore/commit/65bfd9745ca2cb757326b79cf3ef88959239af4d). As expected, the new implementation is a lot
faster, even with a small member count. The performance of the new implementation stays the same,
regardless of member count. The old implementation gets significantly slower the more members
exist. `addMemberCount` can be changed in the benchmark to try different member counts.

Old implementation with small member count (n = 79):

| Method      | Mean      | Error     | StdDev    | Median    | Gen0   | Gen1   | Allocated |
|------------ |----------:|----------:|----------:|----------:|-------:|-------:|----------:|
| Document    |  4.074 us | 0.0811 us | 0.1780 us |  4.002 us | 2.2278 | 0.5569 |  20.53 KB |
| Operation   | 12.692 us | 0.0828 us | 0.0775 us | 12.712 us | 5.0659 | 0.0610 |  47.06 KB |
| Parameter   |  6.878 us | 0.1368 us | 0.2360 us |  6.786 us | 2.6321 | 0.0076 |  24.29 KB |
| RequestBody |  4.122 us | 0.0185 us | 0.0164 us |  4.125 us | 2.2278 | 0.0610 |  20.64 KB |

Old implementation with big member count (n = 10.000):

| Method      | Mean       | Error    | StdDev   | Median     | Gen0     | Gen1   | Allocated  |
|------------ |-----------:|---------:|---------:|-----------:|---------:|-------:|-----------:|
| Document    |   541.3 us | 10.56 us | 13.73 us |   542.8 us |  78.1250 | 1.9531 |  723.66 KB |
| Operation   | 1,061.1 us | 20.70 us | 31.61 us | 1,074.7 us | 156.2500 |      - | 1453.34 KB |
| Parameter   |   489.3 us |  6.42 us |  5.01 us |   489.4 us |  79.1016 |      - |  727.42 KB |
| RequestBody |   476.8 us |  9.43 us | 16.52 us |   467.4 us |  78.1250 |      - |  723.79 KB |

New implementation with any member count:

| Method      | Mean     | Error     | StdDev    | Median   | Gen0   | Gen1   | Allocated |
|------------ |---------:|----------:|----------:|---------:|-------:|-------:|----------:|
| Document    | 2.675 us | 0.0530 us | 0.0955 us | 2.726 us | 1.8959 | 0.6294 |  17.45 KB |
| Operation   | 4.851 us | 0.0907 us | 0.0848 us | 4.869 us | 3.7842 | 0.1526 |  34.84 KB |
| Parameter   | 2.421 us | 0.0313 us | 0.0293 us | 2.411 us | 1.8959 | 0.0687 |  17.46 KB |
| RequestBody | 2.545 us | 0.0136 us | 0.0127 us | 2.545 us | 1.8158 |      - |  16.73 KB |

martincostello:
Thanks @mus65 for doing those benchmarks - I'll take a closer look at them tomorrow 👍

martincostello:
Thanks again for adding these benchmarks. I've pushed up a version of them in #3050 for us to use going forwards.

These are the results I get on my laptop:

### N=79

#### Base

_Pending_

#### This PR

_Pending_

### N=10,000

#### Base

```
BenchmarkDotNet v0.14.0, Windows 11 (10.0.22621.4037/22H2/2022Update/SunValley2)
12th Gen Intel Core i7-1270P, 1 CPU, 16 logical and 12 physical cores
.NET SDK 8.0.401
  [Host]     : .NET 8.0.8 (8.0.824.36612), X64 RyuJIT AVX2
  DefaultJob : .NET 8.0.8 (8.0.824.36612), X64 RyuJIT AVX2
```

| Method      | Mean       | Error    | StdDev   | Gen0     | Gen1   | Allocated  |
|------------ |-----------:|---------:|---------:|---------:|-------:|-----------:|
| Document    |   598.6 μs |  8.41 μs |  7.02 μs |  78.1250 | 1.9531 |  724.36 KB |
| Operation   | 1,105.6 μs | 19.72 μs | 17.48 μs | 156.2500 |      - | 1453.67 KB |
| Parameter   |   516.7 μs |  9.92 μs |  8.28 μs |  79.1016 |      - |  727.55 KB |
| RequestBody |   487.9 μs |  4.22 μs |  3.52 μs |  78.1250 |      - |  723.79 KB |

#### This PR

```
BenchmarkDotNet v0.14.0, Windows 11 (10.0.22621.4037/22H2/2022Update/SunValley2)
12th Gen Intel Core i7-1270P, 1 CPU, 16 logical and 12 physical cores
.NET SDK 8.0.401
  [Host]     : .NET 8.0.8 (8.0.824.36612), X64 RyuJIT AVX2
  DefaultJob : .NET 8.0.8 (8.0.824.36612), X64 RyuJIT AVX2
```

| Method      | Mean      | Error     | StdDev    | Median    | Gen0   | Gen1   | Allocated |
|------------ |----------:|----------:|----------:|----------:|-------:|-------:|----------:|
| Document    |  3.872 μs | 0.1670 μs | 0.4655 μs |  3.857 μs | 1.9150 | 0.4807 |  17.66 KB |
| Operation   | 12.341 μs | 1.2843 μs | 3.7869 μs | 14.136 μs | 3.8147 |      - |  35.17 KB |
| Parameter   |  7.158 μs | 0.3641 μs | 1.0679 μs |  7.320 μs | 1.8921 | 0.0916 |   17.6 KB |
| RequestBody |  7.953 μs | 0.3605 μs | 1.0629 μs |  8.024 μs | 1.8005 |      - |  16.73 KB |

martincostello (approved):

martincostello:
I have [some benchmarks](https://github.com/martincostello/aspnetcore-openapi/tree/main/perf/TodoApp.Benchmarks) I've been using to compare the benchmarks of Swashbuckle, NSwag and the new .NET 9 ASP.NET Core OpenAPI library, and there's a noticeable really improvement from the ingesting changes from this PR.

image

The first major improvement is from updating to .NET 9 RC1 (https://github.com/martincostello/aspnetcore-openapi/commit/7e5e365cd54751db3a9f7eacdb7d6c2c4cf114d6), then the second improvement with the big reduction in memory usage is from updating to 6.8.0 (https://github.com/martincostello/aspnetcore-openapi/commit/b4aba6567d2f114df11d3bf0fd53495b3ba393d8)

Thanks again!

PR #3043, opened 2024-08-24
Fix nested types nullable context check

VladimirTyrin:
# Pull Request

## The issue or feature being addressed

Fixes #3040

## Details on the issue fix or feature implementation

We have to address the case when model type is nested-within-nested. In this case for nullable context analysis we have to traverse the whole declaring type chain

In issue example NullableContextAttribute is on the outer class
Sharplab example: https://sharplab.io/#v2:D4AQTAjAsAUCDMACciDCiDetE+UkALIgLIAUAlJtrgL6x0ywKIBOApgMYD2LAJogC0uXCKSEiAdOLCIAMmwBubADYQANMggAGOYpUQAygBcWASwB2AcwAKLLgAdysLDFx5WnHv2ljhYKcJI8krKYBog2rohYMZmVrYOTq64Lm5uzOzcfIKBpBE6wSrwsRY2do4A3NQ4DDRAA

PR #3046, opened 2024-08-26
Use NullabilityInfoContext to determine if member is nullable

patrikwlund:
Add-on to https://github.com/domaindrivendev/Swashbuckle.AspNetCore/pull/3043 (should be no conflict)

Very similar to my previous PR: https://github.com/domaindrivendev/Swashbuckle.AspNetCore/pull/3041

In his PR, @VladimirTyrin surgically fixes a specific problem for all platforms. I suggest we use the new and fancy NET6+ toys when available, to avoid custom nullable logic completely in the future. I tested it with the same added test as in @VladimirTyrin's original PR.


martincostello:
That's not necessary. I've fixed the conflict and enabled squash merge.

Issue #2248, opened 2021-10-13
What is the recommended way to force a query string array parameter to be treated with style form and explode false


AroglDarthu:
Sorry to post this question here, but hoping to get a more response than on stackoverlow: https://stackoverflow.com/questions/69287190/what-is-the-recommended-way-to-force-a-query-string-array-parameter-to-be-treate

In short, I'm trying to get a comma-separated parameter (`?p=a,b,c,d`) that is typed as an `IEnumerable<string>`, to show up correctly in the generated Swagger. According to the specs it should get `style=form` and `explode=false`. The default here (`style=form` and `explode=true`) matches the asp.net core default.

Is there an annotation attribute that I can place on the parameter? I cannot find anything in previous issues or in the documentation.

sibijohn72:
bump! looking for the same thing.

Issue #2115, opened 2021-05-07
Add Example to Swashbuckle.AspNetCore.Annotations


AchoArnold:
Thanks for contributing to Swashbuckle.AspNetCore! As per the [contributing guidelines](CONTRIBUTING.md), please adhere to the following rules of thumb before submitting your issue:

It would be nice to have an annotation for an example property. Perhaps we could add it to `[SwaggerSchema]`? I know currently you can add the example using XML.

APIWT:
I agree, this would be super helpful. Especially now that we have record types, so for positional records it isn't super feasible to add XML docs in a minimalistic manner.

collinbarrett:
To me, an "example" doesn't have exact correlation with a "default value". But, you can use the existing `DefaultValueAttribute` to kind of achieve this, I think (assuming I am understanding the OP).

See [here](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/575#issuecomment-980675458).

PR #3065, opened 2024-09-04
Added Explode and Style to SwaggerParameter

aberus:
# Pull Request

It is currently not possible to set **Explode** and **Style** on the parameter.

Added `bool Explode` and `string Style` with backing properties `bool? ExplodeFlag` and `Microsoft.OpenApi.Models.ParameterStyle? ParameterStyle` respectively. `AnnotationsParameterFilter` applies these two properties if set.

### DIFF
--- a/src/Swashbuckle.AspNetCore.Annotations/SwaggerParameterAttribute.cs
+++ b/src/Swashbuckle.AspNetCore.Annotations/SwaggerParameterAttribute.cs
@@ -1,4 +1,6 @@
 ﻿using System;
+using Microsoft.OpenApi.Models;
+using Swashbuckle.AspNetCore.SwaggerGen;
 
 namespace Swashbuckle.AspNetCore.Annotations
 {
@@ -30,5 +32,37 @@ namespace Swashbuckle.AspNetCore.Annotations
         }
 
         internal bool? RequiredFlag { get; set; }
+
+        /// <summary>
+        /// Describes how the parameter value will be serialized depending on the type of
+        /// the parameter value. Default values (based on value of in): for query - form;
+        /// for path - simple; for header - simple; for cookie - form.
+        /// </summary>
+        public string Style
+        {
+            get { throw new InvalidOperationException($"Use {nameof(ParameterStyle)} instead"); }
+            set
+            {
+                ParameterStyle = Enum.TryParse(value, ignoreCase: true, out ParameterStyle result) ? result :
+                    throw new InvalidOperationException(
+                        message: $"Style '{value}' not defined in OpenAPI specification");
+            }
+        }
+
+        internal ParameterStyle? ParameterStyle { get; set; }
+
+        /// <summary>
+        /// When this is true, parameter values of type array or object generate separate parameters for
+        /// each value of the array or key-value pair of the map. For other types of parameters this property
+        /// has no effect. When style is form, the default value is true. For all other styles,
+        /// the default value is false.
+        /// </summary>
+        public bool Explode
+        {
+            get { throw new InvalidOperationException($"Use {nameof(ExplodeFlag)} instead"); }
+            set { ExplodeFlag = value; }
+        }
+
+        internal bool? ExplodeFlag { get; set; }
     }
 }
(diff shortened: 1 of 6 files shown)


src/Swashbuckle.AspNetCore.Annotations/SwaggerParameterAttribute.cs
```diff
@@ -30,5 +32,37 @@ public bool Required
...
+        /// for path - simple; for header - simple; for cookie - form.
+        /// </summary>
+        public string Style
+        {
+            get { throw new InvalidOperationException($"Use {nameof(ParameterStyle)} instead"); }
+            set
+            {
+                ParameterStyle = Enum.TryParse(value, ignoreCase: true, out ParameterStyle result) ? result :
+                    throw new InvalidOperationException(
+                        message: $"Style '{value}' not defined in OpenAPI specification");
```
martincostello, line 48: I would make this a `NotSupportedException` and tweak the wording. Then if the specification later adds a new value, but we don't support it, then we're not making claims about the specification that are false.

src/Swashbuckle.AspNetCore.Annotations/SwaggerParameterAttribute.cs
```diff
@@ -30,5 +32,37 @@ public bool Required
...
+        }
+
+        internal ParameterStyle? ParameterStyle { get; set; }
+
+        /// <summary>
+        /// When this is true, parameter values of type array or object generate separate parameters for
+        /// each value of the array or key-value pair of the map. For other types of parameters this property
+        /// has no effect. When style is form, the default value is true. For all other styles,
+        /// the default value is false.
+        /// </summary>
```
martincostello, line 59: I would simplify this explanation so we're not copying too many implementation details that could change in the future into our documentation.

src/Swashbuckle.AspNetCore.Annotations/SwaggerParameterAttribute.cs
```diff
@@ -30,5 +32,37 @@ public bool Required
         }
 
         internal bool? RequiredFlag { get; set; }
+
+        /// <summary>
+        /// Describes how the parameter value will be serialized depending on the type of
+        /// the parameter value. Default values (based on value of in): for query - form;
+        /// for path - simple; for header - simple; for cookie - form.
```
martincostello, line 39: Similarly, can we simplify this?

martincostello (commented):
Could you add at least one integration/verification test that shows the rendered effect of the new properties in an OpenAPI document please?

jgarciadelanoceda:
This could close the issue #2248

jgarciadelanoceda:
@aberus, are you planning to continue with this PR?, because I find it interesting :). I also will like to add the required parameters to solve the issue #2115(That's why I contact you first, to avoid conflicts between us)

martincostello:
@aberus Thanks for responding to the review suggestions. This still needs tests before we can merge it though.

aberus:
@martincostello I plan to add tests. But also I think that generation need to be changed, to support these two scenarions.

The default ASP.NET model binding parses the query `?sizes=a,b,c,d` as a parameter of type `string`.
```
[HttpGet("Products")]
public IActionResult Get([FromQuery(Name = "sizes")][SwaggerParameter(Explode = false)] string sizes)
{
    return Ok();
}
```
The custom model binding (`[CommaSeparated]`) mentioned in #2248 can be mapped to a collection;
```
[HttpGet("Products")]
public IActionResult Get([FromQuery(Name = "sizes")][SwaggerParameter(Explode = false)][CommaSeparated] IEnumerable<string> sizes)
{
    return Ok();
}
```

Both of these scenarios should produce this:

```
paths:
  /Products:
    get:
      parameters:
        - name: sizes
          in: query
          style: form
          explode: false
          schema:
            type: array
            items:
              type: string
      responses:
        '200':
          description: Success
```
Let me know, what do you think?

martincostello:
How would you propose adding metadata (i.e. an attribute) indicating that the parameters should be treated that way to both ASP.NET Core and Swashbuckle? The issue points to a Stack Overflow answer, which points to a blog post, which mentions a custom implementation of ASP.NET Core model binding. We wouldn't want to be shipping model binders in Swashbuckle, because that's not specific to OpenAPI.

aberus:
My suggestion in first example is that a non-collection parameter combination with explode set to false, should generate an array in the OpenAPI specification.
Second example show how currently works (with this change) if explode set to false, but external library need to used to work property. I didn't mean that some external library needs to be added, just that this scenario also needs to be supported if someone wants to use it with external library.

FlameWolf (approved):

PR #3068, opened 2024-09-11
SwaggerGen - Improved handling of dictionaries with enum key

flarestudiopl:
# Pull Request

When enum used as a key for the dictionary contains no values defined, the dictionary schema becomes an object with no properties. In such case SchemaGenerator should build schema based on dictionary key and value type rather than known key values (which are unknown in fact).

SchemaGenerator was modified to use "special case where the set of key values is known" only if the collection of known dictionary keys is not empty.

### DIFF
--- a/test/Swashbuckle.AspNetCore.SwaggerGen.Test/SchemaGenerator/JsonSerializerSchemaGeneratorTests.cs
+++ b/test/Swashbuckle.AspNetCore.SwaggerGen.Test/SchemaGenerator/JsonSerializerSchemaGeneratorTests.cs
@@ -122,6 +122,7 @@ namespace Swashbuckle.AspNetCore.SwaggerGen.Test
 
         [Theory]
         [InlineData(typeof(IDictionary<string, int>), "integer")]
+        [InlineData(typeof(IDictionary<EmptyIntEnum, int>), "integer")]
         [InlineData(typeof(IReadOnlyDictionary<string, bool>), "boolean")]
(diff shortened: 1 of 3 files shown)


test/Swashbuckle.AspNetCore.SwaggerGen.Test/SchemaGenerator/JsonSerializerSchemaGeneratorTests.cs
```diff
@@ -122,6 +122,7 @@ public void GenerateSchema_DedupsEnumValues_IfEnumTypeHasDuplicateValues()
 
         [Theory]
         [InlineData(typeof(IDictionary<string, int>), "integer")]
+        [InlineData(typeof(IDictionary<EmptyIntEnum, int>), "integer")]
```
martincostello, line 125: Does this test fail without your other code change?

flarestudiopl: Yes, it fails.

martincostello:
Thanks!

flarestudiopl:
Thanks for quick review :) When can I expect this to be released?

martincostello:
I'm not sure yet - in the meantime you can consume changes from our MyGet feed: https://github.com/domaindrivendev/Swashbuckle.AspNetCore/deployments/MyGet.org

PR #3071, opened 2024-09-13
Add Scalar UI support

Blquinn:
# Add Scalar UI support

This PR adds support for [Scalar](https://github.com/scalar/scalar) which is an alternative openapi viewing UI.

Scalar is nice because it offers an interface that is quick and easy to browse, but unlike ReDoc also allows you to send requests from the UI.

I essentially just copied the code from the ReDoc modules and corresponding integration test modules and ported it over to use Scalar. They work quite similarly.

I've tested this running it locally with various configuration options set as well as through the integration tests.


martincostello:
Thanks for the PR. I _started_ to review this, and in the process I found [this NuGet package](https://www.nuget.org/packages/AspNetCore.Scalar) which looks _remarkably_ like it was already created by someone taking Swashbuckle and adapting it for Scalar (the project file looks a lot like ours, even down to [MSBuild targets](https://github.com/benirave/AspNetCore.Scalar/blob/b9710f4a4d81d1a513dcf8edfde0523764ec1b74/src/AspNetCore.Scalar/AspNetCore.Scalar.csproj#L30-L35)).

Is there anything this PR does that that other package doesn't already do (or can just be contributed there if not)?

If there's already a solution out there, I think it would be better to promote that other person's work than compete with it directly by effectively providing the same thing. It's also one less for us to maintain.

If we did continue with adding this here, there's a fair few things here that reviewing as-new, I would change even if it's effectively copy-paste from existing code (for example, [the way options are registered](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/pull/3071/files#diff-c53eacc7d9dd1570f7a26bd3272126f841e754a72b5d200f21f28246e491956f) doesn't use it correctly, IMHO). I'll make any such comments if this is something we want to add.

Blquinn:
I hadn't seen that package. I guess my googling skills aren't what they once were :cry:

There are in fact **two** scalar implementations on nuget, including this other one: https://www.nuget.org/packages/Scalar.AspNetCore

At first glance that package you referenced appears to do what I need it to do, so I think we can close out this PR. It also has more of scalar's config available through the c# interface than I added.

I'm going to evaluate those other packages, if they seems solid perhaps a link to one of those packages could be added to the README for discoverability, or something to that effect.

Thanks for the review :)

martincostello:
> I'm going to evaluate those other packages, if they seems solid perhaps a link to one of those packages could be added to the README for discoverability, or something to that effect.

Sure, if they meet your needs we'd accept a PR to add them to our README to aid discoverability.

> Thanks for the review :)

No problem! TIL too.

Issue #3038, opened 2024-08-23
[Bug]: Operation parameter with [FromForm] should not be used to describe a request body if more than one such parameter


martincostello:
### Describe the bug

As part of #3020 to fix #3018, a change was made to fix XML comments not being used to describe the request body for a form.

On reflection, this fix wasn't correct as if there are multiple parameter the first parameter is used to describe the request body in its entirety: https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/3018#issuecomment-2306506421

### Expected behavior

The request body should only be documented from the XML documentation for a `[FromForm]`-annotated argument if there is **exactly one** such parameter.

https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/7a7230dd54be847ac82982010e00497d9fe5f2cf/src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsRequestBodyFilter.cs#L17-L26

### Actual behavior

The request body is documented from the XML documentation from the first of multiple arguments with `[FromForm]`.

### Steps to reproduce

Render the OpenAPI document for an application including the following controller:

```csharp
using Microsoft.AspNetCore.Mvc;

namespace FromFormXmlIssue;

[ApiController]
[Route("[controller]")]
public class ReproController : ControllerBase
{
    [HttpGet]
    public int[] Get([FromForm] string First, [FromForm] string Second)
    {
        return [42];
    }
}
```

### Swashbuckle.AspNetCore version

6.7.1

### .NET Version

8.0.8

qcjxberin:
When will there be an update? @martincostello

martincostello:
Whenever someone picks it up. Contributions welcome.

jgarciadelanoceda:
I am thinking in picking it up.
But I think that if there are more than one property and the schema has the properties then we could asign the description to the Property instead of the whole OpenApiRequestBody. What do you think?

With this code you will see what I am referring to:
```c#
public sealed class BodyFilter : IRequestBodyFilter
{
    public void Apply(OpenApiRequestBody requestBody, RequestBodyFilterContext context)
    {
        if (context.FormParameterDescriptions?.Any() ?? false)
        {
            foreach (var (_, mediaType) in requestBody.Content)
            {
                if (mediaType?.Schema?.Properties?.Any() ?? false)
                {
                    foreach (var (nameForProperty, schemaForProperty) in mediaType.Schema.Properties)
                    {
[...]

Issue #2062, opened 2021-03-17
[FromForm] param description not display.


msshim-dev:
![1](https://user-images.githubusercontent.com/78785087/111398960-afddc680-8707-11eb-9c4e-4c53b4d8d0f5.PNG)
![2](https://user-images.githubusercontent.com/78785087/111398966-b409e400-8707-11eb-8ffd-3df11735a1b7.PNG)

Request Body parameters description not display.

thiagoknop3:
Up

bitilly:
i have same problem with version 6.x.x,
i solve it when downgrade to 5.6.3

thiagoknop3:
I did the same , downgrade to 5.6.3 :(

msshim-dev:
OMG...... Why the version goes up but the function goes down...

datlam162:
I did the same, downgraded to 5.6.3, is there any updates on this?

bbhxwl:
I also have this problem. Can I fix the latest version of the bug?

bitilly:
sorry，i have no idea.



***@***.***
 
From: bbhxwl
Date: 2022-01-13 13:14
To: domaindrivendev/Swashbuckle.AspNetCore
CC: bitilly; Comment
Subject: Re: [domaindrivendev/Swashbuckle.AspNetCore] [FromForm] param description not display. (#2062)
I also have this problem. Can I fix the latest version of the bug?
—
Reply to this email directly, view it on GitHub, or unsubscribe.
Triage notifications on the go with GitHub Mobile for iOS or Android.

DominicSaladin:
This still happens in version 6.4.0... Am I doing something wrong or is still not fixed?

GrononS:
I am using version 6.5.0 and this is still a problem. I even installed Swashbuckle.AspNetCore.Annotations alongside this and it still doesn't work.

Please fix, this is a big oversight for me and is making my documentation pretty much useless since I use [FromForm] everywhere :(

alexminza:
https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/2519

martincostello:
#3076

PR #3076, opened 2024-09-22
Do not fill the RequestBody description with the first parameter of a…

jgarciadelanoceda:
Fixes the issues #3038 and #2062.
Better to document each member of a FromForm independently and do it only on the properties.
The original commit that deleted the possibility of documenting a parameter of a FromForm request was this one:https://github.com/domaindrivendev/Swashbuckle.AspNetCore/commit/c1fb18c2c2f864285a7b811a147d8a9ed8c17993.

As you see in the SchemaFilter it deleted the posibility of documenting a Parameter of a FromForm request. My first try was just to look for it again, but it generated duplicated nodes description. For a parameter it ended up documenting the parameter itself and the Schema.. So I opted to just document the Parameter in the RequestBodyFilter (For C# parameters, because the properties were already set)

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsRequestBodyFilter.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsRequestBodyFilter.cs
@@ -13,6 +13,7 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
         public XmlCommentsRequestBodyFilter(XPathDocument xmlDoc) : this(XmlCommentsDocumentHelper.CreateMemberDictionary(xmlDoc))
         {
         }
+
         internal XmlCommentsRequestBodyFilter(IReadOnlyDictionary<string, XPathNavigator> xmlDocMembers)
         {
             _xmlDocMembers = xmlDocMembers;
@@ -28,26 +29,27 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
                 if (propertyInfo is not null)
                 {
                     ApplyPropertyTagsForBody(requestBody, context, propertyInfo);
-                    return;
                 }
-                var parameterInfo = bodyParameterDescription.ParameterInfo();
-                if (parameterInfo is not null)
+                else
                 {
-                    ApplyParamTagsForBody(requestBody, context, parameterInfo);
-                    return;
+                    var parameterInfo = bodyParameterDescription.ParameterInfo();
+                    if (parameterInfo is not null)
+                    {
+                        ApplyParamTagsForBody(requestBody, context, parameterInfo);
+                    }
                 }
             }
             else
             {
                 var numberOfFromForm = context.FormParameterDescriptions?.Count() ?? 0;
-                if (requestBody.Content?.Count is 0 || numberOfFromForm == 0)
+                if (requestBody.Content?.Count is 0 || numberOfFromForm is 0)
                 {
                     return;
                 }
 
                 foreach (var formParameter in context.FormParameterDescriptions)
                 {
-                    if (formParameter.PropertyInfo() is not null || formParameter.Name is null)
+                    if (formParameter.Name is null || formParameter.PropertyInfo() is not null)
                     {
                         continue;
                     }
@@ -60,8 +62,8 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
 
                     foreach (var item in requestBody.Content.Values)
                     {
-                        if ((item?.Schema?.Properties?.TryGetValue(formParameter.Name, out var value) ?? false)
-                            || (item?.Schema?.Properties?.TryGetValue(formParameter.Name.ToCamelCase(), out value) ?? false))
+                        if (item?.Schema?.Properties is { } properties
+                           && (properties.TryGetValue(formParameter.Name, out var value) || properties.TryGetValue(formParameter.Name.ToCamelCase(), out value)))
                         {
                             var (summary, example) = GetParamTags(parameterFromForm);
                             value.Description = summary;
@@ -89,14 +91,10 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
             {
                 summary = XmlCommentsTextHelper.Humanize(summaryNode.InnerXml);
             }
+
             var exampleNode = propertyNode.SelectFirstChild("example");
-            if (exampleNode is null)
-            {
-                return (summary, null);
-            }
 
-            return (summary, exampleNode.ToString());
-
+            return (summary, exampleNode?.ToString());
         }
 
         private void ApplyPropertyTagsForBody(OpenApiRequestBody requestBody, RequestBodyFilterContext context, PropertyInfo propertyInfo)
@@ -125,6 +123,7 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
             {
                 return (null, null);
             }
+
             var targetMethod = methodInfo.DeclaringType.IsConstructedGenericType
                 ? methodInfo.GetUnderlyingGenericTypeMethod()
                 : methodInfo;
@@ -133,12 +132,14 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
             {
                 return (null, null);
             }
+
             var methodMemberName = XmlCommentsNodeNameHelper.GetMemberNameForMethod(targetMethod);
 
             if (!_xmlDocMembers.TryGetValue(methodMemberName, out var propertyNode))
             {
                 return (null, null);
             }
+
             var paramNode = propertyNode.SelectFirstChildWithAttribute("param", "name", parameterInfo.Name);
 
             if (paramNode is null)
@@ -150,7 +151,6 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
             var example = paramNode.GetAttribute("example");
 
             return (summary, example);
-
         }
 
         private void ApplyParamTagsForBody(OpenApiRequestBody requestBody, RequestBodyFilterContext context, ParameterInfo parameterInfo)
@@ -162,17 +162,14 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
                 requestBody.Description = summary;
             }
 
-            if (requestBody.Content?.Count is 0)
+            if (requestBody.Content?.Count is 0 || !string.IsNullOrEmpty(example))
             {
                 return;
             }
 
-            if (!string.IsNullOrEmpty(example))
+            foreach (var mediaType in requestBody.Content.Values)
             {
-                foreach (var mediaType in requestBody.Content.Values)
-                {
-                    mediaType.Example = XmlCommentsExampleHelper.Create(context.SchemaRepository, mediaType.Schema, example);
-                }
+                mediaType.Example = XmlCommentsExampleHelper.Create(context.SchemaRepository, mediaType.Schema, example);
             }
         }
     }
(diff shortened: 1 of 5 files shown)


src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsRequestBodyFilter.cs
```diff
@@ -13,97 +13,165 @@ public class XmlCommentsRequestBodyFilter : IRequestBodyFilter
...
+
+                    var parameterFromForm = formParameter.ParameterInfo();
+                    if (parameterFromForm is null)
+                    {
+                        continue;
+                    }
+
+                    foreach (var item in requestBody.Content.Values)
+                    {
+                        if ((item?.Schema?.Properties?.TryGetValue(formParameter.Name, out var value) ?? false)
```
jgarciadelanoceda, line 63: This part I am not very confident with it.. The problem is this line: https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/8b0e8961ce01c8e1495d64f404609b01fb07a7e3/src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs#L859

But to do it properly I need to pass the SwaggerGeneratorOptions to the RequestBodyFilter.. Which is a new ctor

jgarciadelanoceda:
Another possibility for fixing this issue is to pass the Binding info and apply in the SchemaFilter the change only when the object is a Parameter and also when is FromForm

src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsRequestBodyFilter.cs
```diff
@@ -13,97 +13,165 @@ public class XmlCommentsRequestBodyFilter : IRequestBodyFilter
         public XmlCommentsRequestBodyFilter(XPathDocument xmlDoc) : this(XmlCommentsDocumentHelper.CreateMemberDictionary(xmlDoc))
         {
         }
-
```
martincostello, line 16: Can you put this back please?

src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsRequestBodyFilter.cs
```diff
@@ -13,97 +13,165 @@ public class XmlCommentsRequestBodyFilter : IRequestBodyFilter
...
+                {
+                    ApplyPropertyTagsForBody(requestBody, context, propertyInfo);
+                    return;
+                }
+                var parameterInfo = bodyParameterDescription.ParameterInfo();
+                if (parameterInfo is not null)
+                {
+                    ApplyParamTagsForBody(requestBody, context, parameterInfo);
+                    return;
+                }
```
martincostello, line 38: I think it would be better to put the second part into an `else`, then we don't need the two `return`s.

src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsRequestBodyFilter.cs
```diff
@@ -13,97 +13,165 @@ public class XmlCommentsRequestBodyFilter : IRequestBodyFilter
...
-
-            var propertyInfo = parameterDescription.PropertyInfo();
-            if (propertyInfo is not null)
+            else
             {
-                ApplyPropertyTags(requestBody, context, propertyInfo);
-                return;
-            }
+                var numberOfFromForm = context.FormParameterDescriptions?.Count() ?? 0;
+                if (requestBody.Content?.Count is 0 || numberOfFromForm == 0)
```
martincostello, line 43: For consistency.

```suggestion
                if (requestBody.Content?.Count is 0 || numberOfFromForm is 0)
```

src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsRequestBodyFilter.cs
```diff
@@ -13,97 +13,165 @@ public class XmlCommentsRequestBodyFilter : IRequestBodyFilter
...
+                    return;
+                }
 
-            var parameterInfo = parameterDescription.ParameterInfo();
-            if (parameterInfo is not null)
-            {
-                ApplyParamTags(requestBody, context, parameterInfo);
+                foreach (var formParameter in context.FormParameterDescriptions)
+                {
+                    if (formParameter.PropertyInfo() is not null || formParameter.Name is null)
```
martincostello, line 50: It would be more performant to swap these conditions.

src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsRequestBodyFilter.cs
martincostello, line 63: The level of `?` is a bit excessive - I think it would be more readable to just check for null first:

```csharp
if (item?.Schema?.Properties is not { } properties)
{
    continue;
}
```

src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsRequestBodyFilter.cs
```diff
@@ -13,97 +13,165 @@ public class XmlCommentsRequestBodyFilter : IRequestBodyFilter
...
-
             var exampleNode = propertyNode.SelectFirstChild("example");
-            if (exampleNode is null || requestBody.Content?.Count is 0)
+            if (exampleNode is null)
             {
-                return;
+                return (summary, null);
+            }
+
+            return (summary, exampleNode.ToString());
```
martincostello, line 98: Could simplify to just `return (summary, exampleNode?.ToString());`?

src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsRequestBodyFilter.cs
```diff
@@ -13,97 +13,165 @@ public class XmlCommentsRequestBodyFilter : IRequestBodyFilter
...
+
```
martincostello, line 99:
```suggestion
```

src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsRequestBodyFilter.cs
```diff
@@ -13,97 +13,165 @@ public class XmlCommentsRequestBodyFilter : IRequestBodyFilter
...
+        private (string summary, string example) GetParamTags(ParameterInfo parameterInfo)
         {
             if (parameterInfo.Member is not MethodInfo methodInfo)
             {
-                return;
+                return (null, null);
             }
-
-            // If method is from a constructed generic type, look for comments from the generic type method
             var targetMethod = methodInfo.DeclaringType.IsConstructedGenericType
```
martincostello, line 128:
```suggestion

            var targetMethod = methodInfo.DeclaringType.IsConstructedGenericType
```

src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsRequestBodyFilter.cs
```diff
@@ -13,97 +13,165 @@ public class XmlCommentsRequestBodyFilter : IRequestBodyFilter
...
                 ? methodInfo.GetUnderlyingGenericTypeMethod()
                 : methodInfo;
 
             if (targetMethod is null)
             {
-                return;
+                return (null, null);
             }
-
```
martincostello, line 89: Add back please.

src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsRequestBodyFilter.cs
```diff
@@ -13,97 +13,165 @@ public class XmlCommentsRequestBodyFilter : IRequestBodyFilter
...
             var methodMemberName = XmlCommentsNodeNameHelper.GetMemberNameForMethod(targetMethod);
 
-            if (!_xmlDocMembers.TryGetValue(methodMemberName, out var propertyNode)) return;
-
+            if (!_xmlDocMembers.TryGetValue(methodMemberName, out var propertyNode))
+            {
+                return (null, null);
+            }
             var paramNode = propertyNode.SelectFirstChildWithAttribute("param", "name", parameterInfo.Name);
```
martincostello, line 142:
```suggestion
            var paramNode = propertyNode.SelectFirstChildWithAttribute("param", "name", parameterInfo.Name);
            
```

src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsRequestBodyFilter.cs
```diff
@@ -13,97 +13,165 @@ public class XmlCommentsRequestBodyFilter : IRequestBodyFilter
...
+            if (paramNode is null)
+            {
+                return (null, null);
+            }
+
+            var summary = XmlCommentsTextHelper.Humanize(paramNode.InnerXml);
+            var example = paramNode.GetAttribute("example");
+
+            return (summary, example);
+
```
martincostello, line 153:
```suggestion
```

src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsRequestBodyFilter.cs
```diff
@@ -13,97 +13,165 @@ public class XmlCommentsRequestBodyFilter : IRequestBodyFilter
...
+            }
+
+            if (requestBody.Content?.Count is 0)
+            {
+                return;
+            }
 
-                var example = paramNode.GetAttribute("example");
-                if (!string.IsNullOrEmpty(example))
+            if (!string.IsNullOrEmpty(example))
```
martincostello, line 170: Could make this part of the previous early return as an `||` and them avoid this extra level of indentation?

test/WebSites/Basic/Controllers/FromFormParamsController.cs
```diff
@@ -22,8 +31,16 @@ public IActionResult PostFormWithIgnoredProperties([FromForm] RegistrationFormWi
...
+        /// <summary>
+        /// Summary for Name
+        /// </summary>
+        /// <example>MyName</example>
         public string Name { get; set; }
 
+        /// <summary>
+        /// Sumary for PhoneNumbers
+        /// </summary>
+
```
martincostello, line 43:
```suggestion
```

src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsRequestBodyFilter.cs
martincostello: Given this has never worked anyway, I'm OK with just "guessing" at camelCase for now. We can do a better job if there's user need for it.

martincostello (approved):

martincostello:
@jgarciadelanoceda In the issue descriptions can you use the syntax `fixes #N` so that the auto-close works please? `fixes the issue #n` doesn't work 😃

PR #3096, opened 2024-10-02
Support Results

### LINKED ISSUE #2595: Please support Results<T> as return type of controllers
Introducing .NET 7 `Results<T>` (https://learn.microsoft.com/en-us/aspnet/core/web-api/action-return-types#resultt-type) Microsoft promised that
> All the [[ProducesResponseType]](https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.mvc.producesresponsetypeattribute) attribute's can be excluded, since the HttpResult implementation contributes automatically to the endpoint metadata.

But if I write this:
```c#
/// <summary> Request the forecast of a specific day. </summary>
/// <param name="date" example="2023-01-24">Date of the requested forecast. </param>
[HttpGet("{date}", Name = "GetWeatherForecastByDate")]
public Results<Ok<WeatherForecast>, NotFound> Get(DateOnly date)
{
    var forecast = _weatherForecasts.FirstOrDefault(forecast => forecast.Date == date);
    return forecast != null ? TypedResults.Ok(forecast) : TypedResults.NotFound();
}
```
The generated `swagger.json` only contains that:
```json
"responses": {
  "200": {
    "description": "Success"
  }
}
```
<hr />

In order to get the correct schema I still need to add the following `ProducesResponseTypeAttribute`s :-(
```c#
[ProducesResponseType(typeof(WeatherForecast), (int)HttpStatusCode.OK)]
[ProducesResponseType((int)HttpStatusCode.NotFound)]
```

jgarciadelanoceda:
Fixes #2595

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
@@ -892,7 +892,7 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
                 var returnType = UnwrapTask(controllerActionDescriptor.MethodInfo.ReturnType);
                 if (typeof(IEndpointMetadataProvider).IsAssignableFrom(returnType))
                 {
-                    var populateMetadataMethod = returnType.GetMethod("Microsoft.AspNetCore.Http.Metadata.IEndpointMetadataProvider.PopulateMetadata", BindingFlags.Static | BindingFlags.NonPublic);
+                    var populateMetadataMethod = returnType.GetMethod("Microsoft.AspNetCore.Http.Metadata.IEndpointMetadataProvider.PopulateMetadata", BindingFlags.Static);
                     if (populateMetadataMethod is not null)
                     {
                         var endpointBuilder = new MetadataEndpointBuilder();
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
@@ -14,8 +14,11 @@ using Microsoft.OpenApi.Models;
 using Swashbuckle.AspNetCore.Annotations;
 using Swashbuckle.AspNetCore.Swagger;
 
+
 #if NET7_0_OR_GREATER
+using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Http.Metadata;
+using Microsoft.AspNetCore.Mvc.Controllers;
 #endif
 
 namespace Swashbuckle.AspNetCore.SwaggerGen
@@ -436,7 +439,7 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
                         if (countOfParameters == 1)
                         {
                             var requestParameter = requestParameters.First();
-                            content.Schema = GenerateSchemaIncludingFormFile(requestParameter, GenerateSchema(
+                            content.Schema = GenerateSchemaIncludingFromFormWithNoReference(requestParameter, GenerateSchema(
                                 requestParameter.ModelMetadata.ModelType,
                                 schemaRepository,
                                 requestParameter.PropertyInfo(),
@@ -449,7 +452,7 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
                             content.Schema = new OpenApiSchema()
                             {
                                 AllOf = requestParameters.Select(s =>
-                                    GenerateSchemaIncludingFormFile(s, GenerateSchema(
+                                    GenerateSchemaIncludingFromFormWithNoReference(s, GenerateSchema(
                                     s.ModelMetadata.ModelType,
                                     schemaRepository,
                                     s.PropertyInfo(),
@@ -475,7 +478,7 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
                         }
                     }
 
-                    static OpenApiSchema GenerateSchemaIncludingFormFile(ApiParameterDescription apiParameterDescription, OpenApiSchema generatedSchema, OpenApiMediaType mediaType)
+                    static OpenApiSchema GenerateSchemaIncludingFromFormWithNoReference(ApiParameterDescription apiParameterDescription, OpenApiSchema generatedSchema, OpenApiMediaType mediaType)
                     {
                         if (generatedSchema.Reference is null && apiParameterDescription.IsFromForm())
                         {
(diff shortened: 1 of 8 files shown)


src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
```diff
@@ -449,7 +452,7 @@ private async Task<OpenApiOperation> GenerateOpenApiOperationFromMetadataAsync(A
                             content.Schema = new OpenApiSchema()
                             {
                                 AllOf = requestParameters.Select(s =>
-                                    GenerateSchemaIncludingFormFile(s, GenerateSchema(
+                                    GenerateSchemaIncludingFromFormWithNoReference(s, GenerateSchema(
```
jgarciadelanoceda, line 455: Not happy with the name, because now it's not only for IFromFile

martincostello: `GetSchemaForForm()`?

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
```diff
@@ -881,12 +884,51 @@ private OpenApiSchema GenerateSchemaFromFormParameters(
...
+        private static IList<ApiResponseType> GetResponseTypes(ApiDescription apiDescription)
+        {
+#if NET7_0_OR_GREATER
+            var supportedResponseTypes = new List<ApiResponseType>(apiDescription.SupportedResponseTypes);
+            if (apiDescription.ActionDescriptor is ControllerActionDescriptor controllerActionDescriptor)
+            {
+                var returnType = UnwrapTask(controllerActionDescriptor.MethodInfo.ReturnType);
+                if (typeof(IEndpointMetadataProvider).IsAssignableFrom(returnType))
+                {
+                    var populateMetadataMethod = returnType.GetMethod("Microsoft.AspNetCore.Http.Metadata.IEndpointMetadataProvider.PopulateMetadata", BindingFlags.Static | BindingFlags.NonPublic);
```
martincostello, line 896: I wonder if it's best to make this specific so we couldn't be broken in the future by any other overloads being added.

Also, why does it need `NonPublic` - the interface method should be public.

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
```diff
@@ -881,12 +884,51 @@ private OpenApiSchema GenerateSchemaFromFormParameters(
...
+                    if (populateMetadataMethod != null)
+                    {
+                        var endpointBuilder = new MetadataEndpointBuilder();
+                        populateMetadataMethod.Invoke(null, [controllerActionDescriptor.MethodInfo, endpointBuilder]);
+                        var responseTypes = endpointBuilder.Metadata.Cast<IProducesResponseTypeMetadata>().ToList();
```
martincostello, line 901: Use `OfType<T>()`?

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
```diff
@@ -881,12 +884,51 @@ private OpenApiSchema GenerateSchemaFromFormParameters(
...
+                        }
+                    }
+                }
+            }
+
+            static Type UnwrapTask(Type type)
+            => type.IsGenericType && type.GetGenericTypeDefinition() == typeof(Task<>)
+                ? type.GetGenericArguments()[0]
+                : type;
+            return supportedResponseTypes;
```
martincostello, line 921: IMHO it's more readable to not embed static methods in the middle of the code using them.

```suggestion
            return supportedResponseTypes;

            static Type UnwrapTask(Type type)
                => type.IsGenericType && type.GetGenericTypeDefinition() == typeof(Task<>)
                    ? type.GetGenericArguments()[0]
                    : type;
```

test/Swashbuckle.AspNetCore.SwaggerGen.Test/Fixtures/FakeController.cs
```diff
@@ -109,5 +113,8 @@ public void ActionHavingFromFormAttributeButNotWithIFormFile([FromForm] string p
         { }
         public void ActionHavingFromFormAttributeWithSwaggerIgnore([FromForm] SwaggerIngoreAnnotatedType param1)
         { }
+#if NET7_0_OR_GREATER
+        public Results<Ok<int>,NotFound> ActionHavingResults() => TypedResults.Ok(1);
```
martincostello, line 117:
```suggestion
        public Results<Ok<int>, NotFound> ActionHavingResults() => TypedResults.Ok(1);
```

test/WebSites/MvcWithNullable/Program.cs
```diff
@@ -28,7 +29,7 @@
 public class EnumController : ControllerBase
 {
     [HttpGet]
-    public IActionResult Get(LogLevel? logLevel = LogLevel.Error) => Ok(new { logLevel });
+    public Results<Ok<LogLevel?>, NotFound> Get(LogLevel? logLevel = LogLevel.Error) => TypedResults.Ok(logLevel);
```
martincostello, line 32: Rather than change an existing method, can you add an extra one?

Also can you add a test for `IResult` via `Results.Ok()` too if that's part of the results support?

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
jgarciadelanoceda: GetSchemaIncludingForForm?

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
jgarciadelanoceda: Just with static is enough, because it's marked as static

test/WebSites/MvcWithNullable/Program.cs
jgarciadelanoceda: The IResult does not propagate the typeInfo, I have to look at it.. Because the interface IResult is independent from the IEndpointMetadataProvider :(

jgarciadelanoceda: I think it cannot be fixed from our side

jgarciadelanoceda:
It turns out to be nonPublic the method

martincostello:
> It turns out to be nonPublic the method

_Feels_ wrong - if it's a method on an interface that's public it should be public, and if it's genuinely non public we shouldn't depend on it.

jgarciadelanoceda:
> > It turns out to be nonPublic the method
> 
> _Feels_ wrong - if it's a method on an interface that's public it should be public, and if it's genuinely non public we shouldn't depend on it.

Yeah.. That's why I created the PR dotnet/aspnetcore/pull#57464, because at first I did not like either the reflection there involved..  But it's not approved yet, and if it was approved I do not think it'll get backported to dotnet8 or yet dotnet9.. Because it's not a critical Bug this one

jgarciadelanoceda:
I mark this as closed, it's a good addition, but there is risk involved

PR #3119, opened 2024-10-25
Avoid exception checking nullability

### LINKED ISSUE #2793: Fix issues with nullable types
Tracking issue for supporting nullable value/reference types correctly.

Refer to the details in the below issues to resolve this issue:

- #3118
- #3070
- #2877
- #2758
- #2731
- #2698
- #2680
- #2650
- #2633
- #2623
- #2587
- #2564
- #2503
- #2436
- #2225
- #2157
- #2110
- #2059
- #2036
- #2019

### LINKED ISSUE #3070: [Bug]: Wrong exception thrown by MemberInfoExtensions.IsDictionaryValueNonNullable
### Describe the bug

Starting from version 6.7.2, `MemberInfoExtensions.IsDictionaryValueNonNullable` contains a bug that throws an exception when the declaring type of a class property that is assignable to `Dictionary{TKey,TValue}` has a number of generic type arguments that is not 2. This is easily possible when the declaring type extends `Dictionary{TKey,TValue}`, e.g.

`public class MyDictionary: Dictionary<string, string>`

Instead of picking the generic type arguments of the declaring type, you need to look at the type arguments of the `IDictionary`-interface of the declaring type.

### Expected behavior

That no exception is thrown, just like in versions <= 6.7.1.

### Actual behavior

An exception is thrown on initialization, causing Swagger to fail to load API definition.

### Steps to reproduce

_No response_

### Exception(s) (if any)

_No response_

6.7.2

### .NET Version

8

### Anything else?

_No response_

martincostello:
- Do not throw if we cannot determine the nullability of a dictionary. It's better to potentially be wrong, rather than be unable to generate a document at all.
- Clean-up some code analysis suggestions.

Resolves #3070.
Resolves #2793 (as it's the last linked issue).

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/MemberInfoExtensions.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/MemberInfoExtensions.cs
@@ -85,16 +85,19 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
         {
 #if NET6_0_OR_GREATER
             var nullableInfo = GetNullabilityInfo(memberInfo);
-            if (nullableInfo.GenericTypeArguments.Length != 2)
-            {
-                var length = nullableInfo.GenericTypeArguments.Length;
-                var type = nullableInfo.Type.FullName;
-                var container = memberInfo.DeclaringType.FullName;
-                var member = memberInfo.Name;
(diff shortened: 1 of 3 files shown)


dldldlepl (approved):

mesMess (approved):
Code cleanup
Changing to concrete types (in-code) over var 👍🏼
Nullability exception check fix

PR #3125, opened 2024-11-02
Refactor filter descriptor type checks in SwaggerGen

iskandersierra:
# Pull Request

Given that the public API allows to register a filter in this form without any specific validation on not assigning the Type property:

```csharp
genOptions.OperationFilterDescriptors.Add(new FilterDescriptor
{
    // Type = typeof(TdpHeaderParameter),
    FilterInstance = new TdpHeaderParameter(IRequestContext.UserIdKey, "The user ID."),
});
```

in which case, opening the swagger UI returns a `NullReferenceException` (in developer error page) with no reference to user source code.

Looking into the library source code, I found out that the "proper" way to do the registration is:

```csharp
genOptions.AddOperationFilterInstance(
    new TdpHeaderParameter(IRequestContext.TenantIdKey, "The tenant ID."));
```

but I just felt that the first option should also be available with minimal behavior changes, given that it is a possible choice in the public API.

> NOTE: I didn't create a specific issue

I added an alternative validation on the filter type for when the `FilterInstance` is set but not the `Type` of the `FilterDescriptor`.

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenOptions.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenOptions.cs
@@ -32,5 +32,11 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
         public object[] Arguments { get; set; }
 
         public object FilterInstance { get; set; }
+
+        internal bool IsAssignableTo(Type type)
+        {
+            return FilterInstance != null && type.IsInstanceOfType(FilterInstance) ||
+                   Type != null && Type.IsAssignableTo(type);
+        }
     }
 }
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenOptions.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenOptions.cs
@@ -35,8 +35,8 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
 
         internal bool IsAssignableTo(Type type)
         {
-            return FilterInstance != null && type.IsInstanceOfType(FilterInstance) ||
-                   Type != null && Type.IsAssignableTo(type);
+            return (FilterInstance != null && type.IsInstanceOfType(FilterInstance)) ||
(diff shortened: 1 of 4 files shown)


src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenOptionsExtensions.cs
```diff
@@ -370,7 +370,6 @@ public static void AddSchemaFilterInstance<TFilter>(
             if (filterInstance == null) throw new ArgumentNullException(nameof(filterInstance));
             swaggerGenOptions.SchemaFilterDescriptors.Add(new FilterDescriptor
             {
-                Type = typeof(TFilter),
```
iskandersierra, line 373: There is not need to remove this assignments to property `Type`, even when this pass all the tests. Maybe someone out there is relying on the `Type` being assigned even when non needed by the library, because it is only used if the `FilterInstance` is not assigned.

I would revert the changes in this file.

martincostello: Please revert these changes - as you say, they are part of the visible external behaviour and someone might be using them.

src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenOptions.cs
```diff
@@ -32,5 +32,11 @@ public class FilterDescriptor
         public object[] Arguments { get; set; }
 
         public object FilterInstance { get; set; }
+
+        internal bool IsAssignableTo(Type type)
+        {
+            return FilterInstance != null && type.IsInstanceOfType(FilterInstance) ||
+                   Type != null && Type.IsAssignableTo(type);
```
martincostello, line 39: Add parenthesis to make the order of operations obvious:
```suggestion
            return (FilterInstance != null && type.IsInstanceOfType(FilterInstance)) ||
                   (Type != null && Type.IsAssignableTo(type));
```

martincostello (commented):
Can you add a test that validates the `NullReferenceException` has gone away with the change?

martincostello (approved):

PR #3133, opened 2024-11-08
Fix FromForm without WithOpenApi schemas

### LINKED ISSUE #3126: [Bug]: Incorrect swagger.json generation in Minimal API when using FromForm parameters without the WithOpenApi extension method
### Describe the bug

I have the following endpoints:

```
app.MapPost("/api/products", ([FromForm] Product product) =>
{
    return TypedResults.NoContent();
})
.DisableAntiforgery();

app.MapPost("/api/products-with-openapi", ([FromForm] Product product) =>
{
    return TypedResults.NoContent();
})
.DisableAntiforgery()
.WithOpenApi();

public record class Product(string Name, double Price);
```

These endpoints differ only for the `WithOpenApi` extension method, but the `swagger.json` definition is quite different:

```
"/api/products": {
  "post": {
    "tags": [
      "FromFormIssue"
    ],
    "requestBody": {
      "content": {
        "multipart/form-data": {
          "schema": {
            "required": [
              "product"
            ],
            "type": "object",
            "properties": {
              "product": {
                "$ref": "#/components/schemas/Product"
              }
            }
          },
          "encoding": {
            "product": {
              "style": "form"
            }
          }
        },
        "application/x-www-form-urlencoded": {
          "schema": {
            "required": [
              "product"
            ],
            "type": "object",
            "properties": {
              "product": {
                "$ref": "#/components/schemas/Product"
              }
            }
          },
          "encoding": {
            "product": {
              "style": "form"
            }
          }
        }
      }
    },
    "responses": {
      "204": {
        "description": "No Content"
      }
    }
  }
}


"/api/products-with-openapi": {
  "post": {
    "tags": [
      "FromFormIssue"
    ],
    "requestBody": {
      "content": {
        "multipart/form-data": {
          "schema": {
            "$ref": "#/components/schemas/Product"
          }
        },
        "application/x-www-form-urlencoded": {
          "schema": {
            "$ref": "#/components/schemas/Product"
          }
        }
      },
      "required": true
    },
    "responses": {
      "204": {
        "description": "No Content"
      }
    }
  }
}
```

So, in Swagger I get the following result:

WRONG
image

CORRECT
image

### Expected behavior

Both the endpoints should produce the same `swagger.json` definition that defines the parameters from `Form`.

### Actual behavior

_No response_

### Steps to reproduce

Minimal repro here: https://github.com/marcominerva/FromFormIssue

### Exception(s) (if any)

_No response_

6.9.0

### .NET Version

8.0.403

### Anything else?

_No response_

jgarciadelanoceda:
Fixes #3126.
Also I have found that there was an error for WithOpenApi for FromFormParameters that are enums

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
@@ -428,10 +428,9 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
                 foreach (var contentType in requestContentTypes)
                 {
                     var contentTypeValue = operation.RequestBody.Content[contentType];
-                    var fromFormParameters = apiDescription.ParameterDescriptions.Where(desc => desc.IsFromForm());
-                    var fromFormParametersCount = fromFormParameters.Count();
+                    var fromFormParameters = apiDescription.ParameterDescriptions.Where(desc => desc.IsFromForm()).ToList();
                     ApiParameterDescription bodyParameterDescription = null;
-                    if (fromFormParametersCount > 0)
+                    if (fromFormParameters.Count > 0)
                     {
                         var generatedContentTypeValue = GenerateRequestBodyFromFormParameters(
                             apiDescription,
@@ -455,7 +454,7 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
                         }
                     }
 
-                    if (fromFormParametersCount > 0 || bodyParameterDescription is not null)
+                    if (fromFormParameters.Count > 0 || bodyParameterDescription is not null)
                     {
                         var filterContext = new RequestBodyFilterContext(
                             bodyParameterDescription: bodyParameterDescription,
@@ -1059,6 +1058,7 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
 
             new KeyValuePair<string, string>("default", "Error")
         ];
+
 #if NET7_0_OR_GREATER
         private static string GenerateSummary(ApiDescription apiDescription) =>
             apiDescription.ActionDescriptor?.EndpointMetadata
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
@@ -445,7 +445,7 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
                     }
                     else
                     {
-                        bodyParameterDescription =  apiDescription.ParameterDescriptions.Single(desc => desc.IsFromBody());
+                        bodyParameterDescription = apiDescription.ParameterDescriptions.Single(desc => desc.IsFromBody());
                         if (bodyParameterDescription is not null)
                         {
                             contentTypeValue.Schema = GenerateSchema(
@@ -458,7 +458,6 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
 
                     if (fromFormParametersCount > 0 || bodyParameterDescription is not null)
                     {
-
                         var filterContext = new RequestBodyFilterContext(
                             bodyParameterDescription: bodyParameterDescription,
                             formParameterDescriptions: bodyParameterDescription is null ? fromFormParameters : null,
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
@@ -434,11 +434,10 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
                     if (fromFormParametersCount > 0)
                     {
                         var generatedRequestBody = GenerateRequestBodyFromFormParameters(
-                        apiDescription,
-                        schemaRepository,
-                        fromFormParameters,
-                        [contentType])
-                        .Content[contentType];
+                            apiDescription,
+                            schemaRepository,
+                            fromFormParameters,
+                            [contentType]).Content[contentType];
 
                         contentTypeValue.Schema = generatedRequestBody.Schema;
                         contentTypeValue.Encoding = generatedRequestBody.Encoding;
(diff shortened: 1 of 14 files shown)


src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
```diff
@@ -422,45 +422,44 @@ private async Task<OpenApiOperation> GenerateOpenApiOperationFromMetadataAsync(A
...
-                foreach (var content in requestContentTypes)
+                foreach (var contentType in requestContentTypes)
                 {
-                    content.Encoding = new Dictionary<string, OpenApiEncoding>();
-                    var requestParameters = apiDescription.ParameterDescriptions.Where(desc => desc.IsFromBody() || desc.IsFromForm());
-                    var countOfParameters = requestParameters.Count();
-                    if (countOfParameters > 0)
+                    var contentTypeValue = operation.RequestBody.Content[contentType];
+                    var fromFormParameters = apiDescription.ParameterDescriptions.Where(desc => desc.IsFromForm());
+                    var fromFormParametersCount = fromFormParameters.Count();
```
martincostello, line 432: Looks like we end up enumerating the collection multiple times in the case where `>0`, so we should just `ToList()` or `ToArray()` after the `Where()` and then use `Length`/`Count` and remove this count variable.

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
```diff
@@ -1042,7 +1059,6 @@ private static bool IsFromFormAttributeUsedWithIFormFile(ApiParameterDescription
 
             new KeyValuePair<string, string>("default", "Error")
         ];
-
```
martincostello, line 1045: Add this back please.

test/Swashbuckle.AspNetCore.IntegrationTests/SwaggerVerifyIntegrationTest.SwaggerEndpoint_ReturnsValidSwaggerJson_For_WebApi_swaggerRequestUri=v1.verified.txt
martincostello, line 1: Asking as it's hard to check just looking at raw OpenAPI specs - do they documents look "right" in Swagger UI and do the with and without `WithOpenApi` method usage like in the original issue produce the same operation descriptions in the OpenAPI document now?

test/WebSites/Basic/Controllers/FromFormParamsController.cs
```diff
@@ -27,6 +29,19 @@ public IActionResult PostFormWithIgnoredProperties([FromForm] RegistrationFormWi
         {
             throw new System.NotImplementedException();
         }
+        /// <summary>
```
martincostello, line 32:
```suggestion

        /// <summary>
```

test/WebSites/Basic/Controllers/FromFormParamsController.cs
```diff
@@ -43,6 +58,24 @@ public class RegistrationForm
...
+        /// Summary for Name
+        /// </summary>
+        /// <example>MyName</example>
+        public string Name { get; set; }
+
+        /// <summary>
+        /// Sumary for PhoneNumbers
+        /// </summary>
+        public IEnumerable<int> PhoneNumbers { get; set; }
```
martincostello, line 72:
```suggestion
        public IEnumerable<int> PhoneNumbers { get; set; }

```

test/Swashbuckle.AspNetCore.IntegrationTests/SwaggerVerifyIntegrationTest.SwaggerEndpoint_ReturnsValidSwaggerJson_For_WebApi_swaggerRequestUri=v1.verified.txt
jgarciadelanoceda: The WithOpenApi breaking change is on the second commit( You will see there the change).
The documents look right and they are well parsed

martincostello (approved):

martincostello:
Thanks - will merge this after #3007.

martincostello:
> First filter wins?

Probably. I did similar things in my new OpenAPI extensions NuGet package to not overwrite things in the OpenAPI document if they already had a value.

ymg2006:
@jgarciadelanoceda I am writing this because this commit removed enums from FromForm.
Here is simple example from v6.9.0 and 7.0.0 output:
v7.0.0: `"multipart/form-data": {`
`    "schema": {`
`        "allOf": [`
`            {`
`                "$ref": "#/components/schemas/TransactionTypeEnum"`
`            },`
`            {`
`                "required": [`
`                    "FiscalYearId"`
`                ],`
`                "type": "object",`
`                "properties": {`
`                    "FiscalYearId": {`
`                        "type": "integer",`
`                        "format": "int64"`
`                    },`
`                    "Id": {`
`                        "type": "integer",`
`                        "format": "int64"`
`                    }`
`                }`
`            }`
`        ]`
`    }`
`}`
v6.9.0: ` "multipart/form-data": {`
`                "schema": {`
`                    "required": [`
`                        "FiscalYearId",`
`                        "Type"`
`                    ],`
`                    "type": "object",`
`                    "properties": {`
`                        "Type": {`
`                            "$ref": "#/components/schemas/TransactionTypeEnum"`
`                        },`
`                        "FiscalYearId": {`
`                            "type": "integer",`
`                            "format": "int64"`
`                        }`
`                    }`
`                }`
`            }`
This makes nswag to remove Type this from output api.
   createOrEdit(type?: TransactionTypeEnum | undefined, fiscalYearId: number): Observable<boolean>;
   createOrEdit(fiscalYearId: number): Observable<boolean>;
I don't understand why Enum should be removed from required and moved to allOf, isn't it a type? can't enums be passed optional?
@martincostello Please revert this commit until the contributor fixes issues with it.

martincostello:
Please create a new issue for this regression with a fully reproducible example, rather than commenting on a complete PR.

The change won't be reverted, we'll just fix-forward.

ymg2006:
> Please create a new issue for this regression with a fully reproducible example, rather than commenting on a complete PR.
> 
> The change won't be reverted, we'll just fix-forward.

@martincostello  I created an issue, that is a simple from form with enum. it will produce those outputs.

PR #3135, opened 2024-11-08
Initialize Urls property with an empty collection

averichev:
Updated the `Urls` property in `ConfigObject` to initialize with `Enumerable.Empty<UrlDescriptor>()` instead of `null`. This change ensures the property is always a valid collection, eliminating the need for null checks and preventing potential null reference exceptions when accessing or iterating over the property. Improves code safety and aligns with best practices for handling collections.


martincostello:
I don't have a problem with this _unless_ Swagger UI treats empty and null/undefined differently.

> This change ensures the property is always a valid collection

With the qualifier _"by default"_ - there's nothing to stop it being null.

PR #3143, opened 2024-11-13
Update some nugets

Saibamen:
Update some nuget packages

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -1,6 +1,6 @@
 <Project>
   <ItemGroup>
-    <PackageVersion Include="Autofac.Extensions.DependencyInjection" Version="10.0.0" />
+    <PackageVersion Include="Autofac.Extensions.DependencyInjection" Version="4.2.2" />
     <PackageVersion Include="BenchmarkDotNet" Version="0.14.0" />
     <PackageVersion Include="coverlet.msbuild" Version="6.0.2" />
     <PackageVersion Include="GitHubActionsTestLogger" Version="2.4.1" />
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -28,7 +28,7 @@
     <PackageVersion Include="NSubstitute" Version="5.3.0" />
     <PackageVersion Include="NSwag.MSBuild" Version="14.1.0" />
     <PackageVersion Include="ReportGenerator" Version="5.4.1" />
-    <PackageVersion Include="System.Text.Json" Version="4.7.2" />
+    <PackageVersion Include="System.Text.Json" Version="4.6.0" />
     <PackageVersion Include="Verify.Xunit" Version="28.2.1" />
     <PackageVersion Include="xunit" Version="2.9.2" />
     <PackageVersion Include="xunit.core" Version="2.9.2" />
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -27,7 +27,7 @@
     <PackageVersion Include="Newtonsoft.Json" Version="13.0.3" />
     <PackageVersion Include="NSubstitute" Version="5.3.0" />
     <PackageVersion Include="NSwag.MSBuild" Version="14.1.0" />
-    <PackageVersion Include="ReportGenerator" Version="5.3.11" />
+    <PackageVersion Include="ReportGenerator" Version="5.4.0" />
     <PackageVersion Include="System.Text.Json" Version="9.0.0" />
     <PackageVersion Include="Verify.Xunit" Version="28.2.1" />
     <PackageVersion Include="xunit" Version="2.9.2" />
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -27,7 +27,7 @@
     <PackageVersion Include="Newtonsoft.Json" Version="13.0.3" />
     <PackageVersion Include="NSubstitute" Version="5.3.0" />
     <PackageVersion Include="NSwag.MSBuild" Version="14.1.0" />
-    <PackageVersion Include="ReportGenerator" Version="5.4.0" />
+    <PackageVersion Include="ReportGenerator" Version="5.4.1" />
     <PackageVersion Include="System.Text.Json" Version="4.7.2" />
     <PackageVersion Include="Verify.Xunit" Version="28.2.1" />
     <PackageVersion Include="xunit" Version="2.9.2" />
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -7,17 +7,17 @@
     <PackageVersion Include="IdentityServer4" Version="3.1.4" />
     <PackageVersion Include="IdentityServer4.AccessTokenValidation" Version="3.0.1" />
     <PackageVersion Include="Microsoft.AspNetCore" Version="2.1.0" />
-    <PackageVersion Include="Microsoft.AspNetCore.Mvc.ApiExplorer" Version="2.1.3" />
-    <PackageVersion Include="Microsoft.AspNetCore.Mvc.DataAnnotations" Version="2.1.3" />
-    <PackageVersion Include="Microsoft.AspNetCore.Mvc.Formatters.Json" Version="2.1.18" />
+    <PackageVersion Include="Microsoft.AspNetCore.Mvc.ApiExplorer" Version="2.1.0" />
+    <PackageVersion Include="Microsoft.AspNetCore.Mvc.DataAnnotations" Version="2.1.0" />
+    <PackageVersion Include="Microsoft.AspNetCore.Mvc.Formatters.Json" Version="2.1.0" />
     <PackageVersion Include="Microsoft.AspNetCore.Mvc.NewtonsoftJson" Version="8.0.7" />
     <PackageVersion Include="Microsoft.AspNetCore.Mvc.Testing" Version="6.0.32" />
     <PackageVersion Include="Microsoft.AspNetCore.Mvc.Versioning" Version="5.1.0" />
     <PackageVersion Include="Microsoft.AspNetCore.Mvc.Versioning.ApiExplorer" Version="5.1.0" />
     <PackageVersion Include="Microsoft.AspNetCore.OpenApi" Version="8.0.7" />
     <PackageVersion Include="Microsoft.AspNetCore.Razor.Design" Version="2.1.2" />
-    <PackageVersion Include="Microsoft.AspNetCore.Routing" Version="2.1.1" />
-    <PackageVersion Include="Microsoft.AspNetCore.StaticFiles" Version="2.1.1" />
+    <PackageVersion Include="Microsoft.AspNetCore.Routing" Version="2.1.0" />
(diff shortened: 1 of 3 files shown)


Directory.Packages.props
```diff
@@ -1,37 +1,37 @@
 <Project>
   <ItemGroup>
-    <PackageVersion Include="Autofac.Extensions.DependencyInjection" Version="4.2.2" />
+    <PackageVersion Include="Autofac.Extensions.DependencyInjection" Version="10.0.0" />
```
martincostello, line 3: The main reason I hadn't previously updated this was because the version jump was so large I didn't want to break things and I was unsure whether the existing tests would find any breakage - have you checked this works?

martincostello (changes requested):

Directory.Packages.props
```diff
@@ -1,37 +1,37 @@
...
     <PackageVersion Include="GitHubActionsTestLogger" Version="2.4.1" />
     <PackageVersion Include="IdentityServer4" Version="3.1.4" />
     <PackageVersion Include="IdentityServer4.AccessTokenValidation" Version="3.0.1" />
     <PackageVersion Include="Microsoft.AspNetCore" Version="2.1.0" />
-    <PackageVersion Include="Microsoft.AspNetCore.Mvc.ApiExplorer" Version="2.1.0" />
-    <PackageVersion Include="Microsoft.AspNetCore.Mvc.DataAnnotations" Version="2.1.0" />
-    <PackageVersion Include="Microsoft.AspNetCore.Mvc.Formatters.Json" Version="2.1.0" />
+    <PackageVersion Include="Microsoft.AspNetCore.Mvc.ApiExplorer" Version="2.2.0" />
+    <PackageVersion Include="Microsoft.AspNetCore.Mvc.DataAnnotations" Version="2.2.0" />
+    <PackageVersion Include="Microsoft.AspNetCore.Mvc.Formatters.Json" Version="2.2.0" />
```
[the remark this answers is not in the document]
Saibamen: OK, I will revert it, but don't we want to replace these packages? (see: https://github.com/dotnet/announcements/issues/217)

Directory.Packages.props
```diff
@@ -1,37 +1,37 @@
...
-    <PackageVersion Include="NSubstitute" Version="5.1.0" />
+    <PackageVersion Include="NSubstitute" Version="5.3.0" />
     <PackageVersion Include="NSwag.MSBuild" Version="14.1.0" />
-    <PackageVersion Include="ReportGenerator" Version="5.3.8" />
-    <PackageVersion Include="System.Text.Json" Version="4.6.0" />
-    <PackageVersion Include="Verify.Xunit" Version="26.1.6" />
-    <PackageVersion Include="xunit" Version="2.9.0" />
-    <PackageVersion Include="xunit.core" Version="2.9.0" />
+    <PackageVersion Include="ReportGenerator" Version="5.3.11" />
+    <PackageVersion Include="System.Text.Json" Version="9.0.0" />
```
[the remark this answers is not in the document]
Saibamen: Maybe latest 4? `4.7.2` from 12.05.2020?

Directory.Packages.props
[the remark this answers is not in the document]
martincostello: They're only used by the `netstandard2.0` target, which is for users of .NET Framework. Newer versions of .NET will get different versions based on package overrides in the individual `.csproj` files based on the TFM being compiled for. For example:

https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/be0090d5b94dbf7107cbbf583bc4bdd868db60e6/src/Swashbuckle.AspNetCore.Newtonsoft/Swashbuckle.AspNetCore.Newtonsoft.csproj#L16-L30

Directory.Packages.props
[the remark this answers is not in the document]
martincostello: I'd prefer to leave it as it is unless it fixes a demonstrable issue that cannot be resolved by updating the version in the consuming application itself.

Directory.Packages.props
```diff
@@ -1,37 +1,37 @@
...
     <PackageVersion Include="GitHubActionsTestLogger" Version="2.4.1" />
     <PackageVersion Include="IdentityServer4" Version="3.1.4" />
     <PackageVersion Include="IdentityServer4.AccessTokenValidation" Version="3.0.1" />
     <PackageVersion Include="Microsoft.AspNetCore" Version="2.1.0" />
-    <PackageVersion Include="Microsoft.AspNetCore.Mvc.ApiExplorer" Version="2.1.0" />
-    <PackageVersion Include="Microsoft.AspNetCore.Mvc.DataAnnotations" Version="2.1.0" />
-    <PackageVersion Include="Microsoft.AspNetCore.Mvc.Formatters.Json" Version="2.1.0" />
+    <PackageVersion Include="Microsoft.AspNetCore.Mvc.ApiExplorer" Version="2.1.3" />
+    <PackageVersion Include="Microsoft.AspNetCore.Mvc.DataAnnotations" Version="2.1.3" />
+    <PackageVersion Include="Microsoft.AspNetCore.Mvc.Formatters.Json" Version="2.1.18" />
```
martincostello, line 12:
```suggestion
    <PackageVersion Include="Microsoft.AspNetCore.Mvc.ApiExplorer" Version="2.1.0" />
    <PackageVersion Include="Microsoft.AspNetCore.Mvc.DataAnnotations" Version="2.1.0" />
    <PackageVersion Include="Microsoft.AspNetCore.Mvc.Formatters.Json" Version="2.1.0" />
```

Directory.Packages.props
```diff
@@ -1,37 +1,37 @@
...
     <PackageVersion Include="Microsoft.AspNetCore.Mvc.NewtonsoftJson" Version="8.0.7" />
     <PackageVersion Include="Microsoft.AspNetCore.Mvc.Testing" Version="6.0.32" />
     <PackageVersion Include="Microsoft.AspNetCore.Mvc.Versioning" Version="5.1.0" />
     <PackageVersion Include="Microsoft.AspNetCore.Mvc.Versioning.ApiExplorer" Version="5.1.0" />
     <PackageVersion Include="Microsoft.AspNetCore.OpenApi" Version="8.0.7" />
     <PackageVersion Include="Microsoft.AspNetCore.Razor.Design" Version="2.1.2" />
-    <PackageVersion Include="Microsoft.AspNetCore.Routing" Version="2.1.0" />
-    <PackageVersion Include="Microsoft.AspNetCore.StaticFiles" Version="2.1.0" />
+    <PackageVersion Include="Microsoft.AspNetCore.Routing" Version="2.1.1" />
+    <PackageVersion Include="Microsoft.AspNetCore.StaticFiles" Version="2.1.1" />
```
martincostello, line 20:
```suggestion
    <PackageVersion Include="Microsoft.AspNetCore.Routing" Version="2.1.0" />
    <PackageVersion Include="Microsoft.AspNetCore.StaticFiles" Version="2.1.0" />
```

Directory.Packages.props
```diff
@@ -1,37 +1,37 @@
...
     <PackageVersion Include="Newtonsoft.Json" Version="13.0.3" />
-    <PackageVersion Include="NSubstitute" Version="5.1.0" />
+    <PackageVersion Include="NSubstitute" Version="5.3.0" />
     <PackageVersion Include="NSwag.MSBuild" Version="14.1.0" />
-    <PackageVersion Include="ReportGenerator" Version="5.3.8" />
-    <PackageVersion Include="System.Text.Json" Version="4.6.0" />
-    <PackageVersion Include="Verify.Xunit" Version="26.1.6" />
-    <PackageVersion Include="xunit" Version="2.9.0" />
-    <PackageVersion Include="xunit.core" Version="2.9.0" />
+    <PackageVersion Include="ReportGenerator" Version="5.4.0" />
```
martincostello, line 30:
```suggestion
    <PackageVersion Include="ReportGenerator" Version="5.4.1" />
```

Directory.Packages.props
```diff
@@ -1,37 +1,37 @@
...
+    <PackageVersion Include="System.Text.Json" Version="4.7.2" />
```
martincostello, line 31:
```suggestion
    <PackageVersion Include="System.Text.Json" Version="4.6.0" />
```

Directory.Packages.props
```diff
@@ -1,37 +1,37 @@
...
+    <PackageVersion Include="Verify.Xunit" Version="28.2.1" />
+    <PackageVersion Include="xunit" Version="2.9.2" />
+    <PackageVersion Include="xunit.core" Version="2.9.2" />
     <PackageVersion Include="xunit.runner.visualstudio" Version="2.8.2" />
   </ItemGroup>
-</Project>
+</Project>
```
martincostello, line 37: Please undo this change.

Directory.Packages.props
Saibamen: @martincostello: Why we don't want to have latest 2.1.x versions here? They are not marked as deprecated in nuget.org: https://www.nuget.org/packages/Microsoft.AspNetCore.StaticFiles

Directory.Packages.props
Saibamen: `Autofac.Extensions.DependencyInjection` is used only in one project with tests:

image

Directory.Packages.props
martincostello: It's not Swashbuckle's business to force specific versions of dependencies onto our consumers - applications who want the latest versions can update their own application to do so. It makes reasoning about compatibility for use with the hundreds of projects who consume our projects much easier to deal with.

Just the test-dependencies that dependabot is failing to update for us (in an idea world this PR wouldn't be needed at all) are what should be updated.

Directory.Packages.props
martincostello: > I was unsure whether the existing tests would find any breakage

Sure, but I don't know if our tests _actually_ cover its usage - if they don't, then them being green is meaningless. Could you look into it and determine whether or not what we already have is "good enough" to be confident in the upgrade?

I joined the project this year, so I have no prior knowledge to answer this question.

martincostello:
LGTM pending the question about the Autofac upgrade.

Saibamen:
@martincostello: I updated `xUnit` package to `2.9.2`, but in `Swashbuckle.AspNetCore.ApiTesting.Xunit` project it is `2.4.1`. Should I delete `VersionOverride` tag?

`<PackageReference Include="xunit" VersionOverride="2.4.1" />`

martincostello:
No, leave all the version overrides alone - otherwise a future release of our xunit package would force an upgrade on people that they might not otherwise want.

martincostello (approved):

Directory.Packages.props
Saibamen: This `CliExampleWithFactory` project is not working on `master` - it has 4 `DependencyResolutionException` exceptions at startup. This is why it is not included in any tests.

I reverted this Autofac update and this should be fixed in separate PR together with new integration tests

martincostello:
Thanks for being our human dependabot 🤖 😄

PR #3150, opened 2024-11-16
Adding dynamic RoutePrefixes support

### LINKED ISSUE #3144: [Bug]: SwaggerUIMiddleware is now internal
### Describe the bug

With the release of Swashbuckle 7 `SwaggerUIMiddleware` became internal. This is blocking us from upgrade.

We have a multi tenant solution and to make swagger run a little bit different for each tenant we build a custom routing to end up the swagger ui. The url became <our url>/tenantname/swagger.

We achieved that by a Post Configure on the `SwaggerUIOptions` and have our own scoped middleware that passed this settings object to the `SwaggerUIMiddleware`. But now the middleware became internal this is no longer possible.

We are not really aware of other approaches that let us run multiple  a swagger UI per tenant

### Expected behavior

SwaggerUIMiddleware is public

### Actual behavior

SwaggerUIMiddleware is internal

### Steps to reproduce

_No response_

### Exception(s) (if any)

_No response_

_No response_

### .NET Version

_No response_

### Anything else?

_No response_

davesmits:
# Pull Request

Fixes #3144

Adding callback to give a valid routeprefix based on the http context

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerUI/SwaggerUIMiddleware.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerUI/SwaggerUIMiddleware.cs
@@ -27,8 +27,11 @@ namespace Swashbuckle.AspNetCore.SwaggerUI
     {
         private const string EmbeddedFileNamespace = "Swashbuckle.AspNetCore.SwaggerUI.node_modules.swagger_ui_dist";
 
+        private readonly RequestDelegate _next;
+        private readonly IWebHostEnvironment _hostingEnv;
         private readonly SwaggerUIOptions _options;
         private readonly StaticFileMiddleware _staticFileMiddleware;
+        private readonly ILoggerFactory _loggerFactory;
         private readonly JsonSerializerOptions _jsonSerializerOptions;
 
         public SwaggerUIMiddleware(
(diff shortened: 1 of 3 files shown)


src/Swashbuckle.AspNetCore.SwaggerUI/SwaggerUIMiddleware.cs
```diff
@@ -65,12 +74,16 @@ public async Task Invoke(HttpContext httpContext)
...
+                                    _options.GetDynamicRoutePrefix(httpContext) :
+                                    _options.RoutePrefix;
+
             if (HttpMethods.IsGet(httpMethod))
             {
                 var path = httpContext.Request.Path.Value;
 
                 // If the RoutePrefix is requested (with or without trailing slash), redirect to index URL
-                if (Regex.IsMatch(path, $"^/?{Regex.Escape(_options.RoutePrefix)}/?$", RegexOptions.IgnoreCase))
+                if (Regex.IsMatch(path, $"^/?{Regex.Escape(routePrefix)}/?$", RegexOptions.IgnoreCase))
```
martincostello, line 86: I was thinking more something like `Func<HttpContext, bool>` or `Func<HttpRequest, bool>` to match against, then it doesn't necessarily have to be something to do with Regex if the user wants something custom.

PR #3155, opened 2024-11-19
second level inheritance for UseAllOf

k0ka:

The `UseAllOfForInheritance` option of schema generator only accounts for direct parents. It makes impossible to generate correct discriminator information for a big inheritance tree. 

## Example
data class structure:
```C#
[SwaggerDiscriminator("discriminator")]
[SwaggerSubType(typeof(SubSubType), DiscriminatorValue = nameof(SubSubType))]
abstract public class BaseType
{
}

abstract public class SubType : BaseType
{
}

public class SubSubType : SubType
{
    public int Property { get; set; }
}
```

Swagger generator options:
```C#
builder.Services.AddSwaggerGen(
    options =>
    {
        options.UseAllOfForInheritance();
        options.UseOneOfForPolymorphism();
        options.EnableAnnotations(true, true);
    });
```

Generated components of the json:
```json
"components": {
    "schemas": {
      "SubSubType": {
        "type": "object",
        "properties": {
          "property": {
            "type": "integer",
            "format": "int32"
          }
        },
        "additionalProperties": false
      }
    }
  }
```

It doesn't include `discriminator` and `BaseType` definition.

If we remove `SubType` and make `SubSubType` child of `BaseType` we would get the expected generated components:
```json
"components": {
    "schemas": {
      "BaseType": {
        "required": [
          "discriminator"
        ],
        "type": "object",
        "properties": {
          "discriminator": {
            "type": "string"
          }
        },
        "additionalProperties": false,
        "discriminator": {
          "propertyName": "discriminator",
          "mapping": {
            "SubSubType": "#/components/schemas/SubSubType"
          }
        }
      },
      "SubSubType": {
        "allOf": [
          {
            "$ref": "#/components/schemas/BaseType"
          },
          {
            "type": "object",
            "properties": {
              "property": {
                "type": "integer",
                "format": "int32"
              }
            },
            "additionalProperties": false
          }
        ]
      }
    }
  }
```

This PR checks all parents until it finds the one which has this class defined as sub type. I added same assertions to test with one level assertions.

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/SchemaGenerator.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/SchemaGenerator.cs
@@ -475,12 +475,18 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
             baseTypeDataContract = null;
 
             var baseType = dataContract.UnderlyingType.BaseType;
+            while (baseType != null && baseType != typeof(object))
+            {
+                if (_generatorOptions.SubTypesSelector(baseType).Contains(dataContract.UnderlyingType))
+                {
+                    baseTypeDataContract = GetDataContractFor(baseType);
+                    return true;
+                }
 
-            if (baseType == null || baseType == typeof(object) || !_generatorOptions.SubTypesSelector(baseType).Contains(dataContract.UnderlyingType))
-                return false;
+                baseType = baseType.BaseType;
+            }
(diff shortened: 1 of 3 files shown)


martincostello (commented):
Could you also add an example like in the issue description to one of our integration test projects so we have a verification test for this scenario too please?

martincostello (approved):

Eloyjr27 (commented):
ok, all is perfect

Issue #1954, opened 2021-01-13
Support applicable attributes from System.ComponentModel namespace


domaindrivendev:
For example:

- [DescriptionAttribute](https://docs.microsoft.com/en-us/dotnet/api/system.componentmodel.descriptionattribute?view=net-5.0)
- [ReadOnlyAttribute](https://docs.microsoft.com/en-us/dotnet/api/system.componentmodel.readonlyattribute?view=net-5.0)

mikhail-barg:
Is there any progress on this issue? It would be really good to have this because it allows to use framework-defined attributes to annotate classes, without a need to add additional assembly. 

Consider the following setup. There's a common library with business objects, having all the descriptions. And there's a number of services using this library, one of those provides web api with CRUD operations over the business objects, but other services have no relation to OpenAPI, and web at all.

So to have the API service display a nice swagger UI it's necessary to add a reference to `Swashbuckle.AspNetCore.Annotations` to the common library, which would then propagate to all dependent services.

mikhail-barg:
I've looked through the sources and was quite puzzled because `Swashbuckle.AspNetCore.SwaggerGen.SchemaGenerator` does support `RequiredAttribute` and `DefaultValueAttribute` from System.ComponentModel. So it's unclear why not support other ones?

Anyway I've quickly mocked a filter (based on `Swashbuckle.AspNetCore.Annotations.AnnotationsSchemaFilter`) than reads the remaining attributes. So maybe it would help you in some way to speedup the implementation in the library.

[ComponentModelSchemaFilter.cs.txt](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/files/9541149/ComponentModelSchemaFilter.cs.txt)

ouvreboite:
@domaindrivendev I’m part of the API team at Criteo, and as such we provide common, generic, models to be used in our various internal and external APIs to have an homogenous input/output/error structure.

As such we currently have to rely an xmlDoc/filters/customSchemaIds to fine tune the description and schemaIds for our generic models. But this require internal teams to correctly reference the correct assembly and inject the correct filters and such. And it’s quite hard to fully enforce that.

So relying on DescriptionAttribute for schema/field description and DisplayNameAttribute for schemaId would be beneficial as those would not require any custom configuration.

If the support of those attributes is still in line with the Swashbuckle roadmap, I would be happy to work on a PR :)

PR #3162, opened 2024-11-20
Support DescriptionAttribute and ReadOnlyAttribute

jgarciadelanoceda:
Gives support to #1954 
I do not really know what is intended with DisplayName attribute, but I am sure someone can give me advice in what to do with that attribute

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/OpenApiSchemaExtensions.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/OpenApiSchemaExtensions.cs
@@ -1,4 +1,5 @@
 ﻿using System.Collections.Generic;
+using System.ComponentModel;
 using System.ComponentModel.DataAnnotations;
 using Microsoft.AspNetCore.Mvc.ApiExplorer;
 using Microsoft.AspNetCore.Routing.Constraints;
@@ -60,6 +61,13 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
 
                 else if (attribute is StringLengthAttribute stringLengthAttribute)
                     ApplyStringLengthAttribute(schema, stringLengthAttribute);
+
+                else if (attribute is ReadOnlyAttribute readOnlyAttribute)
+                    ApplyReadOnlyAttribute(schema, readOnlyAttribute);
+
+                else if (attribute is DescriptionAttribute descriptionAttribute)
+                    ApplyDescriptionAttribute(schema, descriptionAttribute);
+
             }
         }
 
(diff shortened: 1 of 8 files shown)


martincostello:
> I do not really know what is intended with DisplayName attribute

IIRC, it's typically used to provide non-code text for UI elements for code members. [For example](https://github.com/dotnet/runtime/blob/98049e56a415d9bcf2d002fba71f399ba08cf39c/src/libraries/System.ComponentModel.Annotations/tests/System/ComponentModel/DataAnnotations/AssociatedMetadataTypeTypeDescriptionProviderTests.cs#L306-L307):

```csharp
[DisplayName("First name")]
public string FirstName { get; set; }
```

I've seen them commonly used to for the Visual Studio designer.

For example, [this code](https://github.com/dotnet/winforms/blob/dd5f8e7093051e80887184889c63e0da2203bd6d/src/System.Windows.Forms/src/System/Windows/Forms/Panels/TableLayoutPanel/TableLayoutPanel.cs#L150-L156) renders how property windows like the one below Visual Studio.

image

src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/OpenApiSchemaExtensions.cs
```diff
@@ -230,6 +238,16 @@ private static void ApplyStringLengthAttribute(OpenApiSchema schema, StringLengt
...
         }
 
+        private static void ApplyReadOnlyAttribute(OpenApiSchema schema, ReadOnlyAttribute readOnlyAttribute)
+        {
+            schema.ReadOnly = readOnlyAttribute.IsReadOnly;
+        }
+
+        private static void ApplyDescriptionAttribute(OpenApiSchema schema, DescriptionAttribute descriptionAttribute)
+        {
+            schema.Description = descriptionAttribute.Description;
```
martincostello, line 248: I wonder if we should change this to `??=` so we don't overwrite any existing value?

test/Swashbuckle.AspNetCore.Newtonsoft.Test/SchemaGenerator/NewtonsoftSchemaGeneratorTests.cs
```diff
@@ -788,7 +790,7 @@ public void GenerateSchema_HonorsSerializerAttribute_JsonRequired()
             var referenceSchema = Subject().GenerateSchema(typeof(JsonRequiredAnnotatedType), schemaRepository);
 
             var schema = schemaRepository.Schemas[referenceSchema.Reference.Id];
-            Assert.Equal(new[] { "StringWithConflictingRequired", "StringWithJsonRequired"}, schema.Required.ToArray());
+            Assert.Equal(new[] { "StringWithConflictingRequired", "StringWithJsonRequired" }, schema.Required.ToArray());
```
martincostello, line 793: Can this use collection expressions too?

```suggestion
            Assert.Equal(["StringWithConflictingRequired", "StringWithJsonRequired"], schema.Required);
```

jgarciadelanoceda:
> IIRC, it's typically used to provide non-code text for UI elements for code members. [For example](https://github.com/dotnet/runtime/blob/98049e56a415d9bcf2d002fba71f399ba08cf39c/src/libraries/System.ComponentModel.Annotations/tests/System/ComponentModel/DataAnnotations/AssociatedMetadataTypeTypeDescriptionProviderTests.cs#L306-L307):

I do not see the need for integrating this attribute, because there is already integrated by the serializers the JsonPropertyName and for Query/Path variables you can put the name of the parameter overriding the name of the object like what we do here:
https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/340eef3e812cc012a16c70e183d748aebf3a898e/test/WebSites/Basic/Controllers/CrudActionsController.cs#L41

martincostello (approved):

Issue #3165, opened 2024-11-22
[Feature request]: Distribute debug symbols/PDB files


ldeluigi:
### Is your feature request related to a specific problem? Or an existing feature?

I am trying to debug Swashbuckle behavior but can't due to debug symbols missing

### Describe the solution you'd like

Distributing PDB files/debug symbols in BuGet enables to debug swashbuckle from code dpendning on it

martincostello:
I'll have to check when I'm at a computer, but there should be embedded symbols in the assemblies so you should be able to debug without PDB files.

ldeluigi:
True but the experience is not very seamless and clean

martincostello:
That sounds more like a problem with whatever tooling you're using, than something we need to worry about.

I'm fairly sure the setup here is the same one I put into Polly, and last time I debugged through into non user code in Visual Studio it all "just worked".

Again, I'll check it for myself when I'm at a computer, but if I get the experience I'm expecting to get with the latest version of Swashbuckle, then it's working as intended.

ldeluigi:
You can close the issue if it's the case

martincostello:
Actually, no, I think you're right and it's not setup correctly - I just noticed that's the disassembled code, not the _actual_ code.

Compare with the real top-of-the-method:

https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/bb498849621f36ea17d83e76631869c240732b0d/src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenServiceCollectionExtensions.cs#L17-L19

Saibamen:
snupkg files are created and published for all packages except `Swashbuckle.AspNetCore`:

![image](https://github.com/user-attachments/assets/d7dcf426-e861-486c-9ee4-bbe2d030dcd2)

martincostello:
I'm going to guess that's a detail of the fact it's a meta-package and doesn't contain any actual code.

PR #3174, opened 2024-11-25
Push NuGet symbols to NuGet.org

Saibamen:
Symbols are not published in nuget.org after https://github.com/domaindrivendev/Swashbuckle.AspNetCore/pull/3168:
https://www.nuget.org/packages/Swashbuckle.AspNetCore/
image

Fixes https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/3165


Saibamen:
This could be one command for both nupkg and snupkg when using `nuget pack` command

Saibamen:
But this need testing

martincostello:
Symbol packages are automatically discovered and published when found. You can see them being uploaded in the logs from the tagged build earlier today.

PR #3183, opened 2024-11-27
Drop .NET 6

### LINKED ISSUE #3087: Drop support for .NET 6
#2798 drops support for _most_ out-of-support versions of .NET, but to give users time to react, version 7.0.0 retains support for `net6.0` even though by the time it ships .NET 6 will be out of support.

We should drop support for `net6.0` in a release some time afterwards.

See also https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/2798#issuecomment-2060853358.

martincostello:
Drop support for .NET 6.

Resolves #3087.

Leaving in draft for now - it's best to ship after a period of stability where we've fixed any known regressions caused by v7.


README.md
```diff
@@ -1631,7 +1631,7 @@ Once your application has been setup with Swashbuckle (see [Getting Started](#ge
 
 It's packaged as a [.NET Tool](https://learn.microsoft.com/dotnet/core/tools/global-tools) that can be installed and used via the dotnet SDK.
 
-> :warning: The tool needs to load your Startup DLL and its dependencies at runtime. Therefore, you should use a version of the `dotnet` SDK that is compatible with your application. For example, if your app targets `net6.0`, then you should use version 6.0.xxx of the SDK to run the CLI tool. If it targets `net8.0`, then you should use version 8.0.xxx of the SDK and so on.
+> :warning: The tool needs to load your Startup DLL and its dependencies at runtime. Therefore, you should use a version of the `dotnet` SDK that is compatible with your application. For example, if your app targets `net8.0`, then you should use version 8.0.xxx of the SDK to run the CLI tool. If it targets `net8.0`, then you should use version 8.0.xxx of the SDK and so on.
```
Saibamen, line 1634:
```suggestion
> :warning: The tool needs to load your Startup DLL and its dependencies at runtime. Therefore, you should use a version of the `dotnet` SDK that is compatible with your application. For example, if your app targets `net8.0`, then you should use version 8.0.xxx of the SDK to run the CLI tool. If it targets `net9.0`, then you should use version 9.0.xxx of the SDK and so on.
```

Saibamen:
Please remove all `#if NET6_0_OR_GREATER` and `#if NET7_0_OR_GREATER`.

And if these IF statements have `ELSE`, update check to `#if NET8_0_OR_GREATER`

martincostello:
No, that's unnecessary code churn that makes no difference.

mesMess (approved):

PR #3207, opened 2024-12-20
Update system text json to non vulnerable version

Adishone:
# Pull Request

Used System.Text.Json package is really outdated and has known vulnerabilties

System.Text.Json is known to have vulnerabilities so I'm updating it to version that has them fixed
image

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -28,7 +28,7 @@
     <PackageVersion Include="NSubstitute" Version="5.3.0" />
     <PackageVersion Include="NSwag.MSBuild" Version="14.1.0" />
     <PackageVersion Include="ReportGenerator" Version="5.4.1" />
-    <PackageVersion Include="System.Text.Json" Version="4.6.0" />
+    <PackageVersion Include="System.Text.Json" Version="8.0.5" />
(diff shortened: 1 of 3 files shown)


martincostello:
Thanks for your PR - this package is specifically pinned to the minimum compatible version because:

1. It's only used for 3 packages, and then only for the `netstandard2.0` target framework - other targets will use the in-box version of System.Text.Json, whatever it happens to be for the consuming application.
2. It avoids creating any compatibility issues from a giant leap forward in the version of System.Text.Json (4 major versions, or 5 if we used 9.0.0 instead) used with .NET Framework applications. Testing that such a change doesn't break things is too much of a burden for us to test.
3. We don't want to create churn by having to release a new version of Swashbuckle every time there's a vulnerability in a dependency we happen to use.

Users who aren't otherwise already using a newer patched version in their application's dependency tree can add an explicit reference to a newer version.

You can see which packages we pin here:

https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/4ae6541da310f458bc1ebeae5eee4eb9b3211c68/.github/dependabot.yml#L42-L58

PR #3234, opened 2025-01-25
Humanize multiline para tag

### LINKED ISSUE #3009: [Bug]: Multi-line paragraphs render as pre code html tag pair
### Describe the bug

Given XML comments like
```c#
    /// <remarks>
    ///     <para>
    ///         If the user already exists, returns an HTTP 409 and the existing user profile. If the
    ///         user cannot be created due the email domain being on the block list, an HTTP 406 is returned.  If the user
    ///         cannot be created because the email domain cannot be legally processed (ex. OFAC compliance), then an HTTP 451
    ///         code is returned
    ///     </para>
    ///     <para>
    ///         If the user is in the STAGED or PROVISIONED state, and the user's mail domain is NOT federated to an external
    ///         identity provider (ex. by SAML), the Activation property along with its contained ActivationUrl property will
    ///         be set.  The ActivationUrl may be included as a link in a "Welcome" email to the user.  When the user navigates
    ///         to that URL, they will be asked to set their password.
    ///         **The client application should append a fromURI query string parameter (case-sensitive)**,
    ///         which will cause the user to be redirected to an appropriate page, after setting their
    ///         password.
    ///     </para>
    ///     <para>
    ///         Users associated with federated identity providers do not need to be activated or sent an activation link. The
    ///         user will be activated automatically once authenticated by their identity provider.  The user may be sent an
    ///         email with a URL link that sends them directly to the application.
    ///     </para>
    ///     <para>security group: identity_user_create</para>
```
When the paragraph is multiple lines and the `<para>` tag stands on its own line.  The paragraph is being rendered as a `<pre><code>` HTML pair, which I believe is incorrect, since there is a `<code>` tag for XML comments that should equate to HTML `<pre><code>` (or perhaps just `code`).

Aside from the glaring difference in formatting between the paragraphs, markdown such a bold (two stars) does not get rendered in the `<pre><code>` sections, as shown in this screenshot.

image

### Expected behavior

New lines within a `<para>` tag should not be included in the generated JSON and `<para>` tags should be represented as `\n\r` in the generated json.  This would provide the expected rendering.

### Actual behavior

`<para>` tags in XML comments are rendering as HTML `<pre><code>` pairs when the paragraph extends over more than one line.

### Steps to reproduce

_No response_

### Exception(s) (if any)

_No response_

6.7.0

### .NET Version

net8.0

### Anything else?

_No response_

EvgeniyZ:

Fixes #3009

Swagger.ui (or Redoc, I'm not sure) are displaying `\r\n   ` as `<pre>` tag.

PR changes making sure that <para> text is **trimmed** and all line breaks are ignored

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsTextHelper.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsTextHelper.cs
@@ -106,12 +106,16 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
 
         private static string HumanizeParaTags(this string text)
         {
-            return ParaTag().Replace(text, (match) => "<br>" + match.Groups["display"].Value);
+            return ParaTag().Replace(text, match =>
+            {
+                var paraText = "<br>" + match.Groups["display"].Value.TrimStart();
+                return LineBreaks().Replace(paraText, _ => string.Empty);
+            }).Replace("\r\n<br>", "<br>");
         }
 
         private static string HumanizeBrTags(this string text)
         {
-            return BrTag().Replace(text, m => Environment.NewLine);
+            return BrTag().Replace(text, _ => Environment.NewLine);
         }
 
         private static string DecodeXml(this string text)
@@ -125,6 +129,7 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
         private const string ParaTagPattern = @"<para>(?<display>.+?)</para>";
         private const string HrefPattern = @"<see href=\""(.*)\"">(.*)<\/see>";
         private const string BrPattern = @"(<br ?\/?>)"; // handles <br>, <br/>, <br />
+        private const string LineBreaksPattern = @"\r\n?|\n";
 
 #if NET7_0_OR_GREATER
         [GeneratedRegex(RefTagPattern)]
@@ -144,6 +149,9 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
 
         [GeneratedRegex(BrPattern)]
         private static partial Regex BrTag();
+
+        [GeneratedRegex(LineBreaksPattern)]
+        private static partial Regex LineBreaks();
 #else
         private static readonly Regex _refTag = new(RefTagPattern);
         private static readonly Regex _codeTag = new(CodeTagPattern);
@@ -151,6 +159,7 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
         private static readonly Regex _paraTag = new(ParaTagPattern, RegexOptions.Singleline);
         private static readonly Regex _hrefTag = new(HrefPattern);
         private static readonly Regex _brTag = new(BrPattern);
+        private static readonly Regex _lineBreaks = new(LineBreaksPattern);
 
         private static Regex RefTag() => _refTag;
         private static Regex CodeTag() => _codeTag;
@@ -158,6 +167,7 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
         private static Regex ParaTag() => _paraTag;
         private static Regex HrefTag() => _hrefTag;
         private static Regex BrTag() => _brTag;
+        private static Regex LineBreaks() => _lineBreaks;
 #endif
     }
 }
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsTextHelper.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsTextHelper.cs
@@ -108,9 +108,9 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
         {
             return ParaTag().Replace(text, match =>
             {
-                var paraText = "<br>" + match.Groups["display"].Value.TrimStart();
+                var paraText = "<br>" + match.Groups["display"].Value.Trim();
                 return LineBreaks().Replace(paraText, _ => string.Empty);
-            }).Replace("\r\n<br>", "<br>");
+            });
         }
 
         private static string HumanizeBrTags(this string text)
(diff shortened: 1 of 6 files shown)


test/Swashbuckle.AspNetCore.SwaggerGen.Test/XmlComments/XmlCommentsTextHelperTests.cs
```diff
@@ -129,6 +129,7 @@ Misplaced Tab Indentation
         [InlineData("<c>DoWork</c> is a method in <c>TestClass</c>.", "`DoWork` is a method in `TestClass`.")]
         [InlineData("<code>DoWork</code> is a method in <code>\nTestClass\n</code>.", "```DoWork``` is a method in ```\nTestClass\n```.")]
         [InlineData("<para>This is a paragraph</para>.", "\r\nThis is a paragraph.")]
+        [InlineData("<para>   This is a paragraph   </para>.", "\r\nThis is a paragraph.")]
```
EvgeniyZ, line 132: please, let me know if this OK to trim whitespaces before `<para>` and after `</para>`

test/WebSites/Basic/Controllers/CrudActionsController.cs
```diff
@@ -71,6 +71,14 @@ public void Update(int id, [FromBody, Required]Product product)
         /// <summary>
         /// Updates some properties of a specific product
         /// </summary>
+        /// <remarks>
```
EvgeniyZ, line 74: remarks block translated into

image

src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsTextHelper.cs
```diff
@@ -106,12 +106,16 @@ private static string HumanizeMultilineCodeTags(this string text)
...
+            {
+                var paraText = "<br>" + match.Groups["display"].Value.Trim();
+                return LineBreaks().Replace(paraText, _ => string.Empty);
+            });
         }
 
         private static string HumanizeBrTags(this string text)
         {
-            return BrTag().Replace(text, m => Environment.NewLine);
+            return BrTag().Replace(text, _ => Environment.NewLine);
```
EvgeniyZ, line 118: it seems complicated to make tests pass on macos and ubuntu because for unix `\n` is used for line breaks and here we replacing `<br>` tag as `Environment.NewLine`. But in `...v1.verified.txt` files you're using specific line breaks `\r\n`.

I've searched the codebase and didn't find any examples of `<para>` in /test/WebSites

src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsTextHelper.cs
```diff
@@ -125,6 +129,7 @@ private static string DecodeXml(this string text)
         private const string ParaTagPattern = @"<para>(?<display>.+?)</para>";
         private const string HrefPattern = @"<see href=\""(.*)\"">(.*)<\/see>";
         private const string BrPattern = @"(<br ?\/?>)"; // handles <br>, <br/>, <br />
+        private const string LineBreaksPattern = @"\r\n?|\n";
```
martincostello, line 132: Shame we can't use `string.ReplaceLineEndings()`.

This could be simplified to this? `\r?\n`. I'm not sure whether anyone still actively uses `\r` for a new line on modern macOS?

test/Swashbuckle.AspNetCore.SwaggerGen.Test/XmlComments/XmlCommentsTextHelperTests.cs
martincostello: I'm not sure - as long as things don't get squashed together.

For example we wouldn't want `<para> one </para><para> two </para>` becoming `onetwo`.

src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsTextHelper.cs
martincostello: Maybe there's a setting in [Verify](https://github.com/VerifyTests/Verify) that can be used to ignore line endings in specific tests?

test/Swashbuckle.AspNetCore.SwaggerGen.Test/XmlComments/XmlCommentsTextHelperTests.cs
EvgeniyZ: I've added test case, it will humanized to `"\r\none\r\ntwo"` - I think it's OK

src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsTextHelper.cs
EvgeniyZ: > I'm not sure whether anyone still actively uses \r for a new line on modern macOS?

I think you're right, simplified

src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsTextHelper.cs
EvgeniyZ: good idea, let me try to dig something

EvgeniyZ: I couldn't find anything useful in `Verify` link. I also tried more elegant solution with `.ReplaceLineEndings()` and regex expressions but no avail

test/Swashbuckle.AspNetCore.IntegrationTests/SwaggerVerifyIntegrationTest.cs
```diff
@@ -111,6 +114,25 @@ private static async Task<string> SwaggerResponse(HttpClient client, string swag
...
+                if (swagger[i] == '\\' && i > 0 && i < swagger.Length - 1 && swagger[i+1] == 'n' && swagger[i - 1] != 'r')
+                {
+                    output.Append("\\r");
+                }
+
+                output.Append(swagger[i]);
+            }
+
+            return output;
+        }
```
martincostello, line 134: Maybe a better solution to this would be to read `swagger` into a `StringReader`, then iterate through it line-by-line (so the framework does the parsing for us), and then emit each line read into `output`.

Something like:

```csharp
using var reader = new StringReader(swagger);

string line;

while ((line = reader.ReadLine()) is not null)
{
    output.Append(line);
    output.Append("\r\n");
}
```

EvgeniyZ: it doesn't help much because these `\n` or `\r\n` are part of description field in json and they are actually string values, not new lines characters, so they are not parsed by framework (or string.ReplaceLineEndings)

I managed to simplify method by using regex

martincostello (approved):

PR #3242, opened 2025-01-30
Remove MvcOptions from SchemaGenerator

### LINKED ISSUE #3203: NonNullableReferenceTypesAsRequired should not consider MvcOptions
### Describe the bug

The `NonNullableReferenceTypesAsRequired` option only works if you've also set the `MvcOptions.SuppressImplicitRequiredAttributeForNonNullableReferenceTypes` to false

This is surprising. If I opt into a feature it shouldn't be predicated on an MVC-specific option also being enabled. 

### Expected behavior

`NonNullableReferenceTypesAsRequired` should work no matter how the MvcOption is configured.

### Actual behavior

The feature does not work if `MvcOptions.SuppressImplicitRequiredAttributeForNonNullableReferenceTypes` is `true`

### Steps to reproduce

_No response_

### Exception(s) (if any)

_No response_

7.2.0

### .NET Version

Net9

### Anything else?

We are trying to upgrade to NET9. We've had a custom filter in place for years that does the exact same thing as the new `NonNullableReferenceTypesAsRequired` option and we were excited to finally be able to remove it.

Unfortunately we explicitly configure the `MvcOptions.SuppressImplicitRequiredAttributeForNonNullableReferenceTypes` to `true`. This is because the MVC option is a huge hammer. We'd like our API to be forgiving in certain things it accepts, but have the API specification indicate the desired state of affairs. Required-ness is also not limited to inputs, as read-only schemas can indicate properties are required, ie. that they will exist in the returned data. We also rely on the readonly properties when passing our api specification off to code generators, but again don't want to enforce it across the board at the MVC layer except where done explicitly.

In the PR that added this, there was a little discussion on including the MvcOptions, but it was mostly about breaking APIs. 

It's also surprising that if I were building a netstandard library and opt into the feature that it _would_ work since it ifdef's around, which feels very inconsistent. Also given that this is predicated on an MVC-specific option, does that mean the filter won't work properly with minimal apis?

Please consider either removing the dependency on MvcOptions or providing us a way to forcefully opt-in. We'll keep our custom filter in the meantime, bummer!

EvgeniyZ:

Fixes #3203 

remove any usages of MvcOptions in 'SchemaGenerator'. It should not be a breaking change since bool expression is 'true' by definition if 'SuppressImplicitRequiredAttributeForNonNullableReferenceTypes=false'

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenOptionsExtensions.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenOptionsExtensions.cs
@@ -4,6 +4,7 @@ using System.IO;
 using System.Reflection;
 using System.Xml.XPath;
 using Microsoft.AspNetCore.Authentication;
+using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.Mvc.ApiExplorer;
 using Microsoft.OpenApi.Models;
 using Swashbuckle.AspNetCore.SwaggerGen;
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenOptionsExtensions.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenOptionsExtensions.cs
@@ -4,7 +4,6 @@ using System.IO;
 using System.Reflection;
 using System.Xml.XPath;
 using Microsoft.AspNetCore.Authentication;
-using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.Mvc.ApiExplorer;
 using Microsoft.OpenApi.Models;
 using Swashbuckle.AspNetCore.SwaggerGen;
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/PublicAPI/PublicAPI.Shipped.txt
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/PublicAPI/PublicAPI.Shipped.txt
@@ -202,7 +202,7 @@ Swashbuckle.AspNetCore.SwaggerGen.SchemaFilterContext.Type.get -> System.Type
 Swashbuckle.AspNetCore.SwaggerGen.SchemaGenerator
 Swashbuckle.AspNetCore.SwaggerGen.SchemaGenerator.GenerateSchema(System.Type modelType, Swashbuckle.AspNetCore.SwaggerGen.SchemaRepository schemaRepository, System.Reflection.MemberInfo memberInfo = null, System.Reflection.ParameterInfo parameterInfo = null, Microsoft.AspNetCore.Mvc.ApiExplorer.ApiParameterRouteInfo routeInfo = null) -> Microsoft.OpenApi.Models.OpenApiSchema
 Swashbuckle.AspNetCore.SwaggerGen.SchemaGenerator.SchemaGenerator(Swashbuckle.AspNetCore.SwaggerGen.SchemaGeneratorOptions generatorOptions, Swashbuckle.AspNetCore.SwaggerGen.ISerializerDataContractResolver serializerDataContractResolver) -> void
-Swashbuckle.AspNetCore.SwaggerGen.SchemaGenerator.SchemaGenerator(Swashbuckle.AspNetCore.SwaggerGen.SchemaGeneratorOptions generatorOptions, Swashbuckle.AspNetCore.SwaggerGen.ISerializerDataContractResolver serializerDataContractResolver, Microsoft.Extensions.Options.IOptions<Microsoft.AspNetCore.Mvc.MvcOptions> _) -> void
+Swashbuckle.AspNetCore.SwaggerGen.SchemaGenerator.SchemaGenerator(Swashbuckle.AspNetCore.SwaggerGen.SchemaGeneratorOptions generatorOptions, Swashbuckle.AspNetCore.SwaggerGen.ISerializerDataContractResolver serializerDataContractResolver, Microsoft.Extensions.Options.IOptions<Microsoft.AspNetCore.Mvc.MvcOptions> mvcOptions) -> void
 Swashbuckle.AspNetCore.SwaggerGen.SchemaGeneratorOptions
 Swashbuckle.AspNetCore.SwaggerGen.SchemaGeneratorOptions.CustomTypeMappings.get -> System.Collections.Generic.IDictionary<System.Type, System.Func<Microsoft.OpenApi.Models.OpenApiSchema>>
 Swashbuckle.AspNetCore.SwaggerGen.SchemaGeneratorOptions.CustomTypeMappings.set -> void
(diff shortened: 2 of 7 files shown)


src/Swashbuckle.AspNetCore.SwaggerGen/PublicAPI/PublicAPI.Shipped.txt
```diff
@@ -202,7 +202,6 @@ Swashbuckle.AspNetCore.SwaggerGen.SchemaFilterContext.Type.get -> System.Type
 Swashbuckle.AspNetCore.SwaggerGen.SchemaGenerator
 Swashbuckle.AspNetCore.SwaggerGen.SchemaGenerator.GenerateSchema(System.Type modelType, Swashbuckle.AspNetCore.SwaggerGen.SchemaRepository schemaRepository, System.Reflection.MemberInfo memberInfo = null, System.Reflection.ParameterInfo parameterInfo = null, Microsoft.AspNetCore.Mvc.ApiExplorer.ApiParameterRouteInfo routeInfo = null) -> Microsoft.OpenApi.Models.OpenApiSchema
 Swashbuckle.AspNetCore.SwaggerGen.SchemaGenerator.SchemaGenerator(Swashbuckle.AspNetCore.SwaggerGen.SchemaGeneratorOptions generatorOptions, Swashbuckle.AspNetCore.SwaggerGen.ISerializerDataContractResolver serializerDataContractResolver) -> void
-Swashbuckle.AspNetCore.SwaggerGen.SchemaGenerator.SchemaGenerator(Swashbuckle.AspNetCore.SwaggerGen.SchemaGeneratorOptions generatorOptions, Swashbuckle.AspNetCore.SwaggerGen.ISerializerDataContractResolver serializerDataContractResolver, Microsoft.Extensions.Options.IOptions<Microsoft.AspNetCore.Mvc.MvcOptions> mvcOptions) -> void
```
EvgeniyZ, line 205: it seems that I'm breaking backward compatibility by removing this `ctor` since `SchemaGenerator` is a public class, but imho it will be misleading to accept 'MvcOptions' ctor parameter and not use it at all...

src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenOptionsExtensions.cs
```diff
@@ -4,6 +4,7 @@
 using System.Reflection;
 using System.Xml.XPath;
 using Microsoft.AspNetCore.Authentication;
+using Microsoft.AspNetCore.Mvc;
```
martincostello, line 7: What is this addition for?

src/Swashbuckle.AspNetCore.SwaggerGen/PublicAPI/PublicAPI.Shipped.txt
martincostello: Yep it is - the better approach would be to mark it as obsolete. We could then remove it in some future major version.

test/Swashbuckle.AspNetCore.SwaggerGen.Test/SchemaGenerator/JsonSerializerSchemaGeneratorTests.cs
```diff
@@ -984,29 +984,6 @@ public void GenerateSchema_SupportsOption_NonNullableReferenceTypesAsRequired_Re
             Assert.Equal(required, propertyIsRequired);
         }
 
-        [Theory]
-        [InlineData(typeof(TypeWithNullableContextAnnotated), nameof(TypeWithNullableContextAnnotated.SubTypeWithOneNonNullableContent), nameof(TypeWithNullableContextAnnotated.NonNullableString), false)]
```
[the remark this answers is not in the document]
martincostello: I would keep the tests to verify that the obsolete constructor behaves as expected (i.e. ignoring the `MvcOptions` value). Marking the test as obsolete would remove any warnings about obsolete usage, and then we'd delete those tests when the constructor is removed at a future date.

src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/SchemaGenerator.cs
```diff
@@ -12,29 +12,20 @@
...
 using Microsoft.OpenApi.Models;
-using Microsoft.Extensions.Options;
 
 namespace Swashbuckle.AspNetCore.SwaggerGen
 {
     public class SchemaGenerator : ISchemaGenerator
     {
         private readonly SchemaGeneratorOptions _generatorOptions;
         private readonly ISerializerDataContractResolver _serializerDataContractResolver;
-        private readonly IOptions<MvcOptions> _mvcOptions;
```
martincostello, line 23: We can remove the field and usage of the constructor argument, but we need to keep the signature the same but mark it as obsolete instead.

src/Swashbuckle.AspNetCore.SwaggerGen/PublicAPI/PublicAPI.Shipped.txt
EvgeniyZ: got it, applied suggestion

src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenOptionsExtensions.cs
EvgeniyZ: no need, removed

src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/SchemaGenerator.cs
EvgeniyZ: applied suggestion

src/Swashbuckle.AspNetCore.SwaggerGen/PublicAPI/PublicAPI.Shipped.txt
```diff
@@ -202,7 +202,7 @@ Swashbuckle.AspNetCore.SwaggerGen.SchemaFilterContext.Type.get -> System.Type
...
+Swashbuckle.AspNetCore.SwaggerGen.SchemaGenerator.SchemaGenerator(Swashbuckle.AspNetCore.SwaggerGen.SchemaGeneratorOptions generatorOptions, Swashbuckle.AspNetCore.SwaggerGen.ISerializerDataContractResolver serializerDataContractResolver, Microsoft.Extensions.Options.IOptions<Microsoft.AspNetCore.Mvc.MvcOptions> _) -> void
```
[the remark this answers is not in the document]
EvgeniyZ: right, renamed back

src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/SchemaGenerator.cs
```diff
@@ -10,31 +10,32 @@
...
-            : this(generatorOptions, serializerDataContractResolver, null)
+        public SchemaGenerator(
+            SchemaGeneratorOptions generatorOptions,
+            ISerializerDataContractResolver serializerDataContractResolver)
         {
+            _generatorOptions = generatorOptions;
+            _serializerDataContractResolver = serializerDataContractResolver;
         }
 
+        [Obsolete($"{nameof(IOptions<MvcOptions>)} is not used. This constructor will be removed in future versions")]
```
martincostello, line 32:
```suggestion
        [Obsolete($"{nameof(IOptions<MvcOptions>)} is no longer used. This constructor will be removed in a future major release.")]
```

martincostello (approved):

PR #3245, opened 2025-02-03
Fix JsonSerializerDataContractResolver so that it handles square arrays correctly

### LINKED ISSUE #3244: [Bug]: JsonSerializerDataContractResolver cannot handle square arrays, but NewtonsoftDataContractResolver can
### Describe the bug

If I define a schema that has a square array:

```csharp
public class SwaggerFun
{
    public string[,] StringSquareArray { get; set; }
}
```

when using `AddSwaggerGenNewtonsoftSupport(...)` this is correctly described as:

```json
{
    ...

    "stringSquareArray": {
        "items": {
            "type": "string"
        },
        "nullable": true,
        "type": "array"
    },

    ...
}
```

however, when using the default System.Text.Json-based data contract resolver, this is incorrectly described as:

```json
{
    ...

    "stringSquareArray": {
        "items": {},
        "nullable": true,
        "type": "array"
    },

    ...
}
```

that is - the type of the item in the array is not correctly conveyed.

### Expected behavior

I'd expect the `System.Text.Json` implementation to work the same way as the `Newtonsoft.Json` implementation.

### Actual behavior

The `System.Text.Json` implemenation does not convey the type of the item in the array.

### Steps to reproduce

I'll be supplying a fix for the issue.

### Exception(s) (if any)

_No response_

7.2.0

### .NET Version

9.0.102

### Anything else?

_No response_

ozziepeeps:
# Pull Request

Fix JsonSerializerDataContractResolver so that it handles square arrays correctly
Fixes #3244

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/JsonSerializerDataContractResolver.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/JsonSerializerDataContractResolver.cs
@@ -142,6 +142,12 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
             }
 #endif
 
+            if (type.IsArray)
+            {
+                itemType = type.GetElementType();
+                return true;
+            }
+
             if (typeof(IEnumerable).IsAssignableFrom(type))
             {
                 itemType = typeof(object);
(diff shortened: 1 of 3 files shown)


ozziepeeps:
@martincostello hopefully this is a quick review for you! TIA!

martincostello (approved):
Thanks!

Issue #2349, opened 2022-02-21
Plans on official support for OpenApi 3.1.0


Danielku15:
Unfortunately it is a bit hard to search for "3.1" or "3.1.0" in GitHub so I hope I'm not raising a duplicate here. 

I was wondering what the plans on officially adding support for OpenApi 3.1.0 are? 
https://github.com/OAI/OpenAPI-Specification/releases/tag/3.1.0
 
I'm not aware of all the exact changes but as of today Swashbuckle always generates version "3.0.1" schema and there seems no way to change this (value is hard-coded in SerializeAsV3). In our case it leads to some validation errors in the schema as we are using request bodies in DELETE operations, this was not allowed in "3.0.1" but the restriction was lifted in "3.1.0" because the HTTP RFC allows it.

Eneuman:
Since swagger is dependant on MS openAPI.Net project, I think we need to wait for it to first support the 3.1.0 spec, but I can be wrong here.

darrelmiller:
It is on our roadmap https://github.com/microsoft/OpenAPI.NET/issues/795

szymczakk:
Hi! Any news on this? As https://github.com/microsoft/OpenAPI.NET/issues/795 is closed.

jnv:
@szymczakk Wait for preview release? https://github.com/microsoft/OpenAPI.NET/issues/795#issuecomment-1994528711

szymczakk:
> @szymczakk Wait for preview release? [microsoft/OpenAPI.NET#795 (comment)](https://github.com/microsoft/OpenAPI.NET/issues/795#issuecomment-1994528711)

Sorry, just saw the status that is closed, haven’t read the comments.

thanks :)

vindberg:
When will this be supported? Any workarounds until this is solved?

martincostello:
We can only look into supporting it when Microsoft.OpenApi supports it.

John-Paul-R:
For anyone as confused as I:

Despite https://github.com/microsoft/OpenAPI.NET/issues/795 ("Add OpenAPI 3.1 Support") being closed, the release it is slated to go out in, [OpenAPI.NET:2.0](https://github.com/microsoft/OpenAPI.NET/milestone/8) is currently still in progress (delayed from its original slated release date).

elrolito:
[...]

Issue #2849, opened 2024-05-01
Future of OpenApiOperation in endpoint metadata


captainsafia:
Hi everyone!

Filing this issue because I'd like to hear some perspectives on something I'm reasoning through in the `Microsoft.AspNetCore.OpenApi` implementation.

In .NET 7, we added support for the `WithOpenApi` endpoint extension method for minimal APIs. When this API was invoked, we would generate an `OpenApiOperation` associated with the endpoint using [the OpenApiGenerator](https://github.com/dotnet/aspnetcore/blob/main/src/OpenApi/src/Services/OpenApiGenerator.cs). This `OpenApiOperation` would be inserted into endpoint metadata. From there, we made a change so that Swashbuckle would consume the `OpenApiOperation` that was present in metadata and insert it into the document it was generating (see [here](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/cbc1b7aa34c8662c33dd854efef49b66706965ad/src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs#L223)).

Now that built-in OpenAPI document generation is supported in `Microsoft.AspNetCore.OpenApi`, I am trying to figure out what the best path forward `WithOpenApi` and `OpenApiOperation` in metadata is. Here's some of the options on the table:

1. Do nothing: Keep `WithOpenApi` and `OpenApiGenerator` as detached from the new document generation in Microsoft.AspNetCore.OpenApi.
2. Deprecate `WithOpenApi` and (eventually) remove logic for integrating OpenApiOperation from metadata into Swashbuckle's document.
3. Replat `WithOpenApi` as shorthand for producing an operation transformer but _remove_ functionality for inserting the OpenAPI operation into metadata for Swashbuckle to consume.
4. Replat `WithOpenApi` as shorthand for producing an operation transformer and project the OpenApiOperation from the built-in document into metadata for Swashbuckle to consume.

Of the options above, #4 is probably the most expensive to implement but retains the most amount of behavior for users. #1 is cheap but doesn't leave users in a good place WRT whether they should use `WithOpenApi` at all.

Thoughts?

martincostello:
My thoughts on the above:

1. Makes adopting the new things opt-in so keeps the distinction and maintains backwards compatibility. Doesn't break anything, but doesn't guide anyone to improvements.
[...]

PR #3252, opened 2025-02-05
Upgrade to Microsoft.OpenApi v2

martincostello:
Upgrade to Microsoft.OpenApi v2, which would unlock support for OpenAPI 3.1.

Contributes to #2349.

<hr/>

This currently targets preview 5 and compiles, but **lots** of tests are broken.

The migration was guided by [the release notes for the first preview](https://github.com/microsoft/OpenAPI.NET/releases/tag/2.0.0-preview1).

Some might be bugs in the upgrade, others might be bugs in Microsoft.OpenApi v2 itself. I'll need to investigate further, then I can open any upstream issues as neccessary.

The most worrying failure so far is this one:

```text
System.MissingMethodException: Method not found: 'Void Microsoft.OpenApi.Models.OpenApiOperation.set_Tags(System.Collections.Generic.IList`1<Microsoft.OpenApi.Models.OpenApiTag>)'.
   at Microsoft.AspNetCore.OpenApi.OpenApiGenerator.GetOperation(String httpMethod, MethodInfo methodInfo, EndpointMetadataCollection metadata, RoutePattern pattern)
   at Microsoft.AspNetCore.OpenApi.OpenApiGenerator.GetOpenApiOperation(MethodInfo methodInfo, EndpointMetadataCollection metadata, RoutePattern pattern)
   at Microsoft.AspNetCore.Builder.OpenApiEndpointConventionBuilderExtensions.AddAndConfigureOperationForEndpoint(EndpointBuilder endpointBuilder, Func`2 configure)
   at Microsoft.AspNetCore.Builder.OpenApiEndpointConventionBuilderExtensions.<>c__0`1.<WithOpenApi>b__0_0(EndpointBuilder builder)
   at Microsoft.AspNetCore.Routing.RouteEndpointDataSource.CreateRouteEndpointBuilder(RouteEntry entry, RoutePattern groupPrefix, IReadOnlyList`1 groupConventions, IReadOnlyList`1 groupFinallyConventions)
   at Microsoft.AspNetCore.Routing.RouteEndpointDataSource.GetGroupedEndpoints(RouteGroupContext context)
   at Microsoft.AspNetCore.Routing.RouteGroupBuilder.GroupEndpointDataSource.GetGroupedEndpointsWithNullablePrefix(RoutePattern prefix, IReadOnlyList`1 conventions, IReadOnlyList`1 finallyConventions, IServiceProvider applicationServices)
   at Microsoft.AspNetCore.Routing.RouteGroupBuilder.GroupEndpointDataSource.get_Endpoints()
   at Microsoft.AspNetCore.Routing.CompositeEndpointDataSource.CreateEndpointsUnsynchronized()
   at Microsoft.AspNetCore.Routing.CompositeEndpointDataSource.EnsureEndpointsInitialized()
   at Microsoft.AspNetCore.Routing.CompositeEndpointDataSource.get_Endpoints()
   at Microsoft.AspNetCore.Routing.DataSourceDependentCache`1.Initialize()
   at System.Threading.LazyInitializer.EnsureInitializedCore[T](T& target, Boolean& initialized, Object& syncLock, Func`1 valueFactory)
   at Microsoft.AspNetCore.Routing.Matching.DataSourceDependentMatcher..ctor(EndpointDataSource dataSource, Lifetime lifetime, Func`1 matcherBuilderFactory)
   at Microsoft.AspNetCore.Routing.Matching.DfaMatcherFactory.CreateMatcher(EndpointDataSource dataSource)
   at Microsoft.AspNetCore.Routing.EndpointRoutingMiddleware.InitializeCoreAsync()
--- End of stack trace from previous location ---
   at Microsoft.AspNetCore.Routing.EndpointRoutingMiddleware.<Invoke>g__AwaitMatcher|10_0(EndpointRoutingMiddleware middleware, HttpContext httpContext, Task`1 matcherTask)
   at Microsoft.AspNetCore.Diagnostics.DeveloperExceptionPageMiddlewareImpl.Invoke(HttpContext context)
```

This suggests that the changes are so breaking that it is basically impossible to use Microsoft.OpenApi v2 with ASP.NET Core versions earlier than 10. This creates a difficult situation for us, as we would potentially need to conditionally compile across TFMs and only be able to support OpenAPI 3.1 with `net10.0` and need to stay on 1.x for lower TFMs while we still support them. This could get quite messy and complex quite quickly.

Maybe it's not actually that bad and it's an artifact of how our integration test suite is set up, but it's not an encouraging start. We can get further insight on that once ASP.NET Core 10 Preview 1 is available, as then this PR can be updated to additionally target `net10.0` as that has support for OpenAPI 3.1.

/cc @captainsafia For thoughts on the above.

~~Also applies some refactorings in files that needed to be touched, but those will be extracted out and merged ahead of time where possible in a future separate PR.~~ #3254

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -21,14 +21,14 @@
     <PackageVersion Include="Microsoft.AspNetCore.TestHost" Version="6.0.32" />
     <PackageVersion Include="Microsoft.CodeAnalysis.PublicApiAnalyzers" Version="3.3.4" />
     <PackageVersion Include="Microsoft.Extensions.FileProviders.Embedded" Version="2.1.0" />
-    <PackageVersion Include="Microsoft.OpenApi" Version="1.6.22" />
-    <PackageVersion Include="Microsoft.OpenApi.Readers" Version="1.6.22" />
+    <PackageVersion Include="Microsoft.OpenApi" Version="2.0.0-preview5" />
+    <PackageVersion Include="Microsoft.OpenApi.Readers" Version="2.0.0-preview5" />
     <PackageVersion Include="Microsoft.NET.Test.Sdk" Version="17.12.0" />
     <PackageVersion Include="Newtonsoft.Json" Version="13.0.3" />
     <PackageVersion Include="NSubstitute" Version="5.3.0" />
     <PackageVersion Include="NSwag.MSBuild" Version="14.2.0" />
     <PackageVersion Include="ReportGenerator" Version="5.4.1" />
-    <PackageVersion Include="System.Text.Json" Version="4.6.0" />
+    <PackageVersion Include="System.Text.Json" Version="8.0.5" />
     <PackageVersion Include="Verify.Xunit" Version="28.3.2" />
     <PackageVersion Include="xunit" Version="2.9.2" />
     <PackageVersion Include="xunit.core" Version="2.9.2" />
--- a/perf/Swashbuckle.AspNetCore.Benchmarks/XmlCommentsBenchmark.cs
+++ b/perf/Swashbuckle.AspNetCore.Benchmarks/XmlCommentsBenchmark.cs
@@ -101,7 +101,7 @@ public class XmlCommentsBenchmark
         {
             Schema = new()
             {
-                Type = "string",
+                Type = JsonSchemaType.String,
                 Description = "schema-level description",
             },
         };
@@ -120,7 +120,7 @@ public class XmlCommentsBenchmark
                 {
                     Schema = new()
                     {
-                        Type = "string",
+                        Type = JsonSchemaType.String,
                     },
                 },
             },
--- a/src/Swashbuckle.AspNetCore.Annotations/AnnotationsOperationFilter.cs
+++ b/src/Swashbuckle.AspNetCore.Annotations/AnnotationsOperationFilter.cs
@@ -63,12 +63,13 @@ namespace Swashbuckle.AspNetCore.Annotations
             if (swaggerOperationAttribute.OperationId != null)
                 operation.OperationId = swaggerOperationAttribute.OperationId;
 
-            if (swaggerOperationAttribute.Tags != null)
-            {
-                operation.Tags = swaggerOperationAttribute.Tags
-                    .Select(tagName => new OpenApiTag { Name = tagName })
-                    .ToList();
-            }
+            // TODO Fix this
+            ////if (swaggerOperationAttribute.Tags != null)
+            ////{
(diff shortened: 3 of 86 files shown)


captainsafia:
> This suggests that the changes are so breaking that it is basically impossible to use Microsoft.OpenApi v2 with ASP.NET Core versions earlier than 10.

In general, yes. We can't introduce breaking changes in .NET 9/.NET 8 at this point so there's no way for us to retroactively upgrade the OpenAPI.NET version in those versions.

I believe the crux of the issue here is with the `OpenApiOperation` metadata that gets included in the document as part of the `WithOpenApi` call. In https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/2849, we made the decision to keep this behavior for back-compat but perhaps a new major version of the Swashbuckle.AspNetCore package is a safe place to introduce that break?

We can validate this by seeing if a project that contains no references to `Microsoft.AspNetCore.OpenApi`, a reference to the new major version of `Swashbuckle.AspNetCore`, and targets `net10` is also suspectible to this issue.

martincostello:
> We can't introduce breaking changes in .NET 9/.NET 8 at this point so there's no way for us to retroactively upgrade the OpenAPI.NET version in those versions.

I was thinking less _"add support for it"_ and more _"not hard dependent on it"_, but I guess the difference is moot with regards to .NET's servicing branches if any changes need to be made to enable that.

The ideal scenario for Swashbuckle would be that using it in a .NET 8/9 app is _possible_, provided that that any APIs dependent on v1 can be opted-out of. That is to say, it would still be a breaking change (you need to change your filters, customisations etc.), but you **could** still move up without using .NET 10 if you wanted to use the new OpenAPI 3.1 features.

I guess that's something I can experiment with more in our tests by removing the `Microsoft.AspNetCore.OpenApi` dependency here and see what happens.

The most radical solution could be "new versions only support .NET 10" on the basis that 8 and 9 go out of support in 2026, and we just draw a hard line and also drop .NET Framework support going forwards. That would certainly make the codebase easier to support going forwards, it might just be unpopular with users that .NET 8 and 9 apps won't get Swashbuckle bug fixes for 6-12 months (unless we do branches and backporting for earlier versions, which of course has its own costs for an OSS project).

We already have at least one major version in the pipeline to drop .NET 6 support (#3183) with no concrete date yet, so if necessary we could always lay the groundwork for Microsoft.OpenApi v2 by dropping some things incrementally over the year as we move towards .NET 10 shipping so that there's less of a difference and people who care about using latest/using OpenAPI 3.1 have time to prepare.

martincostello:
Closing in favour of #3283.

PR #3263, opened 2025-02-16
Add support for listing available OpenAPI documents

### LINKED ISSUE #472: Expose a way to get the list of documents on ISwaggerProvider
Currently, there is no way to retrieve the list of documents which means that you need to know it's name before hand. 

It would be nice to add an Api to get the list of available documents `string[] ISwaggerProvider.GetSwaggerDocumentNames()`

Are you willing to accept a PR for that?

rassilon:
# Pull Request

* ISwaggerProvider/IAsyncSwaggerProvider: Add GetSwaggerDocumentNames method and implementation (and test coverage). This fixes #472.
* SwaggerUi: Add options to control the behavior of a new /swagger/documentUrls route that just returns the JSON version of ConfigOption.Urls. You can see this working in the MultipleVersions website.

ISwaggerProvider.cs: Add GetSwaggerDocumentNames method.
IAsyncSwaggerProvider.cs: ditto
Inclucled new public methods in PublicAP.Unshipped.txt
SwaggerGenerator.cs: Implement GetSwaggerDocumentNames.
SwaggerUIJsonSerializationContext: Added List<UrlDescriptor> due to various issues with just using IEnumerable<UrlDescriptor>.
SwaggerUIMiddleware.cs: Implement /swagger/documentUrls route.
SwaggerUIOptions.cs: Add a new bool property to turn on the documentUrls route and the path to the documentUrls route.
SwaggerUIOptionsExtensions.cs: Add an extension method to set the new bool property on SwaggerUIOptions.
SwaggerGeneratorTests.cs: Add GetSwaggerDocumentNames testing coverage to a pre-existing test case. 
MultipleVersions/Startup.cs: Enable the documentUrls route.

I need to automate broad security testing (both positive and negative) across all REST APIs our multiple microservice product utilizes. However, there wasn't a way to ask (in an automated fashion) what all of the swagger document URLs are...

Therefore, this PR.  I also decided to fix #472 while I was at it. 

I'm happy to address any changes you'd like to have me make, so I can get this into a public release so that my product team can upgrade to this new version.

Additionally, all tests passed successfully and I made sure I could successfully retrieve the expected JSON from the MultipleVersions website.

Thanks!


martincostello:
Thanks for your PR.

I've only skim read it for now, but I see that it adds members to an interface. This would be a breaking change, so even if we wanted to take that, it would require a major version bump, so isn't something that's going to happen quickly (certainly not as quickly as you seem to want it).

I would suggest looking into adding a _new_ interface, which can be implemented by the same types as the interface you modified, then we could release that in a minor version as it would just be incremental new functionality.

rassilon:
> Thanks for your PR.
> 
> I've only skim read it for now, but I see that it adds members to an interface. This would be a breaking change, so even if we wanted to take that, it would require a major version bump, so isn't something that's going to happen quickly (certainly not as quickly as you seem to want it).
> 
> I would suggest looking into adding a _new_ interface, which can be implemented by the same types as the interface you modified, then we could release that in a minor version as it would just be incremental new functionality.

I'm happy do to that. Any preferences on the naming of the extra interface? It could be ISwaggerProvider2/IAsyncSwaggerProvider2 in a homage to all of the fun COM interface naming practices of old, or it could be something simple like ISwaggerDocumentInformation or something like that with the method name: "GetDocumentNames" or something?

Thanks for your quick response!

rassilon:
@martincostello , I've updated the PR by moving the method into a new ISwaggerDocumentInformation interface instead.
If you'd like a different name, please don't hesitate to let me know.

Thanks!

src/Swashbuckle.AspNetCore.Swagger/ISwaggerDocumentInformation.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.OpenApi.Models;
+
+namespace Swashbuckle.AspNetCore.Swagger
+{
+    public interface ISwaggerDocumentInformation
```
martincostello, line 10: What about `ISwaggerDocumentMetadataProvider` or something like that?

src/Swashbuckle.AspNetCore.Swagger/ISwaggerDocumentInformation.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.OpenApi.Models;
```
martincostello, line 6: Most (all?) of these aren't needed.

src/Swashbuckle.AspNetCore.Swagger/ISwaggerDocumentInformation.cs
```diff
@@ -0,0 +1,14 @@
...
+
+namespace Swashbuckle.AspNetCore.Swagger
+{
+    public interface ISwaggerDocumentInformation
+    {
+        string[] GetDocumentNames();
```
martincostello, line 12: Let's return `IList<string>` so we're abstracting the specific type.

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
```diff
@@ -112,6 +112,11 @@ public OpenApiDocument GetSwagger(string documentName, string host = null, strin
             }
         }
 
+        public string[] GetDocumentNames()
+        {
+            return _options.SwaggerDocs.Keys.ToArray();
+        }
```
martincostello, line 118: This can be simplified to an expression:

```suggestion
        public string[] GetDocumentNames() => _options.SwaggerDocs.Keys.ToArray();
```

src/Swashbuckle.AspNetCore.SwaggerUI/SwaggerUIJsonSerializerContext.cs
```diff
@@ -1,12 +1,14 @@
...
+using System.Collections.Generic;
 using System.Text.Json;
 using System.Text.Json.Nodes;
 using System.Text.Json.Serialization;
 
 namespace Swashbuckle.AspNetCore.SwaggerUI;
 
 [JsonSerializable(typeof(ConfigObject))]
+[JsonSerializable(typeof(List<UrlDescriptor>))]
 [JsonSerializable(typeof(InterceptorFunctions))]
```
martincostello, line 12: Nit: sorting.

src/Swashbuckle.AspNetCore.SwaggerUI/SwaggerUIOptionsExtensions.cs
```diff
@@ -333,5 +333,10 @@ public static void UseResponseInterceptor(this SwaggerUIOptions options, string
         {
             options.Interceptors.ResponseInterceptorFunction = value;
         }
+
+        public static void EnableSwaggerDocumentUrlsRoute(this SwaggerUIOptions options)
```
martincostello, line 337: I wonder if we should call it an "endpoint" rather than a route?

src/Swashbuckle.AspNetCore.SwaggerUI/SwaggerUIOptions.cs
```diff
@@ -63,6 +63,16 @@ public class SwaggerUIOptions
         /// Gets or sets the path or URL to the Swagger UI CSS file.
         /// </summary>
         public string StylesPath { get; set; } = "./swagger-ui.css";
+
+        /// <summary>
+        /// Expose the ConfigObject.Urls object via a new &lt;root&gt;/documentUrls
```
martincostello, line 68: I would change this to point to the property rather than having the default in the comment.

test/Swashbuckle.AspNetCore.SwaggerGen.Test/SwaggerGenerator/SwaggerGeneratorTests.cs
```diff
@@ -50,6 +50,10 @@ public void GetSwagger_GeneratesSwaggerDocument_ForApiDescriptionsWithMatchingGr
                 }
             );
 
+            var provider = subject as ISwaggerDocumentInformation;
+            var documentNames = provider.GetDocumentNames();
+            Assert.Equal(new[] { "v1", "v2" }, documentNames);
```
martincostello, line 55: I think this can be changed to something like this:

```csharp
var provider = Assert.IsOfType<ISwaggerDocumentInformation>(subject);
var documentNames = provider.GetDocumentNames();
Assert.Equal([v1", "v2"], documentNames);
```

src/Swashbuckle.AspNetCore.Swagger/ISwaggerDocumentInformation.cs
rassilon: Fixed

src/Swashbuckle.AspNetCore.Swagger/ISwaggerDocumentInformation.cs
rassilon: Removed unnecessary ones.

src/Swashbuckle.AspNetCore.Swagger/ISwaggerDocumentInformation.cs
rassilon: Fixed

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
rassilon: Fixed

src/Swashbuckle.AspNetCore.SwaggerUI/SwaggerUIJsonSerializerContext.cs
rassilon: Put after InterceptorFunctions, let me know if you'd rather sort by the interior type (UrlDescriptor) and I can move it further down.

src/Swashbuckle.AspNetCore.SwaggerUI/SwaggerUIOptionsExtensions.cs
rassilon: Fixed

src/Swashbuckle.AspNetCore.SwaggerUI/SwaggerUIOptions.cs
rassilon: I've updated the documentation comments to use <see> and <c> as appropriate. Let me know what you think.

test/Swashbuckle.AspNetCore.SwaggerGen.Test/SwaggerGenerator/SwaggerGeneratorTests.cs
rassilon: I'm sure you meant Assert.IsType<...>(subject)
but xunit analyzer complained that we should use IsAssignableFrom<T> so, I updated to that instead.

rassilon:
For giggles I also locally added Cli support for a new 'list' command. I tried to see if that would work with the TestFirst website, but since that website doesn't reference Swagger/SwaggerGen that idea didn't pan out since it uses pre-generated json or imported json swagger documents. 

I also thought about adding a test for the documentUrls route to TesFirst/TestFirst.IntegrationTests but it looks like there's not any easy helpers in this area atm. I can add that if you'd like as well. Just use HttpClient directly maybe?

src/Swashbuckle.AspNetCore.Cli/Program.cs
```diff
@@ -156,9 +126,103 @@ public static int Main(string[] args)
...
+            // > dotnet swagger _list ... (* should only be invoked via "dotnet exec")
+            runner.SubCommand("_list", "", c =>
+            {
+                c.Argument("startupassembly", "");
+                c.Option("--output", "");
+                c.OnRun((namedArgs) =>
+                {
+                    SetupAndRetrieveSwaggerProviderAndOptions(namedArgs, out var swaggerProvider, out var swaggerOptions);
+                    var docMetaProvider = swaggerProvider as ISwaggerDocumentMetadataProvider;
+                    var docNames = docMetaProvider.GetDocumentNames();
```
martincostello, line 154: If `docMetaProvider` is null for some reason, we should log something to the console and bail out gracefully.

src/Swashbuckle.AspNetCore.Cli/Program.cs
```diff
@@ -156,9 +126,103 @@ public static int Main(string[] args)
...
+                    bool outputViaConsole = outputPath == null;
+                    if (!string.IsNullOrEmpty(outputPath))
+                    {
+                        string directoryPath = Path.GetDirectoryName(outputPath);
+                        if (!string.IsNullOrEmpty(directoryPath) && !Directory.Exists(directoryPath))
+                        {
+                            Directory.CreateDirectory(directoryPath);
+                        }
+                    }
+
```
martincostello, line 167:
```suggestion
```

src/Swashbuckle.AspNetCore.Cli/Program.cs
```diff
@@ -156,9 +126,103 @@ public static int Main(string[] args)
...
+
+                    using Stream stream = outputViaConsole ? Console.OpenStandardOutput() : File.Create(outputPath);
+                    using StreamWriter writer = new StreamWriter(stream, outputViaConsole ? Console.OutputEncoding : Encoding.UTF8);
```
martincostello, line 170:
```suggestion
                    using StreamWriter writer = new(stream, outputViaConsole ? Console.OutputEncoding : Encoding.UTF8);
```

src/Swashbuckle.AspNetCore.Cli/Program.cs
```diff
@@ -156,9 +126,103 @@ public static int Main(string[] args)
...
+
+                    writer.WriteLine($"Swagger Document Names (Surrounded by single quotes):");
```
martincostello, line 172: Not sure we need the suffix here.

src/Swashbuckle.AspNetCore.Cli/Program.cs
```diff
@@ -156,9 +126,103 @@ public static int Main(string[] args)
...
+                    foreach (var name in docNames)
+                    {
+                        writer.WriteLine($"'{name}'");
```
martincostello, line 175: Why quotes, and if quotes, why not double-quotes?

src/Swashbuckle.AspNetCore.Cli/Program.cs
```diff
@@ -156,9 +126,103 @@ public static int Main(string[] args)
...
+                    string subProcessCommandLine = PrepareCommandLine(args, namedArgs);
+
+                    var subProcess = Process.Start("dotnet", subProcessCommandLine);
+
+                    subProcess.WaitForExit();
+                    return subProcess.ExitCode;
+                });
+            });
+
+            // > dotnet swagger _list ... (* should only be invoked via "dotnet exec")
```
martincostello, line 145: I don't fully understand why there's two list commands here.

src/Swashbuckle.AspNetCore.SwaggerUI/SwaggerUIOptions.cs
```diff
@@ -63,6 +63,18 @@ public class SwaggerUIOptions
         /// Gets or sets the path or URL to the Swagger UI CSS file.
         /// </summary>
         public string StylesPath { get; set; } = "./swagger-ui.css";
+
+        /// <summary>
+        /// Expose the <c><see cref="SwaggerUIOptions.ConfigObject">ConfigObject</see>.Urls</c> object via a new
+        /// "&lt;<see cref="SwaggerUIOptions.RoutePrefix"/>&gt;/&lt;<see cref="SwaggerUIOptions.SwaggerDocumentUrlsPath"/>&gt;
+        /// route so that external code can auto-discover all swagger documents.
```
martincostello, line 70:
```suggestion
        /// Gets or sets whether to expose the <c><see cref="SwaggerUIOptions.ConfigObject">ConfigObject</see>.Urls</c> object via an
        /// HTTP endpoint with the URL specified by <see cref="SwaggerDocumentUrlsPath"/>
        /// so that external code can auto-discover all Swagger documents.
```

src/Swashbuckle.AspNetCore.SwaggerUI/SwaggerUIOptions.cs
```diff
@@ -63,6 +63,18 @@ public class SwaggerUIOptions
...
+        /// </summary>
+        public bool ExposeSwaggerDocumentUrlsRoute { get; set; } = false;
+
+        /// <summary>
+        /// Gets or sets the relative URL path to the route that exposes the <c>SwaggerUIOptions.ConfigObjet.Urls</c>>.
```
martincostello, line 75:
```suggestion
        /// Gets or sets the relative URL path to the route that exposes the values of the configured <see cref="ConfigObject.Urls"/> values.
```

src/Swashbuckle.AspNetCore.Cli/Program.cs
rassilon: Fixed

src/Swashbuckle.AspNetCore.Cli/Program.cs
rassilon: Removed

src/Swashbuckle.AspNetCore.Cli/Program.cs
rassilon: Well, since the document names can be arbitrary values, it's a good idea to surround them with quotes. 
And, I'll tend to default to picking the opposite quote character for output wrapping than the string literal I'm using by default from too much JavaScript code writing.. :) That said... I've switched this to double quotes.

src/Swashbuckle.AspNetCore.SwaggerUI/SwaggerUIOptions.cs
rassilon: Fixed

src/Swashbuckle.AspNetCore.Cli/Program.cs
```diff
@@ -165,14 +172,22 @@ public static int Main(string[] args)
...
-                    using StreamWriter writer = new StreamWriter(stream, outputViaConsole ? Console.OutputEncoding : Encoding.UTF8);
+                    using StreamWriter writer = new(stream, outputViaConsole ? Console.OutputEncoding : Encoding.UTF8);
 
-                    writer.WriteLine($"Swagger Document Names (Surrounded by single quotes):");
-                    foreach (var name in docNames)
+                    if (docMetaProvider == null)
                     {
-                        writer.WriteLine($"'{name}'");
+                        writer.WriteLine($"ERROR: ISwaggerProvider instance DOES NOT support ISwaggerDocumentMetadaatProvider, so this coommand "
+                            + "DOES NOT support listing out the document names. Try updating the version of Swashbuckle used by the project you are examining to a newer version.");
```
martincostello, line 181: We already checked whether it wasn't null earlier, so the error can be emitted there and stop before doing any more work.

```csharp
if (swaggerProvider is not ISwaggerDocumentMetadataProvider docMetaProvider)
{
    // error message
    return -1;
}

var docNames = docMetaProvider.GetDocumentNames();

// etc.
```

martincostello, line 181: A few comments on this message:

1. Don't use capitalisation to emphasise words - it just reads like it's shouting at the user.
2. There's a number of typos.
3. Use `nameof()` to format in type names, then they'll definitely be spelled correctly.
4. I'm not sure that such a scenario would manifest that way. The tool is loading the app's types to inspect them with reflection, so if it's our provider it must implement the interface by definition. If it's a custom user-provider implementation, then upgrading won't help. I would instead just state the fact that the resolved provider doesn't implement the interface (maybe print the runtime type's name to aid with debugging) and leave it at that.

test/Swashbuckle.AspNetCore.Cli.Test/ToolTests.cs
```diff
@@ -9,6 +9,20 @@ namespace Swashbuckle.AspNetCore.Cli.Test
...
+        public static void Can_Output_Swagger_Document_Names()
+        {
+            var result = RunListCommand((outputPath) =>
+            [
+                "list",
+                "--output",
+                outputPath,
+                Path.Combine(Directory.GetCurrentDirectory(), "MultipleVersions.dll")
+            ], nameof(Can_Output_Swagger_Document_Names));
+            var expected = $"Swagger Document Names:{Environment.NewLine}\"1.0\"{Environment.NewLine}\"2.0\"{Environment.NewLine}";
```
martincostello, line 22: I wonder if it's better to not have the `Swagger Document Names:` header so people could easily pipe the output into other tools?

test/Swashbuckle.AspNetCore.Cli.Test/ToolTests.cs
```diff
@@ -151,7 +165,22 @@ public static void Creates_New_Folder_Path()
...
+                : Path.Combine(temporaryDirectory.Path, "swagger.json");
+
+            string[] args = setup(outputPath);
+
+            Assert.Equal(0, Program.Main(args));
+
+            return File.ReadAllText(outputPath);
+        }
+
+        private static JsonDocument RunToFileCommand(Func<string, string[]> setup, string subOutputPath = default)
```
martincostello, line 183: This can be re-implemented in terms of calling `RunListCommand()` to avoid it being 90% the same:

```
private static JsonDocument RunToFileCommand(Func<string, string[]> setup, string subOutputPath = default)
{
    var json = RunListCommand(setup, subOutputPath);
    return JsonDocument.Parse(json);
}
```

src/Swashbuckle.AspNetCore.Cli/Program.cs
rassilon: I thnk I've fixed this as well.

test/Swashbuckle.AspNetCore.Cli.Test/ToolTests.cs
rassilon: I was thinking of spitting out content similar to CSV output, but I'm not hung up on the exact output shape too much one way or the other. So, I'll remove the header line.

src/Swashbuckle.AspNetCore.Cli/Program.cs
```diff
@@ -156,9 +127,110 @@ public static int Main(string[] args)
...
+                c.Option("--output", "");
+                c.OnRun((namedArgs) =>
+                {
+                    SetupAndRetrieveSwaggerProviderAndOptions(namedArgs, out var swaggerProvider, out var swaggerOptions);
+                    IList<string> docNames = new List<string>();
+                    string outputPath = null;
+
+                    outputPath = namedArgs.TryGetValue("--output", out var arg1)
+                        ? Path.Combine(Directory.GetCurrentDirectory(), arg1)
+                        : null;
```
martincostello, line 159:
```suggestion

                    string outputPath = namedArgs.TryGetValue("--output", out var arg1)
                        ? Path.Combine(Directory.GetCurrentDirectory(), arg1)
                        : null;
```

src/Swashbuckle.AspNetCore.Cli/Program.cs
```diff
@@ -156,9 +127,110 @@ public static int Main(string[] args)
...
+                        }
+                    }
+
+                    using Stream stream = outputViaConsole ? Console.OpenStandardOutput() : File.Create(outputPath);
+                    using StreamWriter writer = new(stream, outputViaConsole ? Console.OutputEncoding : Encoding.UTF8);
+
+                    if (swaggerProvider is not ISwaggerDocumentMetadataProvider docMetaProvider)
+                    {
+                        writer.WriteLine($"ERROR: The {nameof(ISwaggerProvider)} instance does not support {nameof(ISwaggerDocumentMetadataProvider)}, "
+                            + "so this coommand cannot list out the document names.");
```
martincostello, line 176:
```suggestion
                        writer.WriteLine($"The registered {nameof(ISwaggerProvider)} instance does not implement {nameof(ISwaggerDocumentMetadataProvider)}; unable to list the Swagger document names.");
```

src/Swashbuckle.AspNetCore.Cli/Program.cs
```diff
@@ -156,9 +127,110 @@ public static int Main(string[] args)
...
+                        return -1;
+                    }
+                    docNames = docMetaProvider.GetDocumentNames();
+
+                    foreach (var name in docNames)
+                    {
+                        writer.WriteLine($"\"{name}\"");
+                    }
+                    return 0;
```
martincostello, line 185:
```suggestion

                    docNames = docMetaProvider.GetDocumentNames();

                    foreach (var name in docNames)
                    {
                        writer.WriteLine($"\"{name}\"");
                    }

                    return 0;
```

test/Swashbuckle.AspNetCore.Cli.Test/ToolTests.cs
```diff
@@ -9,6 +9,20 @@ namespace Swashbuckle.AspNetCore.Cli.Test
 {
     public static class ToolTests
     {
+        [Fact]
+        public static void Can_Output_Swagger_Document_Names()
+        {
+            var result = RunListCommand((outputPath) =>
```
martincostello, line 15:
```suggestion
            var result = RunToStringCommand((outputPath) =>
```

test/Swashbuckle.AspNetCore.Cli.Test/ToolTests.cs
```diff
@@ -19,7 +33,7 @@ public static void Throws_When_Startup_Assembly_Does_Not_Exist()
         [Fact]
         public static void Can_Generate_Swagger_Json()
         {
-            using var document = RunApplication((outputPath) =>
+            using var document = RunToFileCommand((outputPath) =>
```
martincostello, line 36:
```suggestion
            using var document = RunToJsonCommand((outputPath) =>
```

test/Swashbuckle.AspNetCore.Cli.Test/ToolTests.cs
```diff
@@ -38,7 +52,7 @@ public static void Can_Generate_Swagger_Json()
         [Fact]
         public static void Overwrites_Existing_File()
         {
-            using var document = RunApplication((outputPath) =>
+            using var document = RunToFileCommand((outputPath) =>
```
martincostello, line 55:
```suggestion
            using var document = RunToJsonCommand((outputPath) =>
```

test/Swashbuckle.AspNetCore.Cli.Test/ToolTests.cs
```diff
@@ -62,7 +76,7 @@ public static void Overwrites_Existing_File()
         [Fact]
         public static void CustomDocumentSerializer_Writes_Custom_V2_Document()
         {
-            using var document = RunApplication((outputPath) =>
+            using var document = RunToFileCommand((outputPath) =>
```
martincostello, line 79:
```suggestion
            using var document = RunToJsonCommand((outputPath) =>
```

test/Swashbuckle.AspNetCore.Cli.Test/ToolTests.cs
```diff
@@ -80,7 +94,7 @@ public static void CustomDocumentSerializer_Writes_Custom_V2_Document()
         [Fact]
         public static void CustomDocumentSerializer_Writes_Custom_V3_Document()
         {
-            using var document = RunApplication((outputPath) =>
+            using var document = RunToFileCommand((outputPath) =>
```
martincostello, line 97:
```suggestion
            using var document = RunToJsonCommand((outputPath) =>
```

test/Swashbuckle.AspNetCore.Cli.Test/ToolTests.cs
```diff
@@ -98,7 +112,7 @@ public static void CustomDocumentSerializer_Writes_Custom_V3_Document()
         [Fact]
         public static void Can_Generate_Swagger_Json_ForTopLevelApp()
         {
-            using var document = RunApplication((outputPath) =>
+            using var document = RunToFileCommand((outputPath) =>
```
martincostello, line 115:
```suggestion
            using var document = RunToJsonCommand((outputPath) =>
```

test/Swashbuckle.AspNetCore.Cli.Test/ToolTests.cs
```diff
@@ -117,7 +131,7 @@ public static void Can_Generate_Swagger_Json_ForTopLevelApp()
         [Fact]
         public static void Does_Not_Run_Crashing_HostedService()
         {
-            using var document = RunApplication((outputPath) =>
+            using var document = RunToFileCommand((outputPath) =>
```
martincostello, line 134:
```suggestion
            using var document = RunToJsonCommand((outputPath) =>
```

test/Swashbuckle.AspNetCore.Cli.Test/ToolTests.cs
```diff
@@ -135,7 +149,7 @@ public static void Does_Not_Run_Crashing_HostedService()
         [Fact]
         public static void Creates_New_Folder_Path()
         {
-            using var document = RunApplication(outputPath =>
+            using var document = RunToFileCommand(outputPath =>
```
martincostello, line 152:
```suggestion
            using var document = RunToJsonCommand(outputPath =>
```

test/Swashbuckle.AspNetCore.Cli.Test/ToolTests.cs
```diff
@@ -151,7 +165,7 @@ public static void Creates_New_Folder_Path()
             Assert.True(productsPath.TryGetProperty("post", out _));
         }
 
-        private static JsonDocument RunApplication(Func<string, string[]> setup, string subOutputPath = default)
+        private static string RunListCommand(Func<string, string[]> setup, string subOutputPath = default)
```
martincostello, line 168:
```suggestion
        private static string RunToStringCommand(Func<string, string[]> setup, string subOutputPath = default)
```

test/Swashbuckle.AspNetCore.Cli.Test/ToolTests.cs
```diff
@@ -163,7 +177,12 @@ private static JsonDocument RunApplication(Func<string, string[]> setup, string
 
             Assert.Equal(0, Program.Main(args));
 
-            string json = File.ReadAllText(outputPath);
+            return File.ReadAllText(outputPath);
+        }
+
+        private static JsonDocument RunToFileCommand(Func<string, string[]> setup, string subOutputPath = default)
```
martincostello, line 183:
```suggestion
        private static JsonDocument RunToJsonCommand(Func<string, string[]> setup, string subOutputPath = default)
```

martincostello (approved):

test/Swashbuckle.AspNetCore.Cli.Test/ToolTests.cs
```diff
@@ -163,7 +177,12 @@ private static JsonDocument RunApplication(Func<string, string[]> setup, string
 
             Assert.Equal(0, Program.Main(args));
 
-            string json = File.ReadAllText(outputPath);
+            return File.ReadAllText(outputPath);
+        }
+
+        private static JsonDocument RunToJsonCommand(Func<string, string[]> setup, string subOutputPath = default)
+        {
+            string json = RunListCommand(setup, subOutputPath);
```
martincostello, line 185:
```suggestion
            string json = RunToStringCommand(setup, subOutputPath);
```

martincostello:
Thanks for you contribution!

rassilon:
> Thanks for you contribution!

Thanks for merging them in!

Issue #3099, opened 2024-10-09
[Feature request]: Support for OpenAPI Description in [ProducesResponseType], [Produces], and [ProducesDefaultResponseType] Attributes


sander1095:
### Is your feature request related to a specific problem? Or an existing feature?

Currently, in ASP.NET Core, the [ProducesResponseType], [Produces], and [ProducesDefaultResponseType] attributes enrich OpenAPI documentation by specifying response status codes, response types, and content types for controller actions. However, they lack support for OpenAPI's description property, which is vital for providing detailed context about the responses.

For instance, when an API returns a 422 status code, it is beneficial to include a description explaining why the error was triggered, enabling better client-side error handling. Currently, there is no standardized way to annotate response descriptions through these attributes.

Although some libraries (e.g., Swashbuckle) offer custom attributes like [SwaggerResponse] that support descriptions, these attributes are library-specific, potentially leading to compatibility and maintenance issues.

I'm working on https://github.com/dotnet/aspnetcore/issues/55656, which adds `Description` to these attributes. By introducing support for the Description property within these attributes, we can enrich OpenAPI documentation directly in ASP.NET Core in a library-agnostic way, improving developer experience and simplifying API documentation.

### Describe the solution you'd like

Swashbuckle.AspNetCore needs to support this new Description property

martincostello:
We can consider supporting this once your PR is merged and we start-up a development branch for .NET 10 (probably Februrary/March 2025 once preview 1 ships).

sander1095:
This is not stale :) It's been merged and we just need to wait for .NET 10 preview 1. @martincostello , do we want this marked as stale?

sander1095:
Commenting again to remove that stale label. .NET 10 preview 1 should be coming soon

sander1095:
Commenting again. The description property has been part of several previews now.

martincostello:
You can do a PR targeting the `dotnet-vnext` branch (see #3283) if you want to add support for this.

sander1095:
I sadly won't have time to look into this the next few months due to me moving house :(

martincostello:
Closing - this will be part of our major release expected for November that supports .NET 10 and OpenAPI 3.1.

PR #3283, opened 2025-03-02
Support .NET 10

martincostello:
Add support for .NET 10 by conditionally compiling various code to cater for differences between Microsoft.OpenApi v1 and v2.

This was started from #3252 and is very much a work-in-progress at this stage. The tests need more work to get them passing.

~~The general idea is to factor out any differences into shared helped classes that can be used to localise the Microsoft.OpenApi differences to as few places as possible. Once this gets further along, some of these changes can be cherry-picked into a new branch and merged into the default branch to minimise the diff for adding .NET 10 support.~~ #3285

### TODO

Before merge:

- [ ] Use stable version of .NET 10 SDK
- [ ] Use stable versions of Microsoft.AspNetCore NuGet packages
- [ ] Use stable version of Microsoft.OpenApi 2.0.0
- [ ] Remove NU5104 suppression
- [ ] Remove workarounds for prerelease issues
- [ ] Check `VersionPrefix` is correct for the next major version at the time of merge
- [ ] Use `OpenApiDocument.SerializeAs()` method
- [ ] Create migration guide (see #3365)
  - [ ] Include details such as for #3382
- [ ] Consume new .NET 10 `Description` properties where appropriate (see #3099)

---

Prerelease builds from this PR can be consumed from MyGet:

```xml
<PackageReference Include="Swashbuckle.AspNetCore" Version="9.0.0-pr.3283.*" />
```

```xml
<?xml version="1.0" encoding="utf-8"?>
<configuration>
  <packageSources>
    <clear />
    <add key="domaindrivendev" value="https://www.myget.org/F/domaindrivendev/api/v3/index.json" />
    <add key="NuGet" value="https://api.nuget.org/v3/index.json" />
  </packageSources>
</configuration>
```


test/WebSites/WebApi/WebApi.csproj
```diff
@@ -4,10 +4,15 @@
...
     <RootNamespace>WebApi</RootNamespace>
-    <TargetFrameworks>net9.0;net8.0</TargetFrameworks>
+    <TargetFrameworks>net10.0;net9.0;net8.0</TargetFrameworks>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
   </PropertyGroup>
 
+  <!-- TODO Workaround for error generated by https://github.com/dotnet/core/blob/main/release-notes/10.0/preview/preview1/aspnetcore.md#improvements-to-integration-testing-of-apps-with-top-level-statements -->
+  <PropertyGroup Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net10.0'))">
+    <NoWarn>$(NoWarn);CS1591</NoWarn>
+  </PropertyGroup>
```
martincostello, line 14: https://github.com/dotnet/aspnetcore/issues/60683

martincostello:
Some of the failing tests are https://github.com/microsoft/OpenAPI.NET/pull/2109.

Probably need to wait until .NET 10 preview 2 updates to a newer version of Microsoft.OpenApi to ingest any fixes.

martincostello:
Moved some of the changes into #3286 so we can prepare users for the change and also then change this PR to remove `SerializeAsV2` completely.

captainsafia:
@martincostello It seems like based on this changeset Swashbuckle.AspNetCore will have a similar constraint to Microsoft.AspNetCore.OpenApi in that OpenAPI 2.0 is _only_ used for .NET 10+ above.

Any thoughts about figuring out a way to support 2.0 in .NET 8 and .NET 9?

martincostello:
Not yet - I figured I'd start with the "easiest" option and get everything compiling and passing to unblock any apps using .NET 10, and then go from there.

I'm going to do some more work on this tomorrow(?) probably to try and get the tests passing, though I imagine I'll have to rejig a bunch of stuff next week anyway when 10.0.0-preview.2/2.0.0-preview7(?) ship.

Ideally OpenAPI 3.1 can be made available for our 8 and 9 users, but I'm slightly wary of the bigger "compatibility cliff" that creates with all the breaking changes and any other libraries that might extend Swashbuckle.

Good job there's another 7-8 months left to work that out on our side 😆

martincostello:
Issue discovered via https://github.com/martincostello/aspnetcore-openapi/pull/231 is that in some cases schema references go missing (like [this one](https://github.com/martincostello/aspnetcore-openapi/blob/3041e57ba9e42b3721095f32bf423a0c8afe8666/tests/TodoApp.Tests/OpenApiTests.Schema_Is_Correct_schemaUrl%3Dswagger.verified.txt#L31)).

I'm not sure why and there's been _a lot_ of API changes in that area since 2.0.0-preview5, so will have to wait until I've reacted to whatever changes ASP.NET Core 10 preview 2 brings and go from there.

martincostello:
Also, `preview5 => preview6` brings in more _huge_ breaking changes (`OpenApiSchema` becomes `IOpenApiSchema` and causes `OpenApiReference.Schema` to no longer work) so that might force our hand on whether we do multi-targeting or not, as it might cause an explosion of `#if` directives and make the code too hard to deal with the differences between v1 and v2.

captainsafia:
@martincostello Yes, I should've warned you that the preview6 breaking changes are a bit of a pain to wrangle. Here's the PR where I did it for ASP.NET Core in case it is helpful (https://github.com/dotnet/aspnetcore/pull/60269). The code surface in ASP.NET Core is admittedly smaller so most of the changes needed to be made in tests where I took shortcuts and used explicit casts for things I knew should be `OpenApiSchemaReference` for example.

> OpenApiSchema becomes IOpenApiSchema and causes OpenApiReference.Schema to no longer work

You've probably already gleened this from the implementation, but `OpenApiSchema` with a `Reference` becomes `OpenApiSchemaReference`. You should be able to resolve the properties on it directly and OpenAPI.NET will resolve to the underlying target.

This is assuming that the reference has been correctly initialized in the OpenAPI document's workspace. [Here is an extension method](https://github.com/dotnet/aspnetcore/blob/83a186cdb2cdd2644b7cb9f44f969a9464f46cc8/src/OpenApi/src/Extensions/OpenApiDocumentExtensions.cs) that I wrote in ASP.NET Core to help with this registration.

Directory.Packages.props
```diff
@@ -21,18 +21,21 @@
...
+    <PackageVersion Include="Microsoft.OpenApi.Readers" Version="2.0.0-preview7" />
     <PackageVersion Include="Microsoft.NET.Test.Sdk" Version="17.13.0" />
     <PackageVersion Include="Newtonsoft.Json" Version="13.0.3" />
     <PackageVersion Include="NSubstitute" Version="5.3.0" />
     <PackageVersion Include="NSwag.MSBuild" Version="14.3.0" />
     <PackageVersion Include="ReportGenerator" Version="5.4.4" />
-    <PackageVersion Include="System.Text.Json" Version="4.6.0" />
-    <PackageVersion Include="Verify.XunitV3" Version="29.1.0" />
+    <PackageVersion Include="System.Text.Json" Version="8.0.0" />
+    <PackageVersion Include="Verify.Xunit" Version="29.1.0" />
```
martincostello, line 32:
```suggestion
    <PackageVersion Include="Verify.XunitV3" Version="29.1.0" />
```

src/Swashbuckle.AspNetCore.ApiTesting/Swashbuckle.AspNetCore.ApiTesting.csproj
```diff
@@ -27,6 +27,10 @@
     <PackageReference Include="Microsoft.AspNetCore.Mvc.Testing" VersionOverride="9.0.0" />
   </ItemGroup>
 
+  <ItemGroup Condition=" '$(TargetFramework)' == 'net10.0' ">
+    <PackageReference Include="Microsoft.AspNetCore.Mvc.Testing" VersionOverride="10.0.0-preview.1.25120.3" />
```
martincostello, line 31:
```suggestion
    <PackageReference Include="Microsoft.AspNetCore.Mvc.Testing" VersionOverride="10.0.0-preview.3.25172.1" />
```

src/Swashbuckle.AspNetCore/Swashbuckle.AspNetCore.csproj
```diff
@@ -24,6 +24,10 @@
     <MicrosoftExtensionsApiDescriptionServerPackageVersion>9.0.0</MicrosoftExtensionsApiDescriptionServerPackageVersion>
   </PropertyGroup>
 
+  <PropertyGroup Condition=" '$(TargetFramework)' == 'net10.0' ">
+    <MicrosoftExtensionsApiDescriptionServerPackageVersion>10.0.0-preview.1.25120.3</MicrosoftExtensionsApiDescriptionServerPackageVersion>
```
martincostello, line 28:
```suggestion
    <MicrosoftExtensionsApiDescriptionServerPackageVersion>10.0.0-preview.3.25172.1</MicrosoftExtensionsApiDescriptionServerPackageVersion>
```

Poltuu:
`Microsoft.OpenApi.Models.OperationType` has been removed from the library in preview 13 (https://github.com/microsoft/OpenAPI.NET/commit/8baff287aa9450ad3bd467816de321e30157bcb3 ), but is still referenced by SwaggerGen.
Any chance you could maybe upgrade Microsoft.OpenApi on this branch🙏 . I cannot downgrade myself as i need the fixs in preview14 and 15.

martincostello:
Hi @Poltuu - The short answer is no. Because ASP.NET Core 10 is compiled against a specific version, being able to move the preview forward incrementally will eventually hit a "compatibility wall" due to the breaking changes meaning you can't creep further forward.

The PR here will move forward to match the same version being used in the related .NET 10 preview as each one ships. For example, https://github.com/dotnet/aspnetcore/pull/61541 is moving to preview.17, so in preview 5(?) we'll target the same.

I understand your frustration - there's changes I've made to the library myself that I'd like to consume that I can't currently benefit from either.

Poltuu:
thx for the info, that's very unfortunate for me, i hope the PR you mentionned will make it to the next preview

martincostello:
@Poltuu FYI ASP.NET Core 10 preview.4 will include Microsoft.OpenApi 2.0.0-preview.17.

bkoelman:
> Ideally OpenAPI 3.1 can be made available for our 8 and 9 users, but I'm slightly wary of the bigger "compatibility cliff" that creates with all the breaking changes and any other libraries that might extend Swashbuckle.
> 
> Good job there's another 7-8 months left to work that out on our side 😆

I feel your pain. I'm facing similar issues. We heavily use Swasbuckle extensibility to [produce custom component schemas](https://github.com/json-api-dotnet/JsonApiDotNetCore/tree/master/src/JsonApiDotNetCore.OpenApi.Swashbuckle/SchemaGenerators). The split-up between `OpenApiSchema` and `OpenApiSchemaReference` cascades down the entire ecosystem. The experience is a bit like adopting nullable reference types, re-assessing every usage throughout the codebase. Though I believe the split-up adds clarity. Until now, we'd typically encode this information in variable/parameter names because it's so confusing when something is a schema or a reference.

I haven't figured out all the details yet, but my plan to solve multi-targeting is the following:
- Change signatures from `OpenApiSchema` to `IOpenApiSchema` / `OpenApiSchema` / `OpenApiSchemaReference`.
- Add internal helpers to deal with API changes for nullability, creating a reference schema, etc.
- Then restore support for .NET 8/9 by using namespace aliases. This avoids `#ifdef` all over the codebase, consolidating that to a small compatibility layer.

The compatibility layer looks something like this:
```c#
#if NET10_0
global using IOpenApiSchema = Microsoft.OpenApi.Models.Interfaces.IOpenApiSchema;
global using OpenApiSchema = Microsoft.OpenApi.Models.OpenApiSchema;
global using OpenApiSchemaReference = Microsoft.OpenApi.Models.References.OpenApiSchemaReference;
#else
global using IOpenApiSchema = Microsoft.OpenApi.Models.OpenApiSchema;
global using OpenApiSchema = Microsoft.OpenApi.Models.OpenApiSchema;
global using OpenApiSchemaReference = Microsoft.OpenApi.Models.OpenApiSchema;
#endif

internal static class OpenApiCompatibilityLayer
{
#if NET10_0
    public const JsonSchemaType StringSchemaType = JsonSchemaType.String;
    public const JsonSchemaType ObjectSchemaType = JsonSchemaType.Object;
    public const JsonSchemaType BooleanSchemaType = JsonSchemaType.Boolean;
    public const JsonSchemaType IntegerSchemaType = JsonSchemaType.Integer;
#else
    public const string StringSchemaType = "string";
    public const string ObjectSchemaType = "object";
    public const string BooleanSchemaType = "boolean";
    public const string IntegerSchemaType = "integer";
#endif

    public static void SetNullable(this OpenApiSchema openApiSchema, bool nullable) ......
}
```

By multi-targeting, users on .NET 8/9 will get the old API shape (and a transitive reference to Microsoft.OpenAPI v1.x), while users targeting .NET 10 will get the new API shape (and a transitive reference to Microsoft.OpenAPI v2.x). We are fortunate enough that we don't publicly expose APIs from Microsoft.OpenAPI, so we can get away with that. For Swashbuckle, it would mean the public API varies per target framework, which may not be desirable.

src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/OpenApiSchemaExtensions.cs
```diff
@@ -229,40 +252,38 @@ private static void ApplyBase64Attribute(OpenApiSchema schema)
 
     private static void ApplyRangeAttribute(OpenApiSchema schema, RangeAttribute rangeAttribute)
     {
-#if NET
+        schema.Maximum = decimal.TryParse(rangeAttribute.Maximum.ToString(), out decimal maximum)
```
bkoelman, line 255: Why parse with the current culture, then format using invariant? I think it should take [RangeAttribute.ParseLimitsInInvariantCulture](https://learn.microsoft.com/en-us/dotnet/api/system.componentmodel.dataannotations.rangeattribute.parselimitsininvariantculture?view=net-9.0) and [RangeAttribute.ConvertValueInInvariant](https://learn.microsoft.com/en-us/dotnet/api/system.componentmodel.dataannotations.rangeattribute.convertvalueininvariantculture?view=net-9.0) into account.

src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/SchemaGenerator.cs
```diff
@@ -192,20 +197,20 @@ private OpenApiSchema GenerateSchemaForParameter(
...
 
-        if (schema.Reference == null)
+        if (schema is not OpenApiSchemaReference)
         {
             ApplyFilters(schema, modelType, schemaRepository);
-            if (Nullable.GetUnderlyingType(modelType) != null)
+            if (Nullable.GetUnderlyingType(modelType) != null && schema is OpenApiSchema concrete)
             {
-                schema.Nullable = true;
+                concrete.Type |= JsonSchemaType.Null;
```
bkoelman, line 213:
```suggestion
                SetNullable(concrete, true);
```

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
```diff
@@ -116,9 +129,38 @@ public OpenApiDocument GetSwagger(string documentName, string host = null, strin
...
+        static void SortSchema(IOpenApiSchema schema)
+        {
+            if (schema is OpenApiSchema concrete)
+            {
+                if (concrete.Required is { Count: > 1 } required)
+                {
+                    concrete.Required = [.. new SortedSet<string>(required)];
+                }
+
+                if (concrete.AllOf is { Count: > 0 } allOf)
```
bkoelman, line 155: This should also happen for `oneOf` and `anyOf`.

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
```diff
@@ -420,24 +465,32 @@ private async Task<OpenApiOperation> GenerateOpenApiOperationFromMetadataAsync(A
         }
 
         // Schemas will be generated via Swashbuckle by default.
-        foreach (var parameter in operation.Parameters)
+        if (operation.Parameters is { Count: > 0 } parameters)
         {
-            var apiParameter = apiDescription.ParameterDescriptions.SingleOrDefault(desc => desc.Name == parameter.Name && !desc.IsFromBody() && !desc.IsFromForm() && !desc.IsIllegalHeaderParameter());
-            if (apiParameter is not null)
+            foreach (var parameter in parameters)
```
bkoelman, line 470: I haven't tried, but I _think_ this could be written as: `foreach (var parameter in parameters ?? [])`, to reduce the nesting.

src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/OpenApiSchemaExtensions.cs
martincostello: I tried to follow the current behaviour, rather than do what I would do in a vacuum with no weight of backwards compatibility (which would be to be invariant without exception).

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
martincostello: I haven't tried using C# 14-isms yet (and I think I'd need to opt in to preview language version).

src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/OpenApiSchemaExtensions.cs
bkoelman: The current behavior is that these types are numeric, not strings, so there's nothing to follow. Always invariant is simply wrong; the properties on RangeAttribute exist for a reason.

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
bkoelman: I don't understand, the `??` operator is very old and used in lots of other places already.

martincostello: `?? []` is the part I'm referring to when used within a foreach. It typically doesn't work as it doesn't know the target type to use.

martincostello: Actually, I think I'd confused it with https://github.com/dotnet/csharplang/issues/7913

martincostello:
@Poltuu Sorry that's currently broken due to #3368.

As a workaround you can get the NuGet packages from the CI build and either consume them from disk as a local package source, or upload them into your own custom NuGet feed.

See here for an example: https://github.com/martincostello/aspnetcore-openapi/pull/231/commits/d325e2ab0dbab665acdfe830adc2037e6affbdaf

src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/OpenApiSchemaExtensions.cs
```diff
@@ -229,40 +252,38 @@ private static void ApplyBase64Attribute(OpenApiSchema schema)
...
+            ? maximum.ToString(CultureInfo.InvariantCulture)
+            : schema.Maximum;
+
+        schema.Minimum = decimal.TryParse(rangeAttribute.Minimum.ToString(), out decimal minimum)
+            ? minimum.ToString()
```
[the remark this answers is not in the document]
bkoelman: Fix at #3426.

src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/OpenApiSchemaExtensions.cs
```diff
@@ -229,40 +252,49 @@ private static void ApplyBase64Attribute(OpenApiSchema schema)
 
     private static void ApplyRangeAttribute(OpenApiSchema schema, RangeAttribute rangeAttribute)
     {
-#if NET
+        // Use InvariantCulture if explicitly requested or if the range has been set via the
+        // RangeAttribute(double, double) or RangeAttribute(int, int) constructors.
```
bkoelman: Can you elaborate on why this was changed?

It doesn't seem to take additional culture-sensitive types into account, such as `float`, `decimal`, `DateTime`, `DateTimeOffset`, `TimeSpan`, etc. It also doesn't work with user-defined types for which a `TypeConverter` is provided.

martincostello: https://github.com/dotnet/aspnetcore/blob/6a96694cda568cc8ebec2a0b34e408842eb9805f/src/OpenApi/src/Extensions/JsonNodeSchemaExtensions.cs#L94-L98

bkoelman: Thanks for the link. Please revert 529a32446ae2b390da91f7648cb3db856d00eb99, it's just flawed. Swashbuckle is superior to ASP.NET's OpenAPI in many ways, which is why we stick with it.

martincostello: I suggest you raise an issue there with concrete examples of what doesn't work correctly so it can be fixed, and then this code can mirror it.

Ultimately the intention is that both implementations are accounting for the legacy quirks of how `[Range]` works with multiple different types of parsing behaviours.

We don't want to generate invalid JSON, like https://github.com/dotnet/aspnetcore/pull/62193 fixes (which is how I saw the more specific handling for RangeAttribute in the first place).

bkoelman: > I suggest you raise an issue there with concrete examples of what doesn't work correctly so it can be fixed, and then this code can mirror it.

Really? So YOU replace _working code_ with broken code from a competing product, and then expect ME to put effort into getting _that competing product_ fixed? That's ridiculous.

Based on our tests, I identified issues in Swashbuckle that I fixed. Now you've replaced that fix with flawed code. Don't change code if you're unable or unwilling to properly test, either automated or manually, or establish that the changes are correct by looking at the underlying implementation of `RangeAttribute`. Don't be so naive as to assume any code written by Microsoft must be superior to anything else, so it can be pulled without thinking about it.

> Ultimately the intention is that both implementations are accounting for the legacy quirks of how [Range] works with multiple different types of parsing behaviours.

What quirks? Look at the tests [here](https://github.com/dotnet/runtime/blob/main/src/libraries/System.ComponentModel.Annotations/tests/System/ComponentModel/DataAnnotations/RangeAttributeTests.cs). Copying OpenAPI code, which obviously doesn't align with that, makes no sense. By looking at [RangeAttribute source](https://github.com/dotnet/runtime/blob/main/src/libraries/System.ComponentModel.Annotations/src/System/ComponentModel/DataAnnotations/RangeAttribute.cs#L52), it's pretty obvious how wrong the changed code is. Check the constructor that takes strings.

> We don't want to generate invalid JSON, like https://github.com/dotnet/aspnetcore/pull/62193 fixes (which is how I saw the more specific handling for RangeAttribute in the first place).

That fixes an issue in ASP.NET OpenAPI, which was already working properly in Swashbuckle. There's no reason to replace Swashbuckle's logic with their crappy Range-related code, which introduces _new_ problems in Swashbuckle.

bkoelman: The _only_ improvement to consider is:
```c#
rangeAttribute.ParseLimitsInInvariantCulture || rangeAttribute.Minimum is double || rangeAttribute.Maximum is int
```
For the theoretical case where someone would explicitly set `ParseLimitsInInvariantCulture`, combined with passing numeric values for min/max instead of strings.

The remaining code, converting to string, parsing that, and converting that to string again, is just silly.

bkoelman: Upon further inspection the new code is actually correct. `Convert.ToString` takes custom type converters and `IConvertible` into account. So it's fine to keep.

But my point still stands: don't pull code without fully understanding it.

martincostello: > So YOU replace working code with broken code from a competing product, and then expect ME to put effort into getting that competing product fixed?

1. It's not a competitor, it's an alternative. I contribute to both.
2. You were stating the code was broken, despite it having passing tests. The suggestion that you raise an issue was because there was no information provided by you as to how exactly it was broken. I can't exactly raise an issue that's just _"This person says the code is wrong somehow - fix it"_. You're much better placed to explain what the problem you think you've found is.
    - Alternatively if both were equally broken, then you could have just provided that information here and then I could later have made the same fixes to ASP.NET Core.
    - Some people think people raising PRs or issues on their behalf is _"taking their credit"_, hence the offer to you to report the problem(s) that you claimed to have found. 

> Based on our tests, I identified issues in Swashbuckle that I fixed.

You did contribute a fix yes, but with no tests.

https://github.com/domaindrivendev/Swashbuckle.AspNetCore/pull/3426#issuecomment-2909108440

You were happy to let me test it for you though. I added some tests for the scenario (localization) here: #3431.

That then flushed out a bug in the change you made, because the OpenAPI document was generating broken code (commas for decimals instead of dots).

I fixed that here: 9e6dedfdded3d8d32ce4b4268b22a005f630f20e

> Don't change code if you're unable or unwilling to properly test, either automated or manually

We've established that it was tested. Both here as stated above, and also in this PR, which is a variant of the same underlying problem: https://github.com/dotnet/aspnetcore/pull/62193

> Don't be so naive as to assume any code written by Microsoft must be superior to anything else, so it can be pulled without thinking about it.

Don't assume people's thought processes or intent. I saw the code had the additional parts of the conditional statement while working on the bug fix there, and from recency of working on the code here realised that we didn't have the same handling of the way `[Range]` works.

> What quirks?

The fact that `[Range]` uses the current culture for some code paths, and invariant for others. I would consider that a "quirk" as the behaviour isn't consistent.

> which introduces new problems in Swashbuckle

I'm all ears to fix any actual issues, but I need test cases that demonstrate what these scenarios are. Then issues are easy to fix and stop from regressing in the future.

> The only improvement to consider is:
> ```rangeAttribute.ParseLimitsInInvariantCulture || rangeAttribute.Minimum is double || rangeAttribute.Maximum is int```

That is exactly the change that 529a32446ae2b390da91f7648cb3db856d00eb99 introduced.

The rest of it is basically style refactoring.

> The remaining code, converting to string, parsing that, and converting that to string again

The old code already did that. So either it was already wrong and we didn't know, and it's still wrong and just looks different, or it was right in the first place and now is still right but just looks different.

> Upon further inspection the new code is actually correct.

It's almost like I knew what `Convert.ToString()` does and made an informed choice.

> But my point still stands: don't pull code without fully understanding it.

One could say the same about critiquing other people's efforts and motives.

martincostello: - Various tests cases added by 6c17312a46314d5a9a7b6dd9f47a57c4f9eee20a. Additional suggestions for any missing cases welcome.
- Refactored again in dd8b516ab575f05b7ba810343394363863da491d to make the code clearer and avoid always formatting and re-parsing the values if they aren't strings.

bkoelman: @martincostello You're right, and I apologize. I jumped to conclusions too quickly. I just had a bad day. With so many things going on, I barely have time for anything. Then I saw that previously working (and manually tested) code here got changed again, which meant I had to spend time on that too, and that frustrated me. Combined with earlier poor experiences I've had with ASP.NET OpenAPI, that was the last straw. However, after looking at these changes for a while, I realized it wasn't so bad at all. It's actually fine.

I appreciate your efforts here by adding more tests. The interesting case where `Convert` over `ToString` matters is when custom type converters are involved. For example, a user-defined type `JavaTimestamp` that provides conversion to/from an int. Or a `Fractional` type that internally stores the numerator and denominator, but provides conversion to floating-point types. In both cases, they are not required to implement a meaningful `ToString` method.

PS: I'm not at all concerned with anyone taking credit; I barely get around to logging issues in the first place. In my experience, other OSS projects often expect the bug/regression reporter to also implement the fix, or they'll close or abandon the issue. Years ago, maintainers were generally thankful that someone spent their free time creating a minimal repro, but it seems those days are gone. Either way, thank you for pointing out how other people may feel about taking credit.

martincostello: Would you be able to give some example code for one of your custom types using a custom type converter and how it integrates into Swashbuckle? The `Fractional` one sounds like the best example we could use for a test out of the two you've given.

I got my attempt working correctly at runtime with System.Text.Json, but Swashbuckle wasn't describing it as a primitive float value.

I understand the concept, but playing around with it locally it's not quite fitting into the OpenAPI documentation generation, so I'm not quite following how it feeds through into the call site for the `Convert.ToString()` call. I have a hunch how it does, but writing a test centred around a hunch isn't necessarily going to test the actual reality even if it's the correct intent.

bkoelman: Okay, here's a demo that illustrates how ASP.NET model binding and range validation are affected by culture settings.

[RangeDemo.zip](https://github.com/user-attachments/files/20644645/RangeDemo.zip)

Use the `.http` file for trying it out. View the Output Window for what's happening. Toggle using the commented line in `.csproj`.

bkoelman: For Swashbuckle to recognize a `Fraction` as a `double`, change the long `ToString` method in `Fraction` to return `ToDouble().ToString(formatProvider)` instead of `Format()`.

martincostello: Thanks for the code. However, I'm still getting the same as I was with what I tried yesterday.

In the OpenAPI document the body is an object rather than "just a number":

image

It also doesn't have the maximum or minimum values from the range applied.

Or am I completely misunderstanding the intent here?

What I thought was going on is that you have a number going over the wire and this gets transparently converted into an object when (de)serialized.

My understanding is that `minimum` and `maximum` can only be numbers (in Microsoft.OpenApi 1.x they're typed as a `decimal`), so that's where I'm confused with custom maximum/minimum ranges if your JSON is expected to be a string representing a fraction (e.g. `"1/2"` for `0.5`).

I feel like I'm understanding the individual LEGO bricks here, but I'm not seeing how they fit together 😅

bkoelman: Well, we need Swashbuckle to treat `Fraction` as a primitive type, otherwise it won't add min/max properties. I've revised the original project by pushing an additional commit to add Swashbuckle support, and to allow both fractional and decimal notation.

See https://github.com/bkoelman/RangeValidationDemo. Turns out that Swashbuckle still can't handle it properly. I've created a branch with a fix for that.

martincostello: Awesome - thanks for the update. I'll take a look at it properly when I get the chance.

bkoelman: Note: I've added a `http-client.env.json`, which lists the environments to use with the `.http` file. Select the environment (upper right in VS) that matches what's set in the project file when testing. As before, various toggles are available in the project file.

Directory.Packages.props
```diff
@@ -1,40 +1,40 @@
...
+    <PackageVersion Include="Microsoft.AspNetCore.Mvc.Versioning.ApiExplorer" Version="5.1.0" />
+    <PackageVersion Include="Microsoft.AspNetCore.OpenApi" Version="10.0.0-preview.5.25277.114" />
+    <PackageVersion Include="Microsoft.AspNetCore.Razor.Design" Version="2.3.0" />
+    <PackageVersion Include="Microsoft.AspNetCore.Routing" Version="2.3.0" />
+    <PackageVersion Include="Microsoft.AspNetCore.StaticFiles" Version="2.3.0" />
+    <PackageVersion Include="Microsoft.AspNetCore.TestHost" Version="6.0.32" />
+    <PackageVersion Include="Microsoft.CodeAnalysis.PublicApiAnalyzers" Version="4.14.0" />
+    <PackageVersion Include="Microsoft.Extensions.FileProviders.Embedded" Version="2.1.1" />
+    <PackageVersion Include="Microsoft.OpenApi" Version="2.0.0-preview.17" />
+    <PackageVersion Include="Microsoft.OpenApi.YamlReader" Version="2.0.0-preview.17" />
```
martincostello:
```suggestion
    <PackageVersion Include="Microsoft.OpenApi" Version="2.0.0-preview.18" />
    <PackageVersion Include="Microsoft.OpenApi.YamlReader" Version="2.0.0-preview.18" />
```

src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/OpenApiSchemaExtensions.cs
martincostello: There's no rush, but are you planning on doing a PR to fix the range handling for custom types? If you're too busy (or not planning on it) I'm happy to pick the code out of your fork to pull it in.

bkoelman: I've revised the implementation quite a bit. Last time, I didn't understand why the conversion inside RangeAttribute isn't being cached, which is why I had the two code paths. Turns out that [RangeAttributeAdapter](https://github.com/dotnet/aspnetcore/blob/16438b6e3e694a76186a49a8e700548423e549ea/src/Mvc/Mvc.DataAnnotations/src/RangeAttributeAdapter.cs) is responsible for that: because running the conversion isn't thread-safe, a new RangeAttribute instance is created for each request.

I've pushed a new commit in both repositories. Running against that, most tests in Swashbuckle are green now. The remaining failures are invalid tests.

I'm fine if you take it from here. When adding tests, feel free to use `Fraction` and its related code, or just create something simpler. These are the minimal requirements:
- ExampleType must implement the non-generic `IComparable` (`RangeAttribute` requires that).
- ExampleType's `ToString()` returns garbage, to prove we never depend on that.
- ExampleType is decorated with a `TypeConverter` that provides conversion from/to `decimal`.
- The TypeConverter substitutes an incoming `null` culture with the current culture in its `ConvertFrom`/`ConvertTo` methods.

martincostello: For clarity, could you state here in a comment as JSON what you expect the JSON payload to look like?

Running your demo it renders as a double:

image

I assume it's intended to be sent over the wire as a string?

```json
{
  "fraction": "1/10"
}
```

Just want to confirm separate from the code so I know exactly what we're aiming for.

If it is a string, then the thing I'm unsure about is whether `maximum` and `minimum` being returned in the schema is actually valid. From what I could see, they seem to only be associated with numeric types, and it seems weird to me to have a JSON schema have a string property that has a numeric maximum and minimum as the types don't match.

----

Out of curiosity, is there a specific reason you use `:` for fractions? A cursory ask of Copilot suggest to me that `/` is standard in de-DE but `:` is sometimes used in recipes.

This is what it told me:

| Language/Culture      | Example Fraction | Separator | Notes                                 |
|-----------------------|------------------|-----------|---------------------------------------|
| English (en-US, en-GB)| 1/2              | /       | Most common in English-speaking locales|
| French (fr-FR)        | 1/2              | /       | / is also standard in French        |
| German (de-DE)        | 1/2              | /       | / is standard, but sometimes : is used in recipes|
| Russian (ru-RU)       | 1/2              | /       | / is common                         |
| Arabic (ar)           | ١/٢              | /       | Arabic numerals, / as separator     |
| Swedish (sv-SE)       | 1/2              | /       | / is standard                       |
| Finnish (fi-FI)       | 1/2              | :       | : is sometimes used, especially in recipes|
| Hungarian (hu-HU)     | 1/2              | :       | : is common in recipes              |
| Turkish (tr-TR)       | 1/2              | /       | / is standard                       |
| Japanese (ja-JP)      | 1/2              | /       | / is used, but vertical writing may use special notation|

martincostello: I _think_ this is what would be correct for the intention.

With the following schema filter the property is described as a string that can look like a number or a fraction:

```csharp
public class FractionSchemaFilter : ISchemaFilter
{
    public void Apply(IOpenApiSchema schema, SchemaFilterContext context)
    {
        if (context.Type == typeof(Fraction) && schema is OpenApiSchema concrete)
        {
            concrete.Type = JsonSchemaType.String;
            concrete.Pattern = "\\d+((\\/|:)\\d+)?";
            concrete.Properties?.Clear();
        }
    }
}
```

image

The underlying serializer accepts a number _or_ a fraction, so to work correctly on the above basis it would just need tweaking to accept a string that parsed directly to a number if it wasn't a fraction as right now with the code as-is they're rejected:

image

bkoelman: This is what it looks like on my system:

{676E413D-C5E5-4E0F-8C75-D2E77DF4A601}

```json
"components": {
  "schemas": {
    "ExampleModel": {
      "type": "object",
      "properties": {
        "fraction": {
          "maximum": 0.5,
          "exclusiveMaximum": true,
          "minimum": 0.1,
          "type": "number",
          "nullable": true,
          "format": "double"
        },
        "someNumber1": {
          "maximum": 22.2,
          "minimum": 11.1,
          "type": "number",
          "nullable": true,
          "format": "double"
        },
        "someNumber2": {
          "maximum": 44.4,
          "minimum": 33.3,
          "type": "number",
          "nullable": true,
          "format": "double"
        }
      },
      "additionalProperties": false
    }
  }
}
```

Did you adapt the project references to where you cloned my fork?

> I assume it's intended to be sent over the wire as a string?

When I run the .http file, it sends either:
```json
{
    "fraction": "1:5"
}
```
or:
```json
{
    "fraction": 0.2
}
```
For only the last of these, OpenAPI support is provided in my demo. Because the goal is to test min/max support, which has no meaning for a string in OpenAPI.

> Out of curiosity, is there a specific reason you use : for fractions?

Not really; I just needed an alternative syntax to differentiate between cultures for the demo. In Dutch, we could write the equation `5 : 6 = 5/6`, which means: 5 divided by 6 equals 5/6, that's how I came up with it.

> With the following schema filter the property is described as a string that can look like a number or a fraction:

Perhaps, but I haven't even considered that. My only goal was to analyze min/max handling in OpenAPI, so I didn't look into any form of string representation in OpenAPI.

> The underlying serializer accepts a number or a fraction, so to work correctly on the above basis it would just need tweaking to accept a string that parsed directly to a number if it wasn't a fraction as right now with the code as-is they're rejected:

No, the way the API works is that it accepts either a JSON number, or a fraction in string format. Only the former is expressed in OpenAPI, for the reasons explained above. In theory, this could be represented as a `oneOf` in OpenAPI, consisting of either the number with min/max, or the string with a regex. But I didn't pursue such an approach, because I only cared about min/max handling.

martincostello: Ok thanks for the information.

No, I didn't change the project references I just ran it with the NuGet references.

I'll look to fix it from a "numbers with secret fractions" perspective, rather than a "strings with secret numbers" perspective (which is how I'd probably tackle such a thing).

("secret" only because it's not in the OpenAPI spec so if I couldn't see the code I wouldn't know it had fraction strings)

bkoelman: > No, I didn't change the project references I just ran it with the NuGet references.

Ah, that only demonstrates it doesn't currently work. See the instructions at https://github.com/bkoelman/RangeValidationDemo/blob/master/README.md.

test/Swashbuckle.AspNetCore.IntegrationTests/snapshots/VerifyTests.Swagger_IsValidJson_No_Startup_entryPointType=WebApi.Program_swaggerRequestUri=v1.DotNet8_0.verified.txt
```diff
@@ -531,7 +530,6 @@
           {
             "name": "queryParameter",
             "in": "query",
-            "description": "queryParameter Description",
```
jgarciadelanoceda, line 534: It seems missing . I think that is because you put a If dotnet>10 for WithOpenApi, shouldn't it be if dotnet <10?

test/Swashbuckle.AspNetCore.IntegrationTests/snapshots/VerifyTests.TypesAreRenderedCorrectly.DotNet10_0.verified.txt
```diff
@@ -0,0 +1,1221 @@
+﻿{
+  "openapi": "3.0.4",
```
jgarciadelanoceda, line 2: Why is it not 3.1?
I launched the swagger and show an error because wrong open API version

test/WebSites/OAuth2Integration/OAuth2Integration.csproj
```diff
@@ -20,4 +20,9 @@
     <PackageReference Include="Duende.IdentityServer" VersionOverride="7.2.4" />
   </ItemGroup>
 
+  <ItemGroup Condition="'$(TargetFramework)' == 'net10.0'">
+    <PackageReference Include="Microsoft.AspNetCore.Authentication.JwtBearer" VersionOverride="10.0.0-preview.5.25277.114" />
+    <PackageReference Include="Duende.IdentityServer" VersionOverride="7.1.1" />
```
martincostello:
```suggestion
    <PackageReference Include="Duende.IdentityServer" VersionOverride="7.2.4" />
```

test/WebSites/TestFirst/TestFirst.csproj
```diff
@@ -7,4 +7,5 @@
   <ItemGroup>
     <ProjectReference Include="..\..\..\src\Swashbuckle.AspNetCore.SwaggerUI\Swashbuckle.AspNetCore.SwaggerUI.csproj" />
   </ItemGroup>
+
```
martincostello:
```suggestion
```

test/Swashbuckle.AspNetCore.IntegrationTests/snapshots/VerifyTests.TypesAreRenderedCorrectly.DotNet10_0.verified.txt
martincostello: I have intentionally not changed the default version to 3.1.

jgarciadelanoceda: But there must be a problem with the SwaggerUi in the repo because it's giving this on the main page
image

And also it may be interesting to do IntegrationTest on 3.1 version, if you want a PR I am happy to do it :)

martincostello: Force clear your browser cache, you're probably hitting this: #3336

Directory.Packages.props
```diff
@@ -12,11 +12,11 @@
     <PackageVersion Include="Microsoft.AspNetCore.Mvc.Testing" Version="8.0.17" />
     <PackageVersion Include="Microsoft.AspNetCore.Mvc.Versioning" Version="5.1.0" />
     <PackageVersion Include="Microsoft.AspNetCore.Mvc.Versioning.ApiExplorer" Version="5.1.0" />
-    <PackageVersion Include="Microsoft.AspNetCore.OpenApi" Version="8.0.17" />
+    <PackageVersion Include="Microsoft.AspNetCore.OpenApi" Version="10.0.0-preview.5.25277.114" />
     <PackageVersion Include="Microsoft.AspNetCore.TestHost" Version="8.0.17" />
     <PackageVersion Include="Microsoft.CodeAnalysis.PublicApiAnalyzers" Version="4.14.0" />
-    <PackageVersion Include="Microsoft.OpenApi" Version="1.6.23" />
-    <PackageVersion Include="Microsoft.OpenApi.Readers" Version="1.6.23" />
+    <PackageVersion Include="Microsoft.OpenApi" Version="2.0.0-preview.18" />
```
Frulfump: Microsoft.OpenApi is now GA https://github.com/microsoft/OpenAPI.NET/releases/tag/v2.0.0 🎉

martincostello: Yep, I'm aware 🙂 - however due to various breaking changes, Swashbuckle has to be in lock-step with the version that ASP.NET Core is compiled against. See #3479.

src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/SchemaGenerator.cs
```diff
@@ -50,13 +51,15 @@ private OpenApiSchema GenerateSchemaForMember(
             ? GeneratePolymorphicSchema(schemaRepository, knownTypesDataContracts)
             : GenerateConcreteSchema(dataContract, schemaRepository);
 
-        if (_generatorOptions.UseAllOfToExtendReferenceSchemas && schema.Reference != null)
+        if (_generatorOptions.UseAllOfToExtendReferenceSchemas && schema is OpenApiSchemaReference reference)
```
jgarciadelanoceda, line 52: @martincostello this could be a Breaking change, but I think that the UseAllOfToExtendReference schemas was created only because the OpenApi 3.0 did not support ref siblings.
So I ask myself if we should Obsolete? the parameter... But it still is needed for OpenApi < 3.1

martincostello: This entire PR is lots of breaking changes unfortunately. The object model changes apply to all OpenAPI versions. The representation differences when serialized to JSON/YAML are dealt with internally in the Microsoft.OpenApi library.

Directory.Packages.props
martincostello: 2.0.0 seems to have no issues as of .NET 10 preview 6, so I've updated the dependency. It can be moved back if it causes issues and then updated again in preview 7.

test/WebSites/OAuth2Integration/OAuth2Integration.csproj
```diff
@@ -20,4 +20,9 @@
     <PackageReference Include="Duende.IdentityServer" VersionOverride="7.3.1" />
   </ItemGroup>
 
+  <ItemGroup Condition="'$(TargetFramework)' == 'net10.0'">
+    <PackageReference Include="Microsoft.AspNetCore.Authentication.JwtBearer" VersionOverride="10.0.0-preview.7.25380.108" />
+    <PackageReference Include="Duende.IdentityServer" VersionOverride="7.2.4" />
```
martincostello:
```suggestion
    <PackageReference Include="Duende.IdentityServer" VersionOverride="7.3.1" />
```

test/WebSites/OAuth2Integration/OAuth2Integration.csproj
```diff
@@ -20,4 +20,9 @@
     <PackageReference Include="Duende.IdentityServer" VersionOverride="7.3.2" />
   </ItemGroup>
 
+  <ItemGroup Condition="'$(TargetFramework)' == 'net10.0'">
+    <PackageReference Include="Microsoft.AspNetCore.Authentication.JwtBearer" VersionOverride="10.0.0-rc.1.25451.107" />
+    <PackageReference Include="Duende.IdentityServer" VersionOverride="7.3.1" />
```
martincostello:
```suggestion
    <PackageReference Include="Duende.IdentityServer" VersionOverride="7.3.2" />
```

Directory.Packages.props
```diff
@@ -1,31 +1,31 @@
...
+    <PackageVersion Include="JunitXml.TestLogger" Version="6.1.0" />
+    <PackageVersion Include="MartinCostello.Logging.XUnit.v3" Version="0.6.0" />
+    <PackageVersion Include="Microsoft.AspNetCore.Mvc.NewtonsoftJson" Version="8.0.20" />
+    <PackageVersion Include="Microsoft.AspNetCore.Mvc.Testing" Version="8.0.20" />
+    <PackageVersion Include="Microsoft.AspNetCore.Mvc.Versioning" Version="5.1.0" />
+    <PackageVersion Include="Microsoft.AspNetCore.Mvc.Versioning.ApiExplorer" Version="5.1.0" />
+    <PackageVersion Include="Microsoft.AspNetCore.TestHost" Version="8.0.20" />
+    <PackageVersion Include="Microsoft.CodeAnalysis.PublicApiAnalyzers" Version="4.14.0" />
+    <PackageVersion Include="Microsoft.OpenApi" Version="2.3.2" />
+    <PackageVersion Include="Microsoft.OpenApi.YamlReader" Version="2.3.2" />
```
martincostello:
```suggestion
    <PackageVersion Include="Microsoft.OpenApi" Version="2.3.4" />
    <PackageVersion Include="Microsoft.OpenApi.YamlReader" Version="2.3.4" />
```

Directory.Packages.props
```diff
@@ -1,31 +1,31 @@
...
+    <PackageVersion Include="JunitXml.TestLogger" Version="6.1.0" />
+    <PackageVersion Include="MartinCostello.Logging.XUnit.v3" Version="0.6.0" />
+    <PackageVersion Include="Microsoft.AspNetCore.Mvc.NewtonsoftJson" Version="8.0.21" />
+    <PackageVersion Include="Microsoft.AspNetCore.Mvc.Testing" Version="8.0.21" />
+    <PackageVersion Include="Microsoft.AspNetCore.Mvc.Versioning" Version="5.1.0" />
+    <PackageVersion Include="Microsoft.AspNetCore.Mvc.Versioning.ApiExplorer" Version="5.1.0" />
+    <PackageVersion Include="Microsoft.AspNetCore.TestHost" Version="8.0.21" />
+    <PackageVersion Include="Microsoft.CodeAnalysis.PublicApiAnalyzers" Version="4.14.0" />
+    <PackageVersion Include="Microsoft.OpenApi" Version="2.3.5" />
+    <PackageVersion Include="Microsoft.OpenApi.YamlReader" Version="2.3.5" />
```
martincostello:
```suggestion
    <PackageVersion Include="Microsoft.OpenApi" Version="2.3.6" />
    <PackageVersion Include="Microsoft.OpenApi.YamlReader" Version="2.3.6" />
```

Directory.Packages.props
```diff
@@ -1,31 +1,31 @@
...
+    <PackageVersion Include="JunitXml.TestLogger" Version="6.1.0" />
+    <PackageVersion Include="MartinCostello.Logging.XUnit.v3" Version="0.6.0" />
+    <PackageVersion Include="Microsoft.AspNetCore.Mvc.NewtonsoftJson" Version="8.0.21" />
+    <PackageVersion Include="Microsoft.AspNetCore.Mvc.Testing" Version="8.0.21" />
+    <PackageVersion Include="Microsoft.AspNetCore.Mvc.Versioning" Version="5.1.0" />
+    <PackageVersion Include="Microsoft.AspNetCore.Mvc.Versioning.ApiExplorer" Version="5.1.0" />
+    <PackageVersion Include="Microsoft.AspNetCore.TestHost" Version="8.0.21" />
+    <PackageVersion Include="Microsoft.CodeAnalysis.PublicApiAnalyzers" Version="4.14.0" />
+    <PackageVersion Include="Microsoft.OpenApi" Version="2.3.6" />
+    <PackageVersion Include="Microsoft.OpenApi.YamlReader" Version="2.3.6" />
```
martincostello:
```suggestion
    <PackageVersion Include="Microsoft.OpenApi" Version="2.3.7" />
    <PackageVersion Include="Microsoft.OpenApi.YamlReader" Version="2.3.7" />
```

Directory.Packages.props
```diff
@@ -1,31 +1,31 @@
...
+    <PackageVersion Include="JunitXml.TestLogger" Version="7.0.2" />
+    <PackageVersion Include="MartinCostello.Logging.XUnit.v3" Version="0.6.0" />
+    <PackageVersion Include="Microsoft.AspNetCore.Mvc.NewtonsoftJson" Version="8.0.21" />
+    <PackageVersion Include="Microsoft.AspNetCore.Mvc.Testing" Version="8.0.21" />
+    <PackageVersion Include="Microsoft.AspNetCore.Mvc.Versioning" Version="5.1.0" />
+    <PackageVersion Include="Microsoft.AspNetCore.Mvc.Versioning.ApiExplorer" Version="5.1.0" />
+    <PackageVersion Include="Microsoft.AspNetCore.TestHost" Version="8.0.21" />
+    <PackageVersion Include="Microsoft.CodeAnalysis.PublicApiAnalyzers" Version="4.14.0" />
+    <PackageVersion Include="Microsoft.OpenApi" Version="2.3.7" />
+    <PackageVersion Include="Microsoft.OpenApi.YamlReader" Version="2.3.7" />
```
martincostello:
```suggestion
    <PackageVersion Include="Microsoft.OpenApi" Version="2.3.8" />
    <PackageVersion Include="Microsoft.OpenApi.YamlReader" Version="2.3.8" />
```

Directory.Packages.props
```diff
@@ -1,31 +1,31 @@
...
+    <PackageVersion Include="JunitXml.TestLogger" Version="7.0.2" />
+    <PackageVersion Include="MartinCostello.Logging.XUnit.v3" Version="0.6.0" />
+    <PackageVersion Include="Microsoft.AspNetCore.Mvc.NewtonsoftJson" Version="8.0.21" />
+    <PackageVersion Include="Microsoft.AspNetCore.Mvc.Testing" Version="8.0.21" />
+    <PackageVersion Include="Microsoft.AspNetCore.Mvc.Versioning" Version="5.1.0" />
+    <PackageVersion Include="Microsoft.AspNetCore.Mvc.Versioning.ApiExplorer" Version="5.1.0" />
+    <PackageVersion Include="Microsoft.AspNetCore.TestHost" Version="8.0.21" />
+    <PackageVersion Include="Microsoft.CodeAnalysis.PublicApiAnalyzers" Version="4.14.0" />
+    <PackageVersion Include="Microsoft.OpenApi" Version="2.3.8" />
+    <PackageVersion Include="Microsoft.OpenApi.YamlReader" Version="2.3.8" />
```
martincostello:
```suggestion
    <PackageVersion Include="Microsoft.OpenApi" Version="2.3.9" />
    <PackageVersion Include="Microsoft.OpenApi.YamlReader" Version="2.3.9" />
```

frankyjquintero:
Hi, team 

I’ve been reviewing the recent [PR #3283](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/pull/3283)
 and would like to confirm how the proposed changes impact existing setups that define JWT authentication through AddSecurityDefinition and AddSecurityRequirement.

Currently, our Swagger configuration looks like this:

````
services.AddSwaggerGen(options =>
                {
                    options.EnableAnnotations();
                    options.AddSecurityDefinition("bearerAuth", new OpenApiSecurityScheme
                    {
                        Type = SecuritySchemeType.Http,
                        Scheme = "bearer",
                        BearerFormat = "JWT",
                        Description = "Encabezado de autorización JWT usando el esquema Bearer."
                    });
                    options.AddSecurityRequirement(new OpenApiSecurityRequirement
                    {
                    {
                        new OpenApiSecurityScheme
                        {
                            Reference = new OpenApiReference { Type = ReferenceType.SecurityScheme, Id = "bearerAuth" }
                        },
                        Array.Empty<string>()
                    }
                    });

                    var xmlFilename = $"{Assembly.GetEntryAssembly()?.GetName().Name}.xml";
                    var xmlPath = Path.Combine(AppContext.BaseDirectory, xmlFilename);
                    if (File.Exists(xmlPath))
                    {
                        options.IncludeXmlComments(xmlPath, includeControllerXmlComments: true);
                    }
                });


````


I’d like to understand:

Will this pattern remain valid after the integration of the new OpenAPI features in .NET 10?

Are there any planned breaking changes or deprecations around AddSecurityDefinition or AddSecurityRequirement?

Should we expect OpenApiReference / OpenApiSecurityRequirement to continue being the correct way to define security references, or will the library move toward new abstractions aligned with Microsoft.AspNetCore.OpenApi?

Thank you for your time and for maintaining such a key library for the .NET ecosystem.

martincostello:
@frankyjquintero The changes should be covered by [these updates to the docs](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/16a393d824c2f694533304648e249bee264c9307/docs/configure-and-customize-swaggergen.md#add-security-definitions-and-requirements) and [this migration guide](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/16a393d824c2f694533304648e249bee264c9307/docs/migrating-to-v10.md) (which are both included in this PR).

I have however noticed that the changes for `AddSecurityRequirement()` aren't called out in the migration guide summary, so I'll add that today.

PR #3309, opened 2025-03-17
Fix DescribeAllParametersInCamelCase usage in GenerateOpenApiOperationFromMetadataAsync

### LINKED ISSUE #3271: [Bug]: GenerateOpenApiOperationFromMetadataAsync Does not respect DescribeAllParametersInCamelCase
### Describe the bug

If I call `DescribeAllParametersInCamelCase()` on my `SwaggerGenOptions`, then I expect my Query Parameters to be `CamelCase` instead of `PascalCase`.

This seems to work correctly when they are generated using `SwaggerGenerator.GenerateParametersAsync` in `GenerateOperationAsync` but when the result from `GenerateOpenApiOperationFromMetadataAsync` is used then `DescribeAllParametersInCamelCase` seems to have no effect.

I tried to patch it using a `IParameterFilter` but when i change the name of the parameter, then the associated type(`components\schemas`) is no longer generated.

So the unit test `GetSwagger_SupportsOption_DescribeAllParametersInCamelCase` does test the `SwaggerGenerator.GenerateParametersAsync` part of the code, but totaly misses `GenerateOpenApiOperationFromMetadataAsync`

### Expected behavior

`DescribeAllParametersInCamelCase` should effect Parameters even when the source is `Metadata`

### Actual behavior

Parameters are in PascalCase.

### Steps to reproduce

_No response_

### Exception(s) (if any)

None

7.2.0

### .NET Version

8.0

maksim-sovkov:
# Pull Request

Fixes #3271 

The correct parameter name was generated in the `GenerateParameterAndContext` method, but was not used further.

Added assignment of a new parameter name to the `GenerateOpenApiOperationFromMetadataAsync` method.
Added a unit test.

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
@@ -409,6 +409,7 @@ namespace Swashbuckle.AspNetCore.SwaggerGen
                 if (apiParameter is not null)
                 {
                     var (parameterAndContext, filterContext) = GenerateParameterAndContext(apiParameter, schemaRepository);
+                    parameter.Name = parameterAndContext.Name;
                     parameter.Schema = parameterAndContext.Schema;
                     parameter.Description ??= parameterAndContext.Description;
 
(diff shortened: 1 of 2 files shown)


martincostello:
Thanks!

PR #3317, opened 2025-03-18
Improve release automation

martincostello:
- Move script to bump the version from inline PowerShell to its own script.
- Copy [script from Polly](https://github.com/App-vNext/Polly/blob/1b7defac50fa00395c3e70e303ec111d4370fabe/eng/update-baselines.ps1) to update Public API baselines.
- Fix incorrectly sorted Public API document.
- Remove obsolete Public API baselines for `net6.0`.

### DIFF
--- a/.github/bump-version.ps1
+++ b/.github/bump-version.ps1
@@ -1,0 +1,55 @@
+#! /usr/bin/env pwsh
+param(
+    [Parameter(Mandatory = $false)][string] $NextVersion
+)
+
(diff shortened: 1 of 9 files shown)


martincostello:
Tested with #3318 (see https://github.com/domaindrivendev/Swashbuckle.AspNetCore/pull/3318/commits/174d23525a90d7f2f2b10173019f377966470c0d).

PR #3364, opened 2025-04-11
Update to fix Lists/Arrays of nullables not getting marked as nullable

### LINKED ISSUE #3284: How do I get arrays of nullable types to be represented with a nullable flag in "items" in swagger
### What are you wanting to achieve?

I have a project which has a model in particular that contains arrays of nullable doubles.
When I generate the swagger for the project (using release 7.2.0), the arrays are marked as nullable, but the array items are not.

I have been reading through the issues that were resolved as part of release 7.1.0, and as far as I can tell the issue I'm experiencing should have been resolved as part of that release.

I must be missing something, but not sure what.

### What code or approach do you have so far?

For example, this is a field in my project:
`public double?[] Field1 { get; set; }`
I would expect this to be represented like this in the swagger output:
```
"Field1": {
  "type": "array",
  "items": {
    "type": "number",
    "format": "double",
    "nullable": true  
  },
  "nullable": true
}
```
But instead it is represented like this:
```
"Field1": {
  "type": "array",
  "items": {
    "type": "number",
    "format": "double"
  },
  "nullable": true
}
```

### Additional context

_No response_

Scarecrow7250:
# Pull Request

This Pull Request should hopefully address an issue where Arrays/Lists of Nullable types where not getting correctly marked as Nullable in the generated swagger.

Resolves #3284

In GenerateSchema -> GenerateSchemaForType I added a check to see if the Array underlying type was nullable. If so, it would mark the corresponding section in the schema (in this case items) with nullable = true.

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/SchemaGenerator.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/SchemaGenerator.cs
@@ -183,6 +183,8 @@ public class SchemaGenerator(
         if (schema.Reference == null)
         {
             ApplyFilters(schema, modelType, schemaRepository);
+            if (Nullable.GetUnderlyingType(modelType) != null)
+                schema.Nullable = true;
         }
 
         return schema;
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/SchemaGenerator.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/SchemaGenerator.cs
@@ -184,7 +184,9 @@ public class SchemaGenerator(
         {
             ApplyFilters(schema, modelType, schemaRepository);
             if (Nullable.GetUnderlyingType(modelType) != null)
+            {
                 schema.Nullable = true;
+            }
         }
 
         return schema;
--- a/test/Swashbuckle.AspNetCore.SwaggerGen.Test/SchemaGenerator/JsonSerializerSchemaGeneratorTests.cs
+++ b/test/Swashbuckle.AspNetCore.SwaggerGen.Test/SchemaGenerator/JsonSerializerSchemaGeneratorTests.cs
@@ -161,13 +161,13 @@ public class JsonSerializerSchemaGeneratorTests
         Assert.Equal(schema.AdditionalProperties.Reference.Id, referenceSchema.Reference.Id); // ref to self
     }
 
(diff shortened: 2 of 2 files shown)


src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/SchemaGenerator.cs
```diff
@@ -183,6 +183,8 @@ private OpenApiSchema GenerateSchemaForType(Type modelType, SchemaRepository sch
         if (schema.Reference == null)
         {
             ApplyFilters(schema, modelType, schemaRepository);
+            if (Nullable.GetUnderlyingType(modelType) != null)
+                schema.Nullable = true;
```
martincostello, line 187:
```suggestion
            if (Nullable.GetUnderlyingType(modelType) != null)
            {
                schema.Nullable = true;
            }
```

test/Swashbuckle.AspNetCore.SwaggerGen.Test/SchemaGenerator/JsonSerializerSchemaGeneratorTests.cs
```diff
@@ -161,28 +161,30 @@ public void GenerateSchema_GeneratesReferencedDictionarySchema_IfDictionaryTypeI
         Assert.Equal(schema.AdditionalProperties.Reference.Id, referenceSchema.Reference.Id); // ref to self
     }
 
-    public static TheoryData<Type, JsonSchemaType, string> EnumerableTypeData => new()
+    public static TheoryData<Type, JsonSchemaType, string, bool> EnumerableTypeData => new()
```
martincostello, line 164: There should be some test cases where we expect `true` (e.g. for `int?[]`).

martincostello (commented):
Just some minor comments.

test/Swashbuckle.AspNetCore.SwaggerGen.Test/SchemaGenerator/JsonSerializerSchemaGeneratorTests.cs
Scarecrow7250: Yep, I'll pop in a couple more scenarios

Scarecrow7250: I've added some more test cases.
All unit tests are passing

martincostello (approved):
Thanks!

Issue #3371, opened 2025-04-15
[Feature request]: swagger cli verify no changes


ItsVeryWindy:
### Is your feature request related to a specific problem? Or an existing feature?

On build we generate an open api yaml document using something similar to `dotnet swagger tofile  --yaml --output xyz.yml`. But there are no guarantees that people will either remember or choose not to commit those changes to that file.

### Describe the solution you'd like

Other tools like dotnet format have a --verify-no-changes option, where instead of making the changes it verifies that the changes have been applied.

In this case I would expect the option to, instead of outputting a file, it would compare the expected output with what's already there, and error, if they are different.

martincostello:
Sounds reasonable when comparing the tool to other linters, but if it were me I would use a [Verify](https://github.com/VerifyTests/Verify) test to check the OpenAPI specification matches a baseline (in fact, that's _exactly_ what I do for my own applications: [example](https://github.com/martincostello/api/blob/e7c71efc50075ff01dcb71200e9e90fc714566ec/tests/API.Tests/Integration/OpenApiTests.cs#L13-L28)).

We'd be willing to accept a contribution to add support for this, but given there's other ways to achieve the same goal I don't think it's something we would prioritise doing ourselves.

PR #3375, opened 2025-04-18
add --verify-no-changes option

ItsVeryWindy:
# Pull Request

Fixes https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/3371

This adds the --verify-no-changes option to the tofile command.

When --verify-no-changes is added, --output is required, and is used as an input instead of an output, when not specified, this will return exit code 1.

If the output file does not exist or is different from what has already been output this will return exit code 2.

For this purpose, I have introduced a new class called the `VerifyingWriter`, as the file is being "written out", the contents being "written" are compared with the file specified by --output.

### DIFF
--- a/src/Swashbuckle.AspNetCore.Cli/Program.cs
+++ b/src/Swashbuckle.AspNetCore.Cli/Program.cs
@@ -40,6 +40,7 @@ internal class Program
             c.Option("--basepath", "a specific basePath to include in the Swagger output");
             c.Option(OpenApiVersionOption, "output Swagger in the specified version, defaults to 3.0");
             c.Option("--yaml", "exports swagger in a yaml format", true);
+            c.Option("--verify-no-changes", "verifies that the output requires no changes", true);
 
             // TODO Remove this option in the major version that adds support for OpenAPI 3.1
             c.Option(SerializeAsV2Flag, "output Swagger in the V2 format rather than V3 [deprecated]", true);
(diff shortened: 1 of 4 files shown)


martincostello:
Thanks for your PR. Without digging into the specifics of the code too much, these are my initial thoughts.

> When --verify-no-changes is added, --output ... is used as an input instead of an output

I do not find this intuitive. I think it would be better to add a new mode `verify` which accepts most of the same arguments as `tofile`, but has a mandatory argument to specify the document to verify (`--path`, `--input`, `--document`, `--baseline`, something like that). Otherwise we have an argument that has the opposite meaning based on another argument, and that doesn't `tofile` anything as it produces nothing in the case the file is determined to be equal to the verified document.

> I have introduced a new class called the `VerifyingWriter`

What was the rationale behind implementing things in this specific way?

Would a simpler option be to just read the document for comparison into one `MemoryStream`, generate the document into another, and then use [`SequenceEqual()`](https://learn.microsoft.com/dotnet/api/system.memoryextensions.sequenceequal) to compare them?

```csharp
MemoryStream baseline = ...;
MemoryStream generated = ...;

if (baseline.Length != generated.Length ||
    !baseline.ToArray().SequenceEqual(generated.ToArray()))
{
    // Not equal
}
```

ItsVeryWindy:
The rationale was that if there's a difference, it can be picked up sooner and that you don't have to buffer as much as you are comparing in chunks. Running a few benchmarks with different approaches, the difference between them all is imperceptible. So there's nothing much to be gained from one approach over another, so I don't mind changing that.

As far as having it as a separate commands goes, the bulk of the logic is the same, so I would probably have one common method for setting up both and use a flag to switch between the two.

martincostello:
If you've benchmarked things and it's rather here nor there, I think I'd prefer the simpler `SequenceEqual()` over a custom writer.

We can refactor the logic for `tofile` to make it easier to share the logic between the two commands.

Issue #3225, opened 2025-01-20
[Bug]: required nullable enum should include null


havarnov:
### Describe the bug

When adding including a nullable enum it should be added to the schema with `nullable: true` and `null` should be included in the options, see https://swagger.io/docs/specification/v3_0/data-models/enums/#nullable-enums.

### Expected behavior

An nullable enum should be added to the openapi schema on the following format:

```yaml
type: string

nullable: true  # <---

enum:

  - asc

  - desc

  - null        # <--- without quotes, i.e. null not "null"
```

### Actual behavior

This is how it's added now

```yaml
type: string

enum:

  - asc

  - desc
```

### Steps to reproduce

```cs
using Microsoft.AspNetCore.Mvc;

var builder = WebApplication.CreateBuilder(args);

builder.Services.AddControllers();

builder.Services.AddEndpointsApiExplorer();
builder.Services.AddSwaggerGen();

var app = builder.Build();
app.UseSwagger();
app.UseSwaggerUI();
app.Run();

[ApiController]
public class IndexController : ControllerBase
{
    [HttpGet("/index")]
    [ProducesResponseType<IndexResponseModel>(StatusCodes.Status200OK)]
    public IActionResult Index()
    {
        return Ok(new IndexResponseModel()
        {
            Timestamp = null,
            State = null,
        });
    }
}

public class IndexResponseModel
{
    public required DateTimeOffset? Timestamp { get; init; }
    public State? State { get; init; }
}

public enum State
{
    On = 1,
    Off = 2,
}
```

### Swashbuckle.AspNetCore version

7.2.0

### .NET Version

9.0.101

PR #3377, opened 2025-04-20
fix schema for nullable enums

ItsVeryWindy:
# Pull Request

Fixes https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/3225

This makes changes to the data contract resolvers so that the nullability of enums comes through with the type information.

From there we can ensure that we can set nullable to true and append null as a possible value.

As the values are different there are two different schemas that get built, one which contains the null value, and one that doesn't.

This does mean however that other sections of code now also need to handle nullable types.

Anywhere that does checks if a type is an enum.
Checking for properties that are nullable but have either `Required`, `JsonRequired` or `JsonProperty` with required set.
Checking for required parameters.

### DIFF
--- a/test/Swashbuckle.AspNetCore.Newtonsoft.Test/SchemaGenerator/NewtonsoftSchemaGeneratorTests.cs
+++ b/test/Swashbuckle.AspNetCore.Newtonsoft.Test/SchemaGenerator/NewtonsoftSchemaGeneratorTests.cs
@@ -287,35 +287,28 @@ public class NewtonsoftSchemaGeneratorTests
         Assert.Equal(expectedNullable, schema.Properties[propertyName].Nullable);
     }
 
-    [Theory]
-    [InlineData(typeof(TypeWithNullableProperties), nameof(TypeWithNullableProperties.NullableIntEnumProperty), false)]
-    public void GenerateSchema_DoesNotSetNullableFlag_IfReferencedEnum(
-        Type declaringType,
-        string propertyName,
-        bool expectedNullable)
+    [Fact]
+    public void GenerateSchema_DoesNotSetNullableFlag_IfReferencedEnum()
     {
         var schemaRepository = new SchemaRepository();
 
-        var referenceSchema = Subject().GenerateSchema(declaringType, schemaRepository);
+        var referenceSchema = Subject().GenerateSchema(typeof(TypeWithNullableProperties), schemaRepository);
 
         var schema = schemaRepository.Schemas[referenceSchema.Reference.Id];
-        Assert.Equal(expectedNullable, schema.Properties[propertyName].Nullable);
+        const string propertyName = nameof(TypeWithNullableProperties.NullableIntEnumProperty);
+        Assert.False(schema.Properties[propertyName].Nullable);
         Assert.Equal("IntEnumNullable", schema.Properties[propertyName].Reference.Id);
     }
 
-    [Theory]
-    [InlineData(typeof(TypeWithNullableProperties), nameof(TypeWithNullableProperties.NullableIntEnumProperty), true)]
-    public void GenerateSchema_SetNullableFlag_IfInlineEnum(
-        Type declaringType,
-        string propertyName,
-        bool expectedNullable)
+    [Fact]
+    public void GenerateSchema_SetNullableFlag_IfInlineEnum()
     {
         var schemaRepository = new SchemaRepository();
 
-        var referenceSchema = Subject(o => o.UseInlineDefinitionsForEnums = true).GenerateSchema(declaringType, schemaRepository);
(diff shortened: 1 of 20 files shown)


test/Swashbuckle.AspNetCore.Newtonsoft.Test/SchemaGenerator/NewtonsoftSchemaGeneratorTests.cs
```diff
@@ -287,6 +287,37 @@ public void GenerateSchema_SetsNullableFlag_IfPropertyIsReferenceOrNullableType(
         Assert.Equal(expectedNullable, schema.Properties[propertyName].Nullable);
     }
 
+    [Theory]
+    [InlineData(typeof(TypeWithNullableProperties), nameof(TypeWithNullableProperties.NullableIntEnumProperty), false)]
```
martincostello, line 291: Does this need to be a `[Theory]` if there's only one test case? Same for the test below.

ItsVeryWindy: Not in particular, originally I had started changing the test above it but I wanted to add an extra assertion for the schema reference id.

martincostello: Yeah I figured it was probably an artifact from some refactoring - let's just make them both a `[Fact]`.

martincostello (approved):

martincostello:
Thanks!

ErikPilsits-RJW:
FYI this PR constitutes a breaking change for generated clients, ie nswag. In our typescript client the enum types we were importing don't exist anymore (our response models only had nullable enum properties). This represents a large effort to fix so we will have to hold off any version upgrades until we can find the capacity to do it. This change should have been reserved for a major version update, not a patch version.

martincostello:
https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/3425#issuecomment-2913653741

KirkMunroSagent:
@martincostello: I don't think this PR was implemented properly. I'm still digging into the issue, but since updating to the latest swashbuckle I'm now getting schema errors. We use a unit test that validates the OpenApi spec of our schema, and that unit test fails with 8.1.2, but it was working fine with 8.1.1.

What seems to be happening now is that nullable enums result in a nullable type being added to the schema like this:

```json
 "System.Nullable`1[[MyApp.V1.RoleTypes, MyApp, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null]]": {
   "enum": [
      "Consumer",
      "Administrator",
      "Enterprise",
      "SystemAdministrator",
      null
   ],
   "type": "string",
   "nullable": true
 },
```

That type is now referenced in the OpenApi schema like this:

```json
 {
   "name": "roleType",
   "in": "query",
   "schema": {
     "allOf": [
       {
         "$ref": "#/components/schemas/System.Nullable`1[[MyApp.V1.RoleTypes, MyApp, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null]]"
       }
     ]
   }
 },
```

That won't pass OpenApi validation because the schema type name contains invalid characters.

When I looked back at the issue logged behind this PR, I wondered why that was even logged as an issue. Shouldn't nullable enums be defined in the schema like this?

```json
 "MyApp.V1.RoleTypes": {
   "enum": [
      "Consumer",
      "Administrator",
      "Enterprise",
      "SystemAdministrator"
   ],
   "type": "string"
 },
```

And then, when marking them as nullable, that would be done in the OpenApi schema like this:

```json
 {
   "name": "roleType",
   "in": "query",
   "schema": {
     "allOf": [
       {
         "$ref": "#/components/schemas/MyApp.V1.RoleTypes"
       }
     ],
     "nullable": true
   }
 },
```

Anyway, regardless of what the intent was here, it broke existing code, and I believe the implementation is either missing something that would help the schema be properly generated or it really wasn't an issue at all to begin with, and it was incorrectly logged as a bug.

If this is unclear or confusing, I apologize, I'm still trying to piece together what broke and what should be fixed before I roll back to the previous version.

Thoughts?

martincostello:
The broken type names looks the same as this issue that was reported: https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/3424

I don't have access to a computer for several days, so there will be no new releases until later this week at the earliest.

At this stage it's likely there'll be a 8.1.3 "soon" that reverts these changes until we can resolve the original issues they were intending to fix without causing issues for consumers.

Anyone hitting these problems should just roll back to 8.1.1.

martincostello:
This change was reverted as part of the 8.1.3 release, which is now available from NuGet.org.

PR #3378, opened 2025-04-21
fix minimal api json options not being respected

### LINKED ISSUE #2293: Enum string converter is not respected using .NET 6 minimal API
Using Swashbuckle.AspNetCore 6.2.3, I have configured JSON serializer options per the [docs](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/minimal-apis?view=aspnetcore-6.0#customize-json-binding). Based on the Swashbuckle docs, I expected the enum string converter

```
services.Configure<JsonOptions>(options =>
{
    options.SerializerOptions.Converters.Add(new JsonStringEnumConverter());
    options.SerializerOptions.PropertyNamingPolicy = JsonNamingPolicy.CamelCase;
    options.SerializerOptions.ReferenceHandler = ReferenceHandler.Preserve;
});
``` 

However, Swagger UI still reports enums as ints unless I add also use AddControllers().AddJsonOptions():

```
services.AddControllers()
	.AddJsonOptions(options =>
	{
            options.SerializerOptions.Converters.Add(new JsonStringEnumConverter());
            options.SerializerOptions.PropertyNamingPolicy = JsonNamingPolicy.CamelCase;
            options.SerializerOptions.ReferenceHandler = ReferenceHandler.Preserve;	
        });
```

Once the second block is added, the swagger UI output references the enum string values as expected. As far as I know, the latter simply does the former under the hood, but there seems to be some order of operations affecting the schema generation here?

ItsVeryWindy:
# Pull Request

Fixes https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/2293

This changes how it detects which JsonOptions is used. Both sets of JsonOptions default to the same value, so if none are configured it won't make any difference. This attempts to detect which one is in use based on which one has been configured.

In addition I've added some methods to force it one way or the other.

One possible issue with this approach is if for example a library had to configure json options they may opt to do both for broader support and then you're back to not know which one.

As an alternative approach, we could try and detect based on the `ApiDescription`, if it's from a controller, use mvc options, or use the other one. This is also not without problems, as if someone used the same types between mvc and minimal apis, depending on the order they're processed in you'd possibly get different results if they are configured differently.

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,6 +115,19 @@ In versions of Swashbuckle.AspNetCore prior to `5.0.0`, Swashbuckle would genera
 on the behavior of the [Newtonsoft.Json serializer][newtonsoft-json]. This made sense because that was the serializer that shipped with ASP.NET Core
 at the time. However, since ASP.NET Core 3.0, ASP.NET Core introduces a new serializer, [System.Text.Json (STJ)][system-text-json] out-of-the-box.
 
+If you find that the *STJ* options/attributes are not being honored, this may be because you are using a combination of minimal apis and mvc which have separate json options.
+To force the swagger generation to use either set of json options you can use one of the following methods.
+
+```csharp
+services.AddSwaggerGen(c =>
+{
+    c.SwaggerDoc("v1", new OpenApiInfo { Title = "My API", Version = "v1" });
+});
+services.AddSwaggerGenMinimalApisJsonOptions();
+// or ...
+services.AddSwaggerGenMvcJsonOptions();
+```
+
 If you want to use Newtonsoft.Json instead, you need to install a separate package and explicitly opt-in. By default Swashbuckle.AspNetCore will assume
 you're using the System.Text.Json serializer and generate Schemas based on its behavior. If you're using Newtonsoft.Json then you'll need to install a
 separate Swashbuckle package, [Swashbuckle.AspNetCore.Newtonsoft][swashbuckle-aspnetcore-newtonsoft] to explicitly opt-in.
--- a/README.md
+++ b/README.md
@@ -115,14 +115,15 @@ In versions of Swashbuckle.AspNetCore prior to `5.0.0`, Swashbuckle would genera
 on the behavior of the [Newtonsoft.Json serializer][newtonsoft-json]. This made sense because that was the serializer that shipped with ASP.NET Core
 at the time. However, since ASP.NET Core 3.0, ASP.NET Core introduces a new serializer, [System.Text.Json (STJ)][system-text-json] out-of-the-box.
 
-If you find that the *STJ* options/attributes are not being honored, this may be because you are using a combination of minimal apis and mvc which have separate json options.
-To force the swagger generation to use either set of json options you can use one of the following methods.
+If you find that the *STJ* options/attributes are not being honored, this may be because you are using a combination of Minimal APIs and MVC, which have separate JSON options.
+To force the OpenAPI document generation to use either set of JSON options you can use one of the following methods:
 
 ```csharp
 services.AddSwaggerGen(c =>
 {
     c.SwaggerDoc("v1", new OpenApiInfo { Title = "My API", Version = "v1" });
 });
+// Then either:
 services.AddSwaggerGenMinimalApisJsonOptions();
 // or ...
 services.AddSwaggerGenMvcJsonOptions();
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/ConfigureMinimalApiSwaggerGenJsonOptions.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/ConfigureMinimalApiSwaggerGenJsonOptions.cs
@@ -1,0 +1,12 @@
+﻿using Microsoft.AspNetCore.Http.Json;
+using Microsoft.Extensions.Options;
+
+namespace Swashbuckle.AspNetCore.SwaggerGen.DependencyInjection;
+
+internal class ConfigureMinimalApiSwaggerGenJsonOptions(IOptions<JsonOptions> jsonOptions) : IConfigureOptions<SwaggerGenJsonOptions>
+{
+    public void Configure(SwaggerGenJsonOptions options)
+    {
+        options.SerializerOptions = jsonOptions.Value.SerializerOptions;
+    }
+}
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/ConfigureMinimalApiSwaggerGenJsonOptions.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/ConfigureMinimalApiSwaggerGenJsonOptions.cs
@@ -3,7 +3,7 @@ using Microsoft.Extensions.Options;
 
 namespace Swashbuckle.AspNetCore.SwaggerGen.DependencyInjection;
 
-internal class ConfigureMinimalApiSwaggerGenJsonOptions(IOptions<JsonOptions> jsonOptions) : IConfigureOptions<SwaggerGenJsonOptions>
+internal sealed class ConfigureMinimalApiSwaggerGenJsonOptions(IOptions<JsonOptions> jsonOptions) : IConfigureOptions<SwaggerGenJsonOptions>
 {
     public void Configure(SwaggerGenJsonOptions options)
     {
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/ConfigureMvcSwaggerGenJsonOptions.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/ConfigureMvcSwaggerGenJsonOptions.cs
@@ -1,0 +1,12 @@
+﻿using Microsoft.AspNetCore.Mvc;
+using Microsoft.Extensions.Options;
+
+namespace Swashbuckle.AspNetCore.SwaggerGen.DependencyInjection;
+
+internal class ConfigureMvcSwaggerGenJsonOptions(IOptions<JsonOptions> jsonOptions) : IConfigureOptions<SwaggerGenJsonOptions>
+{
+    public void Configure(SwaggerGenJsonOptions options)
+    {
+        options.SerializerOptions = jsonOptions.Value.JsonSerializerOptions;
+    }
+}
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/ConfigureMvcSwaggerGenJsonOptions.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/ConfigureMvcSwaggerGenJsonOptions.cs
@@ -3,7 +3,7 @@ using Microsoft.Extensions.Options;
 
 namespace Swashbuckle.AspNetCore.SwaggerGen.DependencyInjection;
 
-internal class ConfigureMvcSwaggerGenJsonOptions(IOptions<JsonOptions> jsonOptions) : IConfigureOptions<SwaggerGenJsonOptions>
+internal sealed class ConfigureMvcSwaggerGenJsonOptions(IOptions<JsonOptions> jsonOptions) : IConfigureOptions<SwaggerGenJsonOptions>
 {
     public void Configure(SwaggerGenJsonOptions options)
     {
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/ConfigureSwaggerGenJsonOptions.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/ConfigureSwaggerGenJsonOptions.cs
@@ -1,0 +1,59 @@
+﻿using System.Text.Json;
+using Microsoft.Extensions.Options;
+
+namespace Swashbuckle.AspNetCore.SwaggerGen.DependencyInjection;
+
+internal class ConfigureSwaggerGenJsonOptions : IPostConfigureOptions<SwaggerGenJsonOptions>
+{
+    private readonly IEnumerable<IConfigureOptions<Microsoft.AspNetCore.Http.Json.JsonOptions>> _minimalApiConfigureOptions;
+    private readonly IEnumerable<IPostConfigureOptions<Microsoft.AspNetCore.Http.Json.JsonOptions>> _minimalApiPostConfigureOptions;
+    private readonly Microsoft.AspNetCore.Http.Json.JsonOptions _minimalApiJsonOptions;
+    private readonly Microsoft.AspNetCore.Mvc.JsonOptions _mvcJsonOptions;
+
+    public ConfigureSwaggerGenJsonOptions(
+        IEnumerable<IConfigureOptions<Microsoft.AspNetCore.Http.Json.JsonOptions>> minimalApiConfigureOptions,
+        IEnumerable<IPostConfigureOptions<Microsoft.AspNetCore.Http.Json.JsonOptions>> minimalApiPostConfigureOptions,
+        IOptions<Microsoft.AspNetCore.Http.Json.JsonOptions> minimalApiJsonOptions,
+        IOptions<Microsoft.AspNetCore.Mvc.JsonOptions> mvcJsonOptions
+    )
+    {
+        _minimalApiConfigureOptions = minimalApiConfigureOptions;
+        _minimalApiPostConfigureOptions = minimalApiPostConfigureOptions;
+        _minimalApiJsonOptions = minimalApiJsonOptions.Value;
+        _mvcJsonOptions = mvcJsonOptions.Value;
+    }
+
+    public void PostConfigure(string name, SwaggerGenJsonOptions options)
+    {
+        if (options.SerializerOptions != null)
+        {
+            return;
+        }
+
+        /*
+         * There is no surefire way to do this.
+         * However, both JsonOptions are defaulted in the same way.
+         * If neither is configured it makes no difference which one is chosen.
+         * If both are configured, then we just need to make a choice.
+         * As Minimal APIs are newer if someone is configuring them
+         * it's probably more likely that is what they're using.
+         * 
+         * If either JsonOptions is null we will try to create a new instance as
+         * a last resort as this is an expensive operation.
+         */
+
+        var serializerOptions = _mvcJsonOptions.JsonSerializerOptions ?? JsonSerializerOptions.Default;
+
+        if (HasConfiguredMinimalApiJsonOptions())
+        {
+            serializerOptions = _minimalApiJsonOptions.SerializerOptions ?? serializerOptions;
+        }
+
+        options.SerializerOptions = serializerOptions;
+    }
+
+    private bool HasConfiguredMinimalApiJsonOptions()
+    {
+        return _minimalApiConfigureOptions.Any() || _minimalApiPostConfigureOptions.Any();
+    }
+}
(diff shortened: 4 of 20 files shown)


src/Swashbuckle.AspNetCore.SwaggerGen/PublicAPI/net8.0/PublicAPI.Shipped.txt
martincostello, line 1: These changes go in the `Unshipped` files - they're moved by automation when the changes are shipped to NuGet.org.

src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenServiceCollectionExtensions.cs
```diff
@@ -70,21 +94,40 @@ private JsonSerializerOptions ResolveOptions()
...
+                _serviceProvider.GetService<IOptions<AspNetCore.Http.Json.JsonOptions>>()?.Value?.SerializerOptions
                 ?? JsonSerializerOptions.Default;
+
+            if (HasConfiguredMinimalApiJsonOptions())
+            {
+                serializerOptions ??= _serviceProvider.GetService<IOptions<AspNetCore.Http.Json.JsonOptions>>()?.Value?.SerializerOptions;
+            }
 #else
             serializerOptions = new JsonSerializerOptions();
 #endif
```
martincostello, line 118: I wonder if a better thing to do would be:

1. Add a new first-class object (just-like the `JsonOptions`) to hold Swashbuckle's own `JsonSerializerOptions`. Users can then specifically register that if they wish to.
2. We add a `IPostConfigureOptions<T>` for that type that applies the above heuristics if the options in that class are null at the point of being resolved.
3. Then, the new overloads just provide a bridge from one `JsonOptions` to that new type. Then we don't need to replace services.

src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenServiceCollectionExtensions.cs
```diff
@@ -44,6 +44,30 @@ public static IServiceCollection AddSwaggerGen(
...
+    {
+        return services.Replace(
+            ServiceDescriptor.Transient<ISerializerDataContractResolver>((s) =>
+            {
+                var options = s.GetRequiredService<IOptionsSnapshot<AspNetCore.Mvc.JsonOptions>>().Value.JsonSerializerOptions;
+
+                return new JsonSerializerDataContractResolver(options);
+            }));
+    }
+#endif
```
martincostello, line 69: These new methods need tests.

src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenServiceCollectionExtensions.cs
ItsVeryWindy: Instead of or in addition to these methods?

martincostello: In addition as syntactic-sugar.

src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/ConfigureMinimalApiSwaggerGenJsonOptions.cs
```diff
@@ -0,0 +1,14 @@
+﻿#if NET
+using Microsoft.AspNetCore.Http.Json;
+using Microsoft.Extensions.Options;
+
+namespace Swashbuckle.AspNetCore.SwaggerGen.DependencyInjection;
+
+internal class ConfigureMinimalApiSwaggerGenJsonOptions(IOptions<JsonOptions> jsonOptions) : IConfigureOptions<SwaggerGenJsonOptions>
```
martincostello, line 7:
```suggestion
internal sealed class ConfigureMinimalApiSwaggerGenJsonOptions(IOptions<JsonOptions> jsonOptions) : IConfigureOptions<SwaggerGenJsonOptions>
```

src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/ConfigureMvcSwaggerGenJsonOptions.cs
```diff
@@ -0,0 +1,14 @@
+﻿#if NET
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.Extensions.Options;
+
+namespace Swashbuckle.AspNetCore.SwaggerGen.DependencyInjection;
+
+internal class ConfigureMvcSwaggerGenJsonOptions(IOptions<JsonOptions> jsonOptions) : IConfigureOptions<SwaggerGenJsonOptions>
```
martincostello, line 7:
```suggestion
internal sealed class ConfigureMvcSwaggerGenJsonOptions(IOptions<JsonOptions> jsonOptions) : IConfigureOptions<SwaggerGenJsonOptions>
```

src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/ConfigureSwaggerGenJsonOptions.cs
```diff
@@ -0,0 +1,67 @@
+﻿using System.Text.Json;
+using Microsoft.Extensions.Options;
+
+namespace Swashbuckle.AspNetCore.SwaggerGen.DependencyInjection;
+
+internal class ConfigureSwaggerGenJsonOptions : IPostConfigureOptions<SwaggerGenJsonOptions>
```
martincostello, line 6:
```suggestion
internal sealed class ConfigureSwaggerGenJsonOptions : IPostConfigureOptions<SwaggerGenJsonOptions>
```

src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/ConfigureSwaggerGenJsonOptions.cs
```diff
@@ -0,0 +1,67 @@
...
+    private readonly IEnumerable<IPostConfigureOptions<Microsoft.AspNetCore.Http.Json.JsonOptions>> _minimalApiPostConfigureOptions;
+    private readonly Microsoft.AspNetCore.Http.Json.JsonOptions _minimalApiJsonOptions;
+    private readonly Microsoft.AspNetCore.Mvc.JsonOptions _mvcJsonOptions;
+
+    public ConfigureSwaggerGenJsonOptions(
+        IEnumerable<IConfigureOptions<Microsoft.AspNetCore.Http.Json.JsonOptions>> minimalApiConfigureOptions,
+        IEnumerable<IPostConfigureOptions<Microsoft.AspNetCore.Http.Json.JsonOptions>> minimalApiPostConfigureOptions,
+        IOptions<Microsoft.AspNetCore.Http.Json.JsonOptions> minimalApiJsonOptions,
+        IOptions<Microsoft.AspNetCore.Mvc.JsonOptions> mvcJsonOptions
+    )
```
martincostello, line 19:
```suggestion
        IOptions<Microsoft.AspNetCore.Mvc.JsonOptions> mvcJsonOptions)
```

src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/ConfigureSwaggerGenJsonOptions.cs
```diff
@@ -0,0 +1,67 @@
...
+         * it's probably more likely that is what they're using.
+         * 
+         * If either JsonOptions is null we will try to create a new instance as
+         * a last resort as this is an expensive operation.
+         */
+
+#if NET
+        var serializerOptions = _mvcJsonOptions.JsonSerializerOptions ?? JsonSerializerOptions.Default;
+
+        if (HasConfiguredMinimalApiJsonOptions())
```
martincostello, line 50:
```suggestion

        if (_minimalApiConfigureOptions.Any() || _minimalApiPostConfigureOptions.Any())
```

src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/ConfigureSwaggerGenJsonOptions.cs
```diff
@@ -0,0 +1,67 @@
...
+        options.SerializerOptions = new JsonSerializerOptions();
+#endif
+    }
+
+#if NET
+    private bool HasConfiguredMinimalApiJsonOptions()
+    {
+        return _minimalApiConfigureOptions.Any() || _minimalApiPostConfigureOptions.Any();
+    }
+#endif
```
martincostello, line 66:
```suggestion
```

src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenJsonOptions.cs
```diff
@@ -0,0 +1,8 @@
+﻿using System.Text.Json;
+
+namespace Swashbuckle.AspNetCore.SwaggerGen.DependencyInjection;
+
+public class SwaggerGenJsonOptions
+{
+    public JsonSerializerOptions SerializerOptions { get; set; }
+}
```
martincostello, line 8: Could you add XML documentation to this type and the property explaining what it's for please?

test/Swashbuckle.AspNetCore.SwaggerGen.Test/SwaggerGenJsonOptionsTests.cs
```diff
@@ -0,0 +1,105 @@
...
+
+    private sealed class DummyHostEnvironment : IWebHostEnvironment
+    {
+        public string WebRootPath { get; set; }
+        public IFileProvider WebRootFileProvider { get; set; }
+        public string ApplicationName { get; set; }
+        public IFileProvider ContentRootFileProvider { get; set; }
+        public string ContentRootPath { get; set; }
+        public string EnvironmentName { get; set; }
+    }
```
martincostello, line 91: I think we have a few copies of this in the tests. If any are in the same project as this one, could you extract out into shared type that can be re-used?

test/Swashbuckle.AspNetCore.SwaggerGen.Test/SwaggerGenJsonOptionsTests.cs
martincostello, line 1: Are there any existing integration tests we can add usage of the new behaviour to as well?

src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenServiceCollectionExtensions.cs
```diff
@@ -44,47 +44,22 @@ public static IServiceCollection AddSwaggerGen(
         return services;
     }
 
-    public static void ConfigureSwaggerGen(
-        this IServiceCollection services,
-        Action<SwaggerGenOptions> setupAction)
+#if NET
+    public static IServiceCollection AddSwaggerGenMinimalApisJsonOptions(this IServiceCollection services)
```
martincostello, line 48: Could you add something to the README showing how to use these two new methods please?

test/Swashbuckle.AspNetCore.SwaggerGen.Test/SwaggerGenJsonOptionsTests.cs
ItsVeryWindy: There's no specific tests for this, maybe I could add it to `MinimalAppWithHostedServices` and `Basic`?

martincostello:
I should be able to review this next week.

README.md
```diff
@@ -112,6 +112,19 @@ In summary ...
 
 If you're using **System.Text.Json (STJ)**, then the setup described above will be sufficient, and *STJ* options/attributes will be automatically honored by the Swagger generator.
 
+If you find that the *STJ* options/attributes are not being honored, this may be because you are using a combination of minimal apis and mvc which have separate json options.
```
martincostello, line 115:
```suggestion
If you find that the *STJ* options/attributes are not being honored, this may be because you are using a combination of Minimal APIs and MVC, which have separate JSON options.
```

README.md
```diff
@@ -112,6 +112,19 @@ In summary ...
...
+To force the swagger generation to use either set of json options you can use one of the following methods.
+
+```
```
martincostello, line 118:
```suggestion
```csharp
```

README.md
```diff
@@ -112,6 +112,19 @@ In summary ...
...
+services.AddSwaggerGen(c =>
+{
+    c.SwaggerDoc("v1", new OpenApiInfo { Title = "My API", Version = "v1" });
+});
+services.AddSwaggerGenMinimalApisJsonOptions();
```
martincostello, line 123:
```suggestion
// Then either:
services.AddSwaggerGenMinimalApisJsonOptions();
```

src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/ConfigureSwaggerGenJsonOptions.cs
```diff
@@ -0,0 +1,59 @@
...
+         * If neither is configured it makes no difference which one is chosen.
+         * If both are configured, then we just need to make a choice.
+         * As Minimal APIs are newer if someone is configuring them
+         * it's probably more likely that is what they're using.
+         * 
+         * If either JsonOptions is null we will try to create a new instance as
+         * a last resort as this is an expensive operation.
+         */
+
+#if NET
```
martincostello, line 46:
```suggestion
#if NET
        /*
         * There is no surefire way to do this.
         * However, both JsonOptions are defaulted in the same way.
         * If neither is configured it makes no difference which one is chosen.
         * If both are configured, then we just need to make a choice.
         * As Minimal APIs are newer if someone is configuring them
         * it's probably more likely that is what they're using.
         * 
         * If either JsonOptions is null we will try to create a new instance as
         * a last resort as this is an expensive operation.
         */
```

src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenJsonOptions.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System.Text.Json;
+
+namespace Swashbuckle.AspNetCore.SwaggerGen.DependencyInjection;
+
+/// <summary>
+/// Configures the <see cref="JsonSerializerOptions"/> to be used by <see cref="JsonSerializerDataContractResolver"/>.
```
martincostello, line 6: We have two implementations of `ISerializerDataContractResolver` (the other being `NewtonsoftDataContractResolver`), so rather than explicitly reference one in code, I would make this less specific.

Maybe something like:

> A class to specify options to use to configure JSON serialization for OpenAPI documents.

test/Swashbuckle.AspNetCore.SwaggerGen.Test/SwaggerGenJsonOptionsTests.cs
```diff
@@ -0,0 +1,95 @@
...
+public class SwaggerGenJsonOptionsTests
+{
+    [Fact]
+    public static void Ensure_SwaggerGenJsonOptions_Uses_MinimalApi_JsonOptions_When_Overridden()
+    {
+        var services = new ServiceCollection();
+        services.AddSingleton<IWebHostEnvironment, DummyHostEnvironment>();
+        services.AddSwaggerGen();
+        services.AddMvcCore().AddJsonOptions(o => o.JsonSerializerOptions.Converters.Add(new DummyConverter()));
+        services.AddSwaggerGenMinimalApisJsonOptions();
```
martincostello, line 20: Could we explicitly configure the options for Minimal APIs to be different and then assert that  `SerializerOptions` is the same instance instead?

The assert right now shows that it isn't the MVC options, but it doesn't prove it's the Minimal APIs one.

test/Swashbuckle.AspNetCore.SwaggerGen.Test/SwaggerGenJsonOptionsTests.cs
```diff
@@ -0,0 +1,95 @@
...
+
+        using var provider = services.BuildServiceProvider();
+
+        var swaggerGenDummyConverter = provider.GetService<IOptions<SwaggerGenJsonOptions>>().Value.SerializerOptions.Converters.FirstOrDefault();
+
+        Assert.Equal(expectedDummyConverter, swaggerGenDummyConverter);
+    }
+
+    [Fact]
+    public static void Ensure_SwaggerGenJsonOptions_Uses_Mvc_JsonOptions_When_Not_Using_Minimal_Apis()
```
martincostello, line 49: Same in this test.

test/Swashbuckle.AspNetCore.SwaggerGen.Test/SwaggerGenJsonOptionsTests.cs
```diff
@@ -0,0 +1,95 @@
...
+using System.Text.Json.Serialization;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Options;
+using Swashbuckle.AspNetCore.SwaggerGen.DependencyInjection;
+using Swashbuckle.AspNetCore.SwaggerGen.Test.Fixtures;
+
+namespace Swashbuckle.AspNetCore.SwaggerGen.Test;
+
+public class SwaggerGenJsonOptionsTests
```
martincostello, line 11: Could we add a test here that checks for the `JsonSerializerOptions.Default` case please? We might have one already elsewhere, which if that's the case it can be moved here instead.

src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenJsonOptions.cs
```diff
@@ -0,0 +1,14 @@
...
+/// </summary>
+public class SwaggerGenJsonOptions
+{
+    /// <summary>
+    /// Gets or Sets the json serializer options used by <see cref="JsonSerializerDataContractResolver"/>.
```
martincostello, line 11:
```suggestion
    /// Gets or sets the JSON serializer options to use. 
```

test/Swashbuckle.AspNetCore.SwaggerGen.Test/SwaggerGenJsonOptionsTests.cs
martincostello: Sure.

test/Swashbuckle.AspNetCore.SwaggerGen.Test/SwaggerGenJsonOptionsTests.cs
ItsVeryWindy: It's not actually possible to hit the `JsonSerializerOptions.Default` as the mvc json options will never be null, and minimal api json options will also never be null due to how the types are instantiated. They also did not work in the original code for the same reason. Given that code can't be executed I would suggest removing the `??` checks.

martincostello:
Holding this after #3428 is merged so we can bundle the new APIs in the same release.

martincostello:
I'm going to close this as it's got a lot of merge conflicts and has gone stale.

If you'd like to continue with this PR, please open a new one and we can start fresh.

Issue #3396, opened 2025-05-01
Drop support for netstandard2.0


martincostello:
The codebase is getting increasingly complicated (e.g. see #3283), and the lack of APIs available to consume with `netstandard2.0` for the sake of supporting ASP.NET Core 2.1/2.3 for .NET Framework is causing a lot of additional complication we could do without (e.g. `#if NET` directives).

There doesn't appear to be any evidence that people are still actively using Swashbuckle.AspNetCore on .NET Framework _and_ have a need for new features.

We don't even test for this scenario end-to-end anywhere, so there's a good chance it doesn't work anyway.

To make the project sustainable, we should consider dropping the `netstandard2.0` targets and instead only target in-support versions of ASP.NET Core instead. For example, today that would be `net8.0` and `net9.0`, with `net10.0` in November 2025.

This would reduce the complexity of the code and make things more sustainable to maintain.

Support can be dropped in a major version ahead of the major version for .NET 10 support.

PR #3399, opened 2025-05-05
Using GZip to compress swagger-ui-dist files in embedded resource to reduce the output dll size

stratosblue:
Using GZip to compress swagger-ui-dist files in embedded resource to reduce the output dll size

# Pull Request

No

Using gzip to compress the `swagger-ui-dist` at build time. And decompress it before response it.
This change can compress the size of the `Swashbuckle.AspNetCore.SwaggerUI.dll` from 2259k to 682k.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
-[Oo]bj/
+﻿[Oo]bj/
 [Bb]in/
+[Tt]est[Rr]esult*/
 .vs/
 .idea*
 BenchmarkDotNet.Artifacts*/
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -1,4 +1,4 @@
-<Project>
+﻿<Project>
   <ItemGroup>
     <PackageVersion Include="Autofac.Extensions.DependencyInjection" Version="10.0.0" />
     <PackageVersion Include="BenchmarkDotNet" Version="0.14.0" />
@@ -29,6 +29,7 @@
     <PackageVersion Include="NSubstitute" Version="5.3.0" />
     <PackageVersion Include="NSwag.MSBuild" Version="14.4.0" />
     <PackageVersion Include="ReportGenerator" Version="5.4.4" />
+    <PackageVersion Include="ResourceCompressor" Version="1.0.1" />
     <PackageVersion Include="System.Text.Json" Version="4.6.0" />
     <PackageVersion Include="Verify.XunitV3" Version="29.3.0" />
     <PackageVersion Include="xunit" Version="2.9.3" />
--- a/src/Swashbuckle.AspNetCore.SwaggerUI/SwaggerUIMiddleware.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerUI/SwaggerUIMiddleware.cs
@@ -4,7 +4,6 @@ using System.Text.RegularExpressions;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.StaticFiles;
-using Microsoft.Extensions.FileProviders;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using System.Reflection;
@@ -110,7 +109,7 @@ internal sealed partial class SwaggerUIMiddleware
         var staticFileOptions = new StaticFileOptions
         {
             RequestPath = string.IsNullOrEmpty(options.RoutePrefix) ? string.Empty : $"/{options.RoutePrefix}",
-            FileProvider = new EmbeddedFileProvider(typeof(SwaggerUIMiddleware).Assembly, EmbeddedFileNamespace),
+            FileProvider = new GZipCompressedEmbeddedFileProvider(typeof(SwaggerUIMiddleware).Assembly, EmbeddedFileNamespace),
             OnPrepareResponse = (context) => SetCacheHeaders(context.Context.Response, options),
         };
 
--- a/src/Swashbuckle.AspNetCore.SwaggerUI/Swashbuckle.AspNetCore.SwaggerUI.csproj
+++ b/src/Swashbuckle.AspNetCore.SwaggerUI/Swashbuckle.AspNetCore.SwaggerUI.csproj
@@ -16,7 +16,11 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <EmbeddedResource Include="node_modules/swagger-ui-dist/**/*" Exclude="**/*/index.html;**/*/*.map;**/*/*.json;**/*/*.md;**/*/swagger-ui-es-*" />
+    <InternalsVisibleTo Include="Swashbuckle.AspNetCore.IntegrationTests" Key="00240000048000009400000006020000002400005253413100040000010001000b80d002f0437a94c4f6d77dfc29ce54df1040f66374cb7122ec5ddab3907a6dfe609a07ffd49b9aa6731fe3f8de4da7979ddea73e949c86effc4bec3f469cc36e788cf58c42c4be755efd1afd996ccab9f0db3455b834219f6614d00fe0410cc307c4eced320fd65fcc501f135dc537d7d09dcc1d2a572d24f8459a86469ec5" />
+  </ItemGroup>
+  
+  <ItemGroup>
+    <CompressedEmbeddedResource Include="node_modules/swagger-ui-dist/**/*" Exclude="**/*/index.html;**/*/*.map;**/*/*.json;**/*/*.md;**/*/swagger-ui-es-*" GeneratedFileSuffix="None" />
     <None Remove="index.html;index.js" />
     <EmbeddedResource Include="index.html;index.js" />
   </ItemGroup>
@@ -34,6 +38,10 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.CodeAnalysis.PublicApiAnalyzers" PrivateAssets="All" />
+    <PackageReference Include="ResourceCompressor">
+      <PrivateAssets>all</PrivateAssets>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+    </PackageReference>
     <AdditionalFiles Include="PublicAPI\PublicAPI.Shipped.txt" />
     <AdditionalFiles Include="PublicAPI\PublicAPI.Unshipped.txt" />
   </ItemGroup>
--- a/test/Swashbuckle.AspNetCore.IntegrationTests/SwaggerUIIntegrationTests.cs
+++ b/test/Swashbuckle.AspNetCore.IntegrationTests/SwaggerUIIntegrationTests.cs
@@ -1,4 +1,6 @@
 ﻿using System.Net;
+using Microsoft.Extensions.FileProviders;
+using Swashbuckle.AspNetCore.SwaggerUI;
 
 namespace Swashbuckle.AspNetCore.IntegrationTests;
 
@@ -21,6 +23,82 @@ public class SwaggerUIIntegrationTests
 
         Assert.Equal(HttpStatusCode.MovedPermanently, response.StatusCode);
         Assert.Equal(expectedRedirectPath, response.Headers.Location.ToString());
+    }
+
+    [Theory]
+    [InlineData("Swashbuckle.AspNetCore.SwaggerUI.node_modules")]
+    [InlineData("Swashbuckle.AspNetCore.SwaggerUI.node_modules.swagger_ui_dist")]
+    public void ResourceRead_CompressedEmbeddedFileProvider(string baseNamespace)
+    {
+        //confirm that GZipCompressedEmbeddedFileProvider is the same as EmbeddedFileProvider
+        var provider = new EmbeddedFileProvider(typeof(SwaggerUIOptions).Assembly, baseNamespace);
+        var compressedProvider = new GZipCompressedEmbeddedFileProvider(typeof(SwaggerUIOptions).Assembly, baseNamespace);
+        var checkSubpaths = new string[]
+        {
+            "/",
+            null,
+            string.Empty,
+            " ",
+            "\t",
+            "\n",
+            "/swagger_ui_dist",
+            "swagger_ui_dist",
+            "/nodir",
+            "nodir"
+        };
+
+        foreach (var subpath in checkSubpaths)
+        {
+            AssertResources(provider, compressedProvider, subpath);
+        }
+
+        var nonExistentFile = Guid.NewGuid().ToString();
+        AssertFileInfo(provider.GetFileInfo(nonExistentFile), compressedProvider.GetFileInfo(nonExistentFile));
+
+        static void AssertResources(IFileProvider expectedProvider, IFileProvider actualProvider, string subpath)
+        {
+            var expectedContents = expectedProvider.GetDirectoryContents(subpath);
+            var actualContents = actualProvider.GetDirectoryContents(subpath);
+
+            Assert.Equal(expectedContents.Exists, actualContents.Exists);
+            Assert.Equal(expectedContents.Count(), actualContents.Count());
+            var actualResourceMap = actualContents.ToDictionary(m => m.Name);
+
+            foreach (var expectedFileInfo in expectedContents)
+            {
+                Assert.True(actualResourceMap.TryGetValue(expectedFileInfo.Name, out var actualFileInfo));
+                Assert.NotNull(actualFileInfo);
+                Assert.True(actualFileInfo.Exists);
+                AssertFileInfo(expectedFileInfo, actualFileInfo);
+                AssertFileInfo(expectedProvider.GetFileInfo(expectedFileInfo.Name), actualProvider.GetFileInfo(expectedFileInfo.Name));
+            }
+        }
+
+        static void AssertFileInfo(IFileInfo expectedFileInfo, IFileInfo actualFileInfo)
+        {
+            Assert.Equal(expectedFileInfo.Exists, actualFileInfo.Exists);
+            Assert.Equal(expectedFileInfo.IsDirectory, actualFileInfo.IsDirectory);
+            Assert.Equal(expectedFileInfo.LastModified, actualFileInfo.LastModified);
+            Assert.Equal(expectedFileInfo.PhysicalPath, actualFileInfo.PhysicalPath);
+
+            if (expectedFileInfo.Exists && !expectedFileInfo.IsDirectory)
+            {
+                Assert.True(actualFileInfo.Length > 0);
+
+                using var stream = actualFileInfo.CreateReadStream();
+                Assert.NotNull(stream);
+                var buffer = new byte[256];
+                var readLength = stream.Read(buffer, 0, buffer.Length);
+                Assert.True(readLength > 0);
+                //we can check for correctness here
+            }
+            else
+            {
+                Assert.Equal(expectedFileInfo.Length, actualFileInfo.Length);
+                Assert.ThrowsAny<FileNotFoundException>(() => expectedFileInfo.CreateReadStream());
+                Assert.ThrowsAny<FileNotFoundException>(() => actualFileInfo.CreateReadStream());
+            }
+        }
     }
 
     [Theory]
--- a/src/Swashbuckle.AspNetCore.SwaggerUI/GZipCompressedEmbeddedFileProvider.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerUI/GZipCompressedEmbeddedFileProvider.cs
@@ -1,0 +1,143 @@
+﻿using System.Buffers;
+using System.Collections;
+using System.Collections.Concurrent;
+using System.IO.Compression;
+using System.Reflection;
+
+using Microsoft.Extensions.FileProviders;
+using Microsoft.Extensions.Primitives;
+
+namespace Swashbuckle.AspNetCore.SwaggerUI;
+
+/// <summary>
+/// a <see cref="EmbeddedFileProvider"/> wrapper to provider gzip decompressed resource file info
+/// </summary>
+/// <param name="assembly"></param>
+/// <param name="baseNamespace"></param>
+internal sealed class GZipCompressedEmbeddedFileProvider(Assembly assembly, string baseNamespace) : IFileProvider
+{
+    private readonly string _baseNamespace = string.IsNullOrEmpty(baseNamespace) ? string.Empty : baseNamespace + ".";
+
+    private readonly EmbeddedFileProvider _embeddedFileProvider = new(assembly, baseNamespace);
+
+    private readonly ConcurrentDictionary<string, long> _subpathLengthCache = new(StringComparer.Ordinal);
+
+    /// <inheritdoc/>
+    public IDirectoryContents GetDirectoryContents(string subpath)
+    {
+        //logic same as EmbeddedFileProvider
+        return subpath?.Length == 0 || string.Equals(subpath, "/", StringComparison.Ordinal)
+               ? new DecompresDirectoryContents(EnumerateItems().ToList())
+               : new DecompresDirectoryContents([]);
+    }
+
+    /// <inheritdoc/>
+    public IFileInfo GetFileInfo(string subpath) => new GZipDecompresFileInfoWrapper(_embeddedFileProvider.GetFileInfo(subpath), subpath, GetSubpathDecompressedLength);
+
+    /// <inheritdoc/>
+    public IChangeToken Watch(string filter) => _embeddedFileProvider.Watch(filter);
+
+    private static string NormalizeSubPath(string subpath) => string.Join(".", subpath.Split(['/', '\\'], StringSplitOptions.RemoveEmptyEntries).Where(m => !string.IsNullOrWhiteSpace(m)));
+
+    /// <summary>
+    /// Get the <paramref name="stream"/>'s length by decompress and read it
+    /// </summary>
+    /// <param name="stream"></param>
+    /// <returns></returns>
+    private static long ReadStreamDecompressedLength(Stream stream)
+    {
+        const int BufferSize = 4096;
+
+        long length = 0;
+        var buffer = ArrayPool<byte>.Shared.Rent(BufferSize);
+
+        try
+        {
+            using var gzipStream = new GZipStream(stream, CompressionMode.Decompress, true);
+            var readLength = 0;
+            do
+            {
+                readLength = gzipStream.Read(buffer, 0, BufferSize);
+                length += readLength;
+            } while (readLength != 0);
+        }
+        finally
+        {
+            ArrayPool<byte>.Shared.Return(buffer);
+        }
+        return length;
+    }
+
+    private IEnumerable<IFileInfo> EnumerateItems()
+    {
+        return assembly.GetManifestResourceNames()
+                       .Where(name => name.StartsWith(_baseNamespace))
+                       .Select(name => name.Substring(_baseNamespace.Length))
+                       .Select(subpath => this.GetFileInfo(subpath))
+                       .Where(static fileInfo => fileInfo.Exists);
+    }
+
+    /// <summary>
+    /// Get <paramref name="subpath"/>'s decompressed length and cache it
+    /// </summary>
+    /// <param name="subpath"></param>
+    /// <returns></returns>
+    private long GetSubpathDecompressedLength(string subpath)
+    {
+        if (_subpathLengthCache.TryGetValue(subpath, out var length))
+        {
+            return length;
+        }
+
+        var resourceName = $"{_baseNamespace}{NormalizeSubPath(subpath)}";
+
+        using var stream = assembly.GetManifestResourceStream(resourceName);
+        if (stream == null)
+        {
+            return -1;
+        }
+
+        return _subpathLengthCache.GetOrAdd(subpath, _ => ReadStreamDecompressedLength(stream));
+    }
+
+    private sealed class DecompresDirectoryContents(List<IFileInfo> fileInfos) : IDirectoryContents
+    {
+        /// <inheritdoc/>
+        public bool Exists => fileInfos.Count > 0;
+
+        /// <inheritdoc/>
+        public IEnumerator<IFileInfo> GetEnumerator() => fileInfos.GetEnumerator();
+
+        /// <inheritdoc/>
+        IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
+    }
+
+    private sealed class GZipDecompresFileInfoWrapper(IFileInfo fileInfo, string subpath, Func<string, long> getSubpathLengthFunc) : IFileInfo
+    {
+        /// <inheritdoc/>
+        public bool Exists => fileInfo.Exists;
+
+        /// <inheritdoc/>
+        public bool IsDirectory => fileInfo.IsDirectory;
+
+        /// <inheritdoc/>
+        public DateTimeOffset LastModified => fileInfo.LastModified;
+
+        /// <inheritdoc/>
+        public long Length => fileInfo.Exists ? getSubpathLengthFunc(subpath) : -1;
+
+        /// <inheritdoc/>
+        public string Name => fileInfo.Name;
+
+        /// <inheritdoc/>
+        public string PhysicalPath => fileInfo.PhysicalPath;
+
+        /// <inheritdoc/>
+        public Stream CreateReadStream()
+        {
+            return fileInfo.CreateReadStream() is { } stream
+                   ? new GZipStream(stream, CompressionMode.Decompress)
+                   : throw new FileNotFoundException(message: null, fileName: fileInfo.Name);
+        }
+    }
+}


martincostello:
Thanks for your PR. I'll take a proper look at this later this week, but initial thoughts:

1. If we're going to do this, we should also do it for [ReDoc](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/tree/master/src/Swashbuckle.AspNetCore.ReDoc).
2. The test you've added isn't really an integration test, it's a unit test. A better test would be some actual integration tests (like the other tests that do HTTP requests to a server) that verify that the contents requested are byte-for-byte identical to the source files on disk. Then the `InternalsVisibleTo` that this PR adds can be removed.

stratosblue:
> Thanks for your PR. I'll take a proper look at this later this week, but initial thoughts:
> 
>     1. If we're going to do this, we should also do it for [ReDoc](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/tree/master/src/Swashbuckle.AspNetCore.ReDoc).
> 
>     2. The test you've added isn't really an integration test, it's a unit test. A better test would be some actual integration tests (like the other tests that do HTTP requests to a server) that verify that the contents requested are byte-for-byte identical to the source files on disk. Then the `InternalsVisibleTo` that this PR adds can be removed.

- [ReDoc](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/tree/master/src/Swashbuckle.AspNetCore.ReDoc) compressed now. The new size is abot 328KB.
- The HTTP requests test added. And `InternalsVisibleTo` removed and there is a test project `Swashbuckle.AspNetCore.SwaggerUI.Test` for `GZipCompressedEmbeddedFileProvider`'s unit test.

martincostello:
I'm not going to be able to come back to this for over a week or so now for outside-world reasons. Sorry about the delay.

However, I have seen the related PR for scalar, and any improvements made there (like using gzip directly) should also be incorporated into SwaggerUI and ReDoc packages here too please. I would suggest waiting until that PR is merged in case there are other suggestions and tweaks to be made.

stratosblue:
I agree with you. We should confirm this PR after that PR merged.

stratosblue:
Hi @martincostello ,

That PR has been merged.

Now I plan to remove the use of `StaticFileMiddleware` from `SwaggerUIMiddleware` and implement a lighter and faster Swagger static resource access logic, including GZip direct response functionality. Can I start doing this?

stratosblue:
> Hi @martincostello ,
> 
> That PR has been merged.
> 
> Now I plan to remove the use of `StaticFileMiddleware` from `SwaggerUIMiddleware` and implement a lighter and faster Swagger static resource access logic, including GZip direct response functionality. Can I start doing this?

@martincostello  Done. Please review it.

martincostello:
I saw - this is in my TODO pile.

test/Swashbuckle.AspNetCore.IntegrationTests/ReDocIntegrationTests.cs
```diff
@@ -32,13 +34,13 @@ public async Task IndexUrl_ReturnsEmbeddedVersionOfTheRedocUI()
 
         AssertResource(htmlResponse);
         AssertResource(cssResponse);
-        AssertResource(jsResponse);
+        AssertResource(jsResponse, false);
```
martincostello, line 37:
```suggestion
        AssertResource(jsResponse, weakETag: false);
```

test/Swashbuckle.AspNetCore.IntegrationTests/ReDocIntegrationTests.cs
```diff
@@ -162,4 +164,43 @@ public void ReDocOptions_Extensions()
...
+        var diskFile = "redoc/bundles/redoc.standalone.js";
+        var diskFileName = Path.GetFileName("redoc/bundles/redoc.standalone.js");
+
+        using var htmlResponse = await client.GetAsync("/Api-Docs/redoc.standalone.js", TestContext.Current.CancellationToken);
+        Assert.Equal(HttpStatusCode.OK, htmlResponse.StatusCode);
+
+        using var stream = await htmlResponse.Content.ReadAsStreamAsync(TestContext.Current.CancellationToken);
+        using var diskFileStream = File.OpenRead(diskFile);
+
+        Assert.Equal(MD5.HashData(diskFileStream), MD5.HashData(stream));
```
martincostello, line 183: Please use SHA1 - it avoids tools raising warnings related to using insecure hash algorithms. Even if this is only test code, it's still noise we can avoid.

test/Swashbuckle.AspNetCore.IntegrationTests/ReDocIntegrationTests.cs
```diff
@@ -162,4 +164,43 @@ public void ReDocOptions_Extensions()
...
+    [Fact]
+    public async Task ReDocMiddleware_Returns_ExpectedAssetContents()
+    {
+        var site = new TestSite(typeof(ReDocApp.Startup));
+        using var client = site.BuildClient();
+
+        var diskFile = "redoc/bundles/redoc.standalone.js";
+        var diskFileName = Path.GetFileName("redoc/bundles/redoc.standalone.js");
+
+        using var htmlResponse = await client.GetAsync("/Api-Docs/redoc.standalone.js", TestContext.Current.CancellationToken);
```
martincostello, line 177: We should assert on the response headers too.

test/Swashbuckle.AspNetCore.IntegrationTests/SwaggerUIIntegrationTests.cs
```diff
@@ -40,19 +42,19 @@ public async Task IndexUrl_ReturnsEmbeddedVersionOfTheSwaggerUI(
         AssertResource(htmlResponse);
 
         using var jsResponse = await client.GetAsync(swaggerUijsPath, TestContext.Current.CancellationToken);
-        AssertResource(jsResponse);
+        AssertResource(jsResponse, false);
 
         using var indexCss = await client.GetAsync(indexCssPath, TestContext.Current.CancellationToken);
-        AssertResource(indexCss);
+        AssertResource(indexCss, false);
```
martincostello, line 48:
```suggestion
        AssertResource(indexCss, weakETag: false);
```

test/Swashbuckle.AspNetCore.IntegrationTests/SwaggerUIIntegrationTests.cs
```diff
@@ -40,19 +42,19 @@ public async Task IndexUrl_ReturnsEmbeddedVersionOfTheSwaggerUI(
...
 
         using var cssResponse = await client.GetAsync(swaggerUiCssPath, TestContext.Current.CancellationToken);
-        AssertResource(cssResponse);
+        AssertResource(cssResponse, false);
```
martincostello, line 51:
```suggestion
        AssertResource(cssResponse, weakETag: false);
```

test/Swashbuckle.AspNetCore.IntegrationTests/SwaggerUIIntegrationTests.cs
```diff
@@ -190,4 +192,71 @@ public async Task IndexUrl_Returns_ExpectedAssetPaths(
         Assert.Contains($"<script src=\"{scriptBundlePath}\" charset=\"utf-8\">", content);
         Assert.Contains($"<script src=\"{scriptPresetsPath}\" charset=\"utf-8\">", content);
     }
+
+    [Fact]
```
martincostello, line 196: Similar comments on these tests too regarding MD5 and response headers.

test/Swashbuckle.AspNetCore.IntegrationTests/Swashbuckle.AspNetCore.IntegrationTests.csproj
```diff
@@ -42,6 +42,15 @@
...
   </ItemGroup>
 
+  <ItemGroup>
+    <None Include="../../src/Swashbuckle.AspNetCore.SwaggerUI/node_modules/swagger-ui-dist/**/*" Exclude="**/*/index.html;**/*/index.js;**/*/*.map;**/*/*.json;**/*/*.md;**/*/swagger-ui-es-*;**/*/LICENSE;**/*/NOTICE" Link="swagger-ui-dist/%(RecursiveDir)%(FileName)%(Extension)">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
+    <None Include="../../src/Swashbuckle.AspNetCore.ReDoc/node_modules/redoc/bundles/redoc.standalone.js" Link="redoc/bundles/redoc.standalone.js">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
+  </ItemGroup>
```
martincostello, line 52: Rather than copy all these files around, I would suggest instead doing something like this:

1. [Embed the full path of the solution into the test assembly](https://github.com/martincostello/build-kit/blob/5e1fbfd37b76d040c7b9c5a99ecaaac1ea43ae03/tests/BuildKit.Tests/MartinCostello.BuildKit.Tests.csproj#L33)
2. [Extract the value from the assembly at runtime in the tests to find the original file on disk directly](https://github.com/martincostello/build-kit/blob/5e1fbfd37b76d040c7b9c5a99ecaaac1ea43ae03/tests/BuildKit.Tests/IntegrationTests.cs#L12-L16)

src/Swashbuckle.AspNetCore.SwaggerUI/CompressedEmbeddedFileResponser.cs
```diff
@@ -0,0 +1,186 @@
...
+using System.Security.Cryptography;
+using Microsoft.Net.Http.Headers;
+
+#if NET
+using System.Collections.Frozen;
+#endif
+
+namespace Swashbuckle.AspNetCore;
+
+internal class CompressedEmbeddedFileResponser
```
martincostello, line 16: `Responder` would be a more appropriate suffix for this class name.

src/Swashbuckle.AspNetCore.SwaggerUI/CompressedEmbeddedFileResponser.cs
```diff
@@ -0,0 +1,186 @@
...
+{
+    private readonly Assembly _assembly;
+    private readonly string _cacheControlHeaderValue;
+    private readonly string _pathPrefix;
+    private readonly FileExtensionContentTypeProvider _contentTypeProvider = new();
+
+#if NET
```
martincostello, line 23: There's a lot of these in this PR, so it's probably a good idea for me to get #3396 done first.

src/Swashbuckle.AspNetCore.SwaggerUI/CompressedEmbeddedFileResponser.cs
```diff
@@ -0,0 +1,186 @@
...
+#else
+            responseHeaders["Content-Type"] = contentType;
+            responseHeaders["ETag"] = etag;
+            responseHeaders["Cache-Control"] = _cacheControlHeaderValue;
+#endif
+            using var stream = OpenResourceStream(resourceIndexCache);
+            if (responseWithGZip)
+            {
+                responseHeaders.ContentLength = stream.Length;
+                await stream.CopyToAsync(httpContext.Response.Body, 81320, httpContext.RequestAborted);
```
martincostello, line 101: What's the significance of the `81320` magic number?

src/Swashbuckle.AspNetCore.ReDoc/Swashbuckle.AspNetCore.ReDoc.csproj
```diff
@@ -50,6 +54,11 @@
     <AdditionalFiles Include="PublicAPI\$(_TargetFrameworkIdentifier)\PublicAPI.Shipped.txt" />
     <AdditionalFiles Include="PublicAPI\$(_TargetFrameworkIdentifier)\PublicAPI.Unshipped.txt" />
   </ItemGroup>
+  
+  <ItemGroup>
+    <Compile Include="..\Swashbuckle.AspNetCore.SwaggerUI\CompressedEmbeddedFileResponser.cs" Link="CompressedEmbeddedFileResponser.cs" />
+    <Compile Include="..\Swashbuckle.AspNetCore.SwaggerUI\HttpContextAcceptEncodingCheckExtensions.cs" Link="HttpContextAcceptEncodingCheckExtensions.cs" />
```
martincostello, line 60: It would be better to put these into the [Shared](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/tree/master/src/Shared) folder and consume as shared source in both projects.

src/Swashbuckle.AspNetCore.SwaggerUI/HttpContextAcceptEncodingCheckExtensions.cs
```diff
@@ -0,0 +1,64 @@
...
+        return IsGZipAccepted(httpContext.Request.Headers.AcceptEncoding);
+#else
+        return IsGZipAccepted(httpContext.Request.Headers["Accept-Encoding"]);
+#endif
+    }
+
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    private static bool IsGZipAccepted(string? acceptEncoding) => !string.IsNullOrWhiteSpace(acceptEncoding) && s_gzipAcceptedCheckRegex.IsMatch(acceptEncoding);
+
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
martincostello, line 41: Do we _need_ these annotations?

Directory.Packages.props
```diff
@@ -29,6 +29,7 @@
     <PackageVersion Include="NSubstitute" Version="5.3.0" />
     <PackageVersion Include="NSwag.MSBuild" Version="14.4.0" />
     <PackageVersion Include="ReportGenerator" Version="5.4.4" />
+    <PackageVersion Include="ResourceCompressor" Version="1.0.1" />
```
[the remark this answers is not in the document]
stratosblue: > I wonder, is it possible to instead leverage the new [Static asset delivery optimization](https://learn.microsoft.com/en-us/aspnet/core/release-notes/aspnetcore-9.0?view=aspnetcore-9.0#static-asset-delivery-optimization) functionality added in the .NET 9 SDK to achieve the same aim of compressing the source files before we embed them into the assemblies? There seem to be [a bunch of things](https://github.com/dotnet/sdk/blob/3aa9493112a9f4da6c3c88398a7f8e5190ceb80e/src/StaticWebAssetsSdk/Targets/Microsoft.NET.Sdk.StaticWebAssets.Compression.targets) in the SDK for compression that handle gzip.

That seems feasible, but it may require a lot of experimentation. I'm not sure how it works at TFM below net9.0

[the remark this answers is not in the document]
martincostello: Me either - but as we don't need the runtime components if we're just using MSBuild targets/tasks, then it might be feasible.

src/Swashbuckle.AspNetCore.SwaggerUI/CompressedEmbeddedFileResponser.cs
stratosblue: > What's the significance of the `81320` magic number?

Typo of `bufferSize` default value `81920`

src/Swashbuckle.AspNetCore.SwaggerUI/HttpContextAcceptEncodingCheckExtensions.cs
stratosblue: There may be a slight performance improvement, but removing it will not have any impact

src/Swashbuckle.AspNetCore.SwaggerUI/CompressedEmbeddedFileResponser.cs
martincostello: Is there any special significance to the value itself?

stratosblue: > Is there any special significance to the value itself?

None. it's just because netstandard2.0 doesn't have the method `Stream.CopyToAsync(Stream destination, CancellationToken cancellationToken)` and then it got typed incorrectly

stratosblue:
All done except for directly using SDK foundation to compress file

src/Swashbuckle.AspNetCore.SwaggerUI/CompressedEmbeddedFileResponser.cs
martincostello: FYI: #3422

.gitignore
```diff
@@ -1,3 +1,4 @@
+[Tt]est[Rr]esult*/
```
martincostello: Curious why this is needed when we have line 20? If we do need this, then move this to replace that pattern.

test/Swashbuckle.AspNetCore.IntegrationTests/Swashbuckle.AspNetCore.IntegrationTests.csproj
```diff
@@ -24,6 +24,16 @@
...
   </ItemGroup>
 
+  <ItemGroup>
+    <EmbeddedResource Include="../../src/Swashbuckle.AspNetCore.SwaggerUI/node_modules/swagger-ui-dist/**/*" Exclude="**/*/index.html;**/*/index.js;**/*/*.map;**/*/*.json;**/*/*.md;**/*/swagger-ui-es-*;**/*/LICENSE;**/*/NOTICE" Link="Embedded/SwaggerUI/%(RecursiveDir)%(FileName)%(Extension)" />
+    <EmbeddedResource Include="../../src/Swashbuckle.AspNetCore.ReDoc/node_modules/redoc/bundles/redoc.standalone.js" Link="Embedded/ReDoc/%(RecursiveDir)%(FileName)%(Extension)" />
+  </ItemGroup>
+  
+  <ItemGroup Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net8.0'))">
+    <PackageReference Update="Microsoft.AspNetCore.Mvc.Testing" VersionOverride="8.0.14" />
+    <PackageReference Update="Microsoft.AspNetCore.TestHost" VersionOverride="8.0.14" />
```
martincostello:
```suggestion
    <PackageReference Update="Microsoft.AspNetCore.Mvc.Testing" VersionOverride="8.0.17" />
    <PackageReference Update="Microsoft.AspNetCore.TestHost" VersionOverride="8.0.17" />
```

test/Swashbuckle.AspNetCore.IntegrationTests/SwaggerUIIntegrationTests.cs
```diff
@@ -190,4 +192,79 @@ public async Task IndexUrl_Returns_ExpectedAssetPaths(
...
+        }
+    }
+
+    private IEnumerable<(string ResourceName, string FileName)> EnumerateEmbeddedUIFiles()
+    {
+        const string ResourcePrefix = "Swashbuckle.AspNetCore.IntegrationTests.Embedded.SwaggerUI.";
+        return typeof(SwaggerUIIntegrationTests).Assembly.GetManifestResourceNames()
+                                                         .Where(name => name.StartsWith(ResourcePrefix))
+                                                         .Select(name => (name, name.Substring(ResourcePrefix.Length)));
+    }
```
martincostello: Have this return a list. Otherwise in a number of the tests everything is enumerated twice.

```suggestion
    private static List<(string ResourceName, string FileName)> GetEmbeddedUIFiles()
    {
        const string ResourcePrefix = "Swashbuckle.AspNetCore.IntegrationTests.Embedded.SwaggerUI.";
        return typeof(SwaggerUIIntegrationTests).Assembly
            .GetManifestResourceNames()
            .Where(name => name.StartsWith(ResourcePrefix))
            .Select(name => (name, name.Substring(ResourcePrefix.Length)))
            .ToList();
    }
```

test/Swashbuckle.AspNetCore.IntegrationTests/SwaggerUIIntegrationTests.cs
```diff
@@ -190,4 +192,79 @@ public async Task IndexUrl_Returns_ExpectedAssetPaths(
...
+
+        foreach (var (_, fileName) in EnumerateEmbeddedUIFiles())
+        {
+            using var htmlResponse = await client.GetAsync(fileName, TestContext.Current.CancellationToken);
+            Assert.Equal(HttpStatusCode.OK, htmlResponse.StatusCode);
+            Assert.NotNull(htmlResponse.Headers.ETag?.Tag);
+
+            using var requestMessage = new HttpRequestMessage(HttpMethod.Get, fileName);
+            requestMessage.Headers.IfNoneMatch.Add(new(htmlResponse.Headers.ETag?.Tag));
+            using var secondHtmlResponse = await client.SendAsync(requestMessage, TestContext.Current.CancellationToken);
```
martincostello, line 265:
```suggestion

            using var secondHtmlResponse = await client.SendAsync(requestMessage, TestContext.Current.CancellationToken);
```

test/Swashbuckle.AspNetCore.IntegrationTests/SwaggerUIIntegrationTests.cs
```diff
@@ -190,4 +192,79 @@ public async Task IndexUrl_Returns_ExpectedAssetPaths(
...
+        var site = new TestSite(typeof(Basic.Startup), outputHelper);
+        using var client = site.BuildClient();
+
+        Assert.NotEmpty(EnumerateEmbeddedUIFiles());
+
+        foreach (var (resourceName, fileName) in EnumerateEmbeddedUIFiles())
+        {
+            using var requestMessage = new HttpRequestMessage(HttpMethod.Get, fileName);
+            requestMessage.Headers.AcceptEncoding.Add(new("gzip"));
+            using var htmlResponse = await client.SendAsync(requestMessage, TestContext.Current.CancellationToken);
```
martincostello, line 233:
```suggestion

            using var htmlResponse = await client.SendAsync(requestMessage, TestContext.Current.CancellationToken);

```

test/Swashbuckle.AspNetCore.IntegrationTests/SwaggerUIIntegrationTests.cs
```diff
@@ -190,4 +192,79 @@ public async Task IndexUrl_Returns_ExpectedAssetPaths(
...
+        foreach (var (resourceName, fileName) in EnumerateEmbeddedUIFiles())
+        {
+            using var requestMessage = new HttpRequestMessage(HttpMethod.Get, fileName);
+            using var htmlResponse = await client.SendAsync(requestMessage, TestContext.Current.CancellationToken);
+            Assert.Equal(HttpStatusCode.OK, htmlResponse.StatusCode);
+
+            using var stream = await htmlResponse.Content.ReadAsStreamAsync(TestContext.Current.CancellationToken);
+            using var diskFileStream = typeof(SwaggerUIIntegrationTests).Assembly.GetManifestResourceStream(resourceName);
+
+            Assert.Equal(SHA1.HashData(diskFileStream), SHA1.HashData(stream));
```
martincostello, line 215: For completeness, let's also assert that the streams aren't empty before hashing.

src/Swashbuckle.AspNetCore.SwaggerUI/Swashbuckle.AspNetCore.SwaggerUI.csproj
```diff
@@ -67,4 +70,27 @@
     </ItemGroup>
   </Target>
 
+  <!--embed compressed file using sdk task-->
```
martincostello:
```suggestion
  <!-- Embed compressed file using SDK task-->
```

src/Swashbuckle.AspNetCore.SwaggerUI/SwaggerUIMiddleware.cs
```diff
@@ -20,24 +14,26 @@ internal sealed partial class SwaggerUIMiddleware
...
         _options = options ?? new SwaggerUIOptions();
 
-        _staticFileMiddleware = CreateStaticFileMiddleware(next, hostingEnv, loggerFactory, options);
-
         if (options.JsonSerializerOptions != null)
         {
             _jsonSerializerOptions = options.JsonSerializerOptions;
         }
+
+        var pathPrefix = options.RoutePrefix.StartsWith("/") ? options.RoutePrefix : $"/{options.RoutePrefix}";
```
martincostello: Can this use the `char` overload?

```suggestion
        var pathPrefix = options.RoutePrefix.StartsWith('/') ? options.RoutePrefix : $"/{options.RoutePrefix}";
```

test/Swashbuckle.AspNetCore.IntegrationTests/ReDocIntegrationTests.cs
```diff
@@ -162,4 +164,38 @@ public void ReDocOptions_Extensions()
...
+        var site = new TestSite(typeof(ReDocApp.Startup), outputHelper);
+        using var client = site.BuildClient();
+
+        using var htmlResponse = await client.GetAsync("/Api-Docs/redoc.standalone.js", TestContext.Current.CancellationToken);
+        Assert.Equal(HttpStatusCode.OK, htmlResponse.StatusCode);
+
+        using var stream = await htmlResponse.Content.ReadAsStreamAsync(TestContext.Current.CancellationToken);
+        using var rawFileStream = typeof(ReDocIntegrationTests).Assembly.GetManifestResourceStream("Swashbuckle.AspNetCore.IntegrationTests.Embedded.ReDoc.redoc.standalone.js");
+
+        Assert.Equal(SHA1.HashData(rawFileStream), SHA1.HashData(stream));
```
martincostello, line 181: Same here regarding empty.

test/Swashbuckle.AspNetCore.IntegrationTests/ReDocIntegrationTests.cs
```diff
@@ -162,4 +164,38 @@ public void ReDocOptions_Extensions()
...
+
+    [Fact]
+    public async Task ReDocMiddleware_Returns_ExpectedAssetContents_GZipDirectly()
+    {
+        var site = new TestSite(typeof(ReDocApp.Startup), outputHelper);
+        using var client = site.BuildClient();
+
+        using var requestMessage = new HttpRequestMessage(HttpMethod.Get, "/Api-Docs/redoc.standalone.js");
+        requestMessage.Headers.AcceptEncoding.Add(new("gzip"));
+        using var htmlResponse = await client.SendAsync(requestMessage, TestContext.Current.CancellationToken);
```
martincostello, line 193:
```suggestion

        using var htmlResponse = await client.SendAsync(requestMessage, TestContext.Current.CancellationToken);

```

src/Swashbuckle.AspNetCore.ReDoc/Swashbuckle.AspNetCore.ReDoc.csproj
```diff
@@ -73,4 +77,27 @@
     </ItemGroup>
   </Target>
 
+  <!--embed compressed file using sdk task-->
```
martincostello, line 80:
```suggestion
  <!-- Embed compressed file using SDK task -->
```

src/Swashbuckle.AspNetCore.ReDoc/ReDocMiddleware.cs
```diff
@@ -20,24 +14,26 @@ internal sealed class ReDocMiddleware
...
         _options = options ?? new ReDocOptions();
 
-        _staticFileMiddleware = CreateStaticFileMiddleware(next, hostingEnv, loggerFactory, options);
-
         if (options.JsonSerializerOptions != null)
         {
             _jsonSerializerOptions = options.JsonSerializerOptions;
         }
+
+        var pathPrefix = options.RoutePrefix.StartsWith("/") ? options.RoutePrefix : $"/{options.RoutePrefix}";
```
martincostello: Same question here regarding `char`.

src/Shared/CompressedEmbeddedFileResponder.cs
```diff
@@ -0,0 +1,160 @@
+﻿#nullable enable
+
+using Microsoft.AspNetCore.Http;
+using System.Reflection;
+using Microsoft.AspNetCore.StaticFiles;
+using System.IO.Compression;
+using System.Security.Cryptography;
+using Microsoft.Net.Http.Headers;
+using System.Collections.Frozen;
```
martincostello: Sort usings.

src/Shared/CompressedEmbeddedFileResponder.cs
```diff
@@ -0,0 +1,160 @@
...
+    private readonly FrozenDictionary<string, ResourceIndexCache> _resourceIndexes;
+
+    public CompressedEmbeddedFileResponder(Assembly assembly, string resourceNamePrefix, string pathPrefix, TimeSpan? cacheLifetime)
+    {
+        _assembly = assembly ?? throw new ArgumentNullException(nameof(assembly));
+        _pathPrefix = string.IsNullOrWhiteSpace(pathPrefix)
+                      ? string.Empty
+                      : pathPrefix == "/"
+                        ? string.Empty
+                        : pathPrefix.TrimEnd('/');
```
martincostello: This feels overcomplicated using ternary.

src/Shared/CompressedEmbeddedFileResponder.cs
```diff
@@ -0,0 +1,160 @@
...
+        _cacheControlHeaderValue = GetCacheControlHeaderValue(cacheLifetime);
+
+        var resourceIndexes = assembly.GetManifestResourceNames()
```
martincostello: `resourceMap`

src/Shared/CompressedEmbeddedFileResponder.cs
```diff
@@ -0,0 +1,160 @@
...
+            }
+
+            var responseWithGZip = httpContext.IsGZipAccepted();
+            if (responseWithGZip)
+            {
+                responseHeaders.ContentEncoding = "gzip";
+            }
+
+            responseHeaders.ContentType = contentType;
+            responseHeaders.ETag = etag;
```
martincostello: As this is a `EntityTagHeaderValue`, rather than convert back and forth to strings can we just deal in that type?

src/Shared/CompressedEmbeddedFileResponder.cs
```diff
@@ -0,0 +1,160 @@
...
+            responseHeaders.CacheControl = _cacheControlHeaderValue;
+
+            using var stream = OpenResourceStream(resourceIndexCache);
+            if (responseWithGZip)
+            {
+                responseHeaders.ContentLength = stream.Length;
+                await stream.CopyToAsync(httpContext.Response.Body, 81920, httpContext.RequestAborted);
```
martincostello: Put `81920` into a named constant please.

src/Shared/CompressedEmbeddedFileResponder.cs
```diff
@@ -0,0 +1,160 @@
...
+
+            var responseWithGZip = httpContext.IsGZipAccepted();
+            if (responseWithGZip)
+            {
+                responseHeaders.ContentEncoding = "gzip";
+            }
+
+            responseHeaders.ContentType = contentType;
+            responseHeaders.ETag = etag;
+            responseHeaders.CacheControl = _cacheControlHeaderValue;
```
martincostello: Same here with `CacheControlHeaderValue`.

src/Shared/CompressedEmbeddedFileResponder.cs
```diff
@@ -0,0 +1,160 @@
...
+    {
+        return resourceIndexCache.ContentType
+               ?? (_contentTypeProvider.TryGetContentType(resourceIndexCache.ResourceName, out var contentTypeValue)
+                   ? contentTypeValue
+                   : "application/octet-stream");
+    }
+
+    private (string ETag, long DecompressContentLength) GetDecompressContentETag(ResourceIndexCache resourceIndexCache)
+    {
+        //Get decompress content and hash
```
martincostello:
```suggestion
```

src/Shared/CompressedEmbeddedFileResponder.cs
```diff
@@ -0,0 +1,160 @@
...
+
+        using var memoryStream = new MemoryStream((int)stream.Length * 2);
+        using var gzipStream = new GZipStream(stream, CompressionMode.Decompress);
+        gzipStream.CopyTo(memoryStream);
+        memoryStream.Seek(0, SeekOrigin.Begin);
+
+        resourceIndexCache.DecompressContentLength = memoryStream.Length;
+
+        using var sha1 = SHA1.Create();
+        var hashData = sha1.ComputeHash(memoryStream);
```
martincostello:
```suggestion
        var hashData = SHA1.HashData(memoryStream);
```

src/Shared/CompressedEmbeddedFileResponder.cs
```diff
@@ -0,0 +1,160 @@
...
+        resourceIndexCache.ETag = $"\"{Convert.ToBase64String(hashData)}\"";
+
+        return (resourceIndexCache.ETag, resourceIndexCache.DecompressContentLength.Value);
+    }
+
+    private Stream OpenResourceStream(ResourceIndexCache resourceIndexCache)
+    {
+        // Actually, since the name comes from GetManifestResourceNames(), the content can definitely be obtained
+        return _assembly.GetManifestResourceStream(resourceIndexCache.ResourceName)
+               ?? throw new InvalidOperationException($"Can not read resource \"{resourceIndexCache.ResourceName}\" in {_assembly}");
```
martincostello: This can just use `!` and not throw as we know we've already validated it exists.

src/Shared/CompressedEmbeddedFileResponder.cs
```diff
@@ -0,0 +1,160 @@
...
+    {
+        var path = httpContext.Request.Path.Value?.ToString() ?? string.Empty;
+        if (!path.StartsWith(_pathPrefix, StringComparison.OrdinalIgnoreCase))
+        {
+            return false;
+        }
+
+        path = path.Substring(_pathPrefix.Length).Replace('/', '.');
+
+        if (_resourceIndexes.TryGetValue(path, out var resourceIndexCache))
```
martincostello: Invert this condition, then we can unindent by one:

```csharp
if (!_resourceIndexes.TryGetValue(path, out var resourceIndexCache))
{
    return false;
}

// The code

return true;
```

src/Shared/CompressedEmbeddedFileResponder.cs
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    private Stream OpenResourceStream(ResourceIndexCache resourceIndexCache)
+    {
+        // Actually, since the name comes from GetManifestResourceNames(), the content can definitely be obtained
+        return _assembly.GetManifestResourceStream(resourceIndexCache.ResourceName)
+               ?? throw new InvalidOperationException($"Can not read resource \"{resourceIndexCache.ResourceName}\" in {_assembly}");
+    }
+
+    private class ResourceIndexCache(string resourceName)
```
martincostello:
```suggestion
    private sealed class ResourceIndexCache(string resourceName)
```

src/Shared/CompressedEmbeddedFileResponder.cs
```diff
@@ -0,0 +1,160 @@
...
+    {
+        public string? ETag { get; set; }
+
+        public string? ContentType { get; set; }
+
+        public long? DecompressContentLength { get; set; }
+
+        public string ResourceName { get; } = resourceName ?? throw new ArgumentNullException(nameof(resourceName));
```
martincostello: Shouldn't need to throw here as its not nullable and we should have already validated it isn't.

src/Shared/HttpContextAcceptEncodingCheckExtensions.cs
```diff
@@ -0,0 +1,54 @@
...
+/// <summary>
+/// <see cref="HttpContext"/> Accept-Encoding check extensions
+/// </summary>
+internal static partial class HttpContextAcceptEncodingCheckExtensions
+{
+    private static readonly Regex s_gzipAcceptedCheckRegex = GetGZipAcceptedCheckRegex();
+
+    [GeneratedRegex(@"(^|,)\s*gzip\s*(;|,|$)", RegexOptions.IgnoreCase | RegexOptions.CultureInvariant)]
+    private static partial Regex GetGZipAcceptedCheckRegex();
+
```
martincostello:
```suggestion
```

src/Shared/HttpContextAcceptEncodingCheckExtensions.cs
```diff
@@ -0,0 +1,54 @@
...
+using System.Text.RegularExpressions;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Primitives;
+
+namespace Swashbuckle.AspNetCore;
+
+/// <summary>
+/// <see cref="HttpContext"/> Accept-Encoding check extensions
+/// </summary>
```
martincostello:
```suggestion
```

src/Shared/HttpContextAcceptEncodingCheckExtensions.cs
```diff
@@ -0,0 +1,54 @@
...
+internal static partial class HttpContextAcceptEncodingCheckExtensions
+{
+    private static readonly Regex s_gzipAcceptedCheckRegex = GetGZipAcceptedCheckRegex();
+
+    [GeneratedRegex(@"(^|,)\s*gzip\s*(;|,|$)", RegexOptions.IgnoreCase | RegexOptions.CultureInvariant)]
+    private static partial Regex GetGZipAcceptedCheckRegex();
```
martincostello: Is there a way to do this without using Regex?

src/Shared/HttpContextAcceptEncodingCheckExtensions.cs
```diff
@@ -0,0 +1,54 @@
...
+namespace Swashbuckle.AspNetCore;
+
+/// <summary>
+/// <see cref="HttpContext"/> Accept-Encoding check extensions
+/// </summary>
+internal static partial class HttpContextAcceptEncodingCheckExtensions
+{
+    private static readonly Regex s_gzipAcceptedCheckRegex = GetGZipAcceptedCheckRegex();
+
+    [GeneratedRegex(@"(^|,)\s*gzip\s*(;|,|$)", RegexOptions.IgnoreCase | RegexOptions.CultureInvariant)]
```
martincostello: This should be invariant if we keep it.

src/Shared/HttpContextAcceptEncodingCheckExtensions.cs
```diff
@@ -0,0 +1,54 @@
...
+    private static partial Regex GetGZipAcceptedCheckRegex();
+
+
+    /// <summary>
+    /// Check is the <paramref name="httpContext"/> support gzip response
+    /// </summary>
+    /// <param name="httpContext"></param>
+    /// <returns>is gzip response accepted</returns>
```
martincostello:
```suggestion
```

martincostello (approved):
Thanks for contributing this improvement. There's a few minor things I'm going to fix post-merge, but otherwise this looks good.

Issue #3425, opened 2025-05-23
[Bug]: Behaviour change in 8.1.2 for nullable enums


jadu-1337:
### Describe the bug

After updating from 8.1.1 to 8.1.2, there is an issue with type names for nullable enums.

It appears that all enums that are used as nullable values, e.g. `class Request { MyEnum? Value { get; set; } }` have their schema name changed in the generated resulting OpenAPI spec from `MyEnum` to `MyEnumNullable`.

This breaks client code generation that relies on the type names to match C# source, e.g. when using NSwagGen with `GenerateDtoTypes:false`, the generated client code fails to compile because it references a non-existing `MyEnumNullable`.

### Expected behavior

Schema names should match C# source names

### Actual behavior

Schema names are altered.

### Swashbuckle.AspNetCore version

8.1.2

### .NET Version

9.0.5

ItsVeryWindy:
So I've played around with a few different generators, I tried nswag, open api generator and swagger code gen for generating c# code.

In general support seems to vary with what structure works best, with most providing mixed results.

I tried two different approaches whilst still trying to keep the same intent, one which involves using oneOf
```json
"schema": {
              "oneOf": [
                {
                  "$ref": "#/components/schemas/LogLevel"
                },
                {
                  "enum": [
                    null
                  ]
                }
              ],
```
and another which was suggested in the open api faqs (LogLevel in this instance would include null by default).

```json
"schema": {
              "allOf": [
                {
                  "$ref": "#/components/schemas/LogLevel"
                },
                {
                  "not" : {
                    "enum": [
                      null
                    ]
                  }
                }
              ],
```

At least for nswag, it was able to generate the expected result when using oneOf, open api generator tried to use the options pattern which I guess is usable but not the best, and swagger code gen came out with garbage.

Using "not" approach in all cases generated either subpar or unusable code.

There are many generators out there, and it's very time consuming to test each an every one. But I wouldn't be surprised if most of them would suck at dealing with this.

I'd maybe suggest going with the first option and adding a flag to allow it to be turned on/off just to allow it to play nicer with code generators.

mrange:
Hi. Thanks for looking into it. I understand that this might be the new behaviour in the future but then it should, as you said, it should be a new major version number.

Thanks again.

martincostello:
It wasn't a major version bump because it was an unintentional change in behaviour as already stated.

oyzar:
[...]

PR #3428, opened 2025-05-27
use one of behind a flag instead for nullable enums

ItsVeryWindy:
# Pull Request

This is to fix https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/3425.

This changes how the nullable enums work so that instead of creating a reference to a separate type, it uses one of with a reference and a null enum. This is behind a flag `UseOneOfForNullableEnums` which defaults to off.

Also for non body parameters it will always default to non nullable as there is no way of specifying a nullable parameter. 

I've extended the snapshot tests to include more nullable examples. This did uncover another example where if the enum itself is the body of the request it should be marked as nullable.

At the moment I've left inline enums to include null, I'm not sure if this should also behind a flag.

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenOptionsExtensions.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenOptionsExtensions.cs
@@ -300,6 +300,15 @@ public static class SwaggerGenOptionsExtensions
     }
 
     /// <summary>
+    /// Extend enum schemas (using the oneOf construct) to allow null when an enum is referenced
+    /// </summary>
+    /// <param name="swaggerGenOptions"></param>
+    public static void UseOneOfForNullableEnums(this SwaggerGenOptions swaggerGenOptions)
+    {
+        swaggerGenOptions.SchemaGeneratorOptions.UseOneOfForNullableEnums = true;
+    }
+
+    /// <summary>
     /// Enable detection of non nullable reference types to set Nullable flag accordingly on schema properties
     /// </summary>
     /// <param name="swaggerGenOptions"></param>
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/SchemaGenerator.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/SchemaGenerator.cs
@@ -155,6 +155,13 @@ public class SchemaGenerator(
         ParameterInfo parameterInfo,
         ApiParameterRouteInfo routeInfo)
     {
+        var customAttributes = parameterInfo.GetCustomAttributes();
+
+        if (customAttributes.OfType<RequiredAttribute>().Any())
+        {
+            modelType = Nullable.GetUnderlyingType(modelType) ?? modelType;
+        }
+
         var dataContract = GetDataContractFor(modelType);
 
         var schema = _generatorOptions.UseOneOfForPolymorphism && IsBaseTypeWithKnownTypesDefined(dataContract, out var knownTypesDataContracts)
@@ -169,8 +176,6 @@ public class SchemaGenerator(
 
         if (schema.Reference == null)
         {
-            var customAttributes = parameterInfo.GetCustomAttributes();
-
             var defaultValue = parameterInfo.HasDefaultValue
                 ? parameterInfo.DefaultValue
                 : customAttributes.OfType<DefaultValueAttribute>().FirstOrDefault()?.Value;
@@ -178,6 +183,11 @@ public class SchemaGenerator(
             if (defaultValue != null)
             {
                 schema.Default = GenerateDefaultValue(dataContract, modelType, defaultValue);
+            }
+
+            if (Nullable.GetUnderlyingType(modelType) is not null && !customAttributes.OfType<RequiredAttribute>().Any())
+            {
+                schema.Nullable = true;
             }
 
             schema.ApplyValidationAttributes(customAttributes);
@@ -283,8 +293,8 @@ public class SchemaGenerator(
             case DataType.Number:
             case DataType.String:
                 {
-                    schemaFactory = () => CreatePrimitiveSchema(dataContract);
-                    returnAsReference = (Nullable.GetUnderlyingType(dataContract.UnderlyingType) ?? dataContract.UnderlyingType).IsEnum && !_generatorOptions.UseInlineDefinitionsForEnums;
+                    schemaFactory = () => CreatePrimitiveSchema(dataContract, schemaRepository);
+                    returnAsReference = dataContract.UnderlyingType.IsEnum && !_generatorOptions.UseInlineDefinitionsForEnums;
                     break;
                 }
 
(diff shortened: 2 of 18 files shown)


src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenOptionsExtensions.cs
```diff
@@ -299,6 +299,15 @@ public static void UseAllOfToExtendReferenceSchemas(this SwaggerGenOptions swagg
         swaggerGenOptions.SchemaGeneratorOptions.UseAllOfToExtendReferenceSchemas = true;
     }
 
+    /// <summary>
+    /// Extend enum schemas (using the oneOf construct) to allow null when an enum is referenced
```
martincostello, line 303:
```suggestion
    /// Extend enumeration schemas using the <c>oneOf</c> construct to allow <see langword="null"/> when referenced.
```

src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/SchemaGenerator.cs
```diff
@@ -180,6 +185,11 @@ private OpenApiSchema GenerateSchemaForParameter(
                 schema.Default = GenerateDefaultValue(dataContract, modelType, defaultValue);
             }
 
+            if (Nullable.GetUnderlyingType(modelType) is not null && !customAttributes.OfType<RequiredAttribute>().Any())
```
martincostello, line 188: `customAttributes.OfType<RequiredAttribute>().Any()` is being evaluated twice. Instead, cache the result into a local variable on the first access.

martincostello: Similarly, `Nullable.GetUnderlyingType(modelType)` is evaluated twice.

src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/SchemaGenerator.cs
```diff
@@ -329,8 +339,31 @@ private bool TryGetCustomTypeMapping(Type modelType, out Func<OpenApiSchema> sch
...
+
+            if (_generatorOptions.UseInlineDefinitionsForEnums)
+            {
+                enumSchema.Enum.Add(null);
+                return enumSchema;
+            }
+
+            if (_generatorOptions.UseOneOfForNullableEnums)
+            {
+                enumSchema.OneOf = [new OpenApiSchema { Reference = enumSchema.Reference }, new OpenApiSchema() { Enum = [null] }];
```
martincostello, line 360:
```suggestion
                enumSchema.OneOf =
                [
                    new OpenApiSchema { Reference = enumSchema.Reference },
                    new OpenApiSchema { Enum = [null] }
                ];
```

src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/SchemaGeneratorOptions.cs
```diff
@@ -40,6 +40,11 @@ public SchemaGeneratorOptions()
 
     public IList<ISchemaFilter> SchemaFilters { get; set; }
 
+    /// <summary>
+    /// Uses oneOf for enums when the value can be null.
```
martincostello, line 44:
```suggestion
    /// Gets or sets a value indicating whether to extend enumeration schemas using the <c>oneOf</c> construct to allow <see langword="null"/> when referenced.
```

test/Swashbuckle.AspNetCore.Newtonsoft.Test/SchemaGenerator/NewtonsoftSchemaGeneratorTests.cs
```diff
@@ -297,7 +297,24 @@ public void GenerateSchema_DoesNotSetNullableFlag_IfReferencedEnum()
...
+        var referenceSchema = Subject(o => o.UseOneOfForNullableEnums = true).GenerateSchema(typeof(TypeWithNullableProperties), schemaRepository);
+
+        var schema = schemaRepository.Schemas[referenceSchema.Reference.Id];
+        const string propertyName = nameof(TypeWithNullableProperties.NullableIntEnumProperty);
+        Assert.True(schema.Properties[propertyName].Nullable);
+        Assert.Null(schema.Properties[propertyName].Reference);
+        Assert.Equal(2, schema.Properties[propertyName].OneOf.Count);
+        Assert.Equal("IntEnum", schema.Properties[propertyName].OneOf[0].Reference.Id);
+        Assert.Single(schema.Properties[propertyName].OneOf[1].Enum);
+        Assert.Null(schema.Properties[propertyName].OneOf[1].Enum[0]);
```
martincostello, line 317: Missing some additional null asserts and bounds check assert.

test/Swashbuckle.AspNetCore.Newtonsoft.Test/SchemaGenerator/NewtonsoftSchemaGeneratorTests.cs
```diff
@@ -311,6 +328,17 @@ public void GenerateSchema_SetNullableFlag_IfInlineEnum()
...
 
+    [Fact]
+    public void GenerateSchema_AddNullEnumValue_IfInlineEnum()
+    {
+        var schemaRepository = new SchemaRepository();
+
+        var referenceSchema = Subject(o => o.UseInlineDefinitionsForEnums = true).GenerateSchema(typeof(TypeWithNullableProperties), schemaRepository);
+
+        var schema = schemaRepository.Schemas[referenceSchema.Reference.Id];
+        Assert.Null(schema.Properties[nameof(TypeWithNullableProperties.NullableIntEnumProperty)].Enum.Last());
```
martincostello, line 339: Why `Last()`? Looks like more should be asserted here, such as `Enum` not being `null`, as well as the number of values and what they are.

test/Swashbuckle.AspNetCore.SwaggerGen.Test/SchemaGenerator/JsonSerializerSchemaGeneratorTests.cs
```diff
@@ -107,7 +106,33 @@ public void GenerateSchema_GeneratesReferencedEnumSchema_IfEnumOrNullableEnumTyp
...
+        string expectedFormat,
+        params string[] expectedEnumAsJson)
+    {
+        var schemaRepository = new SchemaRepository();
+
+        var referenceSchema = Subject(o => o.UseOneOfForNullableEnums = true).GenerateSchema(type, schemaRepository);
+
+        Assert.Null(referenceSchema.Reference);
+        Assert.True(referenceSchema.Nullable);
+        Assert.Equal(2, referenceSchema.OneOf.Count);
```
martincostello, line 128: Assert `OneOf` is not null.

test/WebSites/MvcWithNullable/Program.cs
```diff
@@ -28,16 +30,39 @@
...
+    public IActionResult GetTypeBody([FromBody] TypeWithNullable typeWithNullable) => Ok(new { typeWithNullable.LogLevel });
 }
 
 [ApiController]
 [Route("api/[controller]")]
 public class RequiredEnumController : ControllerBase
 {
     [HttpGet]
-    public IActionResult Get([Required] LogLevel? logLevel = LogLevel.Error) => Ok(new { logLevel });
+    public IActionResult Get([FromBody, Required] LogLevel? logLevel = LogLevel.Error) => Ok(new { logLevel });
```
martincostello, line 54: Why changed to `[FromBody]`?

martincostello:
@captainsafia Any thoughts on this change, as I imagine null handling has come up in the OpenAPI work for .NET 10? As the original change caused unintended issues this PR is aiming to make opt-in to resolve them, I'd like to make sure we're doing it right this time around 😅.

test/WebSites/MvcWithNullable/Program.cs
ItsVeryWindy: When I fixed outputting null for non body parameters, having a test for a query string parameter no longer made sense as `Required` no longer impacted the result. Whilst it isn't possible to submit a query string with null as the value, `null` is a valid json payload.

src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/SchemaGenerator.cs
ItsVeryWindy: I've realised that the second evaluation for required isn't required as modelType would always be a non nullable type at that stage.

Nullable.GetUnderlyingType(modelType) I feel is more complex
* if it's required it will be called at least once, or possibly twice with the second call using the non nullable type
* if it's not required, it could be called either zero times or once depending on whether reference is null

I won't know until after the schema is generated whether or not it will need called. I could store the result above in a nullable boolean and do `isNullable == true || Nullable.GetUnderlyingType(modelType) is not null`, but it doesn't feel like it's worth it.

martincostello:
@ItsVeryWindy The original change has been manually reverted in #3436 so I can push out a new patch release - you'll need to merge with the default branch and restore the relevant changes in this PR for 8.2.0.

martincostello:
My updated thinking here is:

1. Ship v8.1.3 with #3436.
2. Wait a week or so to check there's no more issues from early adopters.
3. Merge #3422 as v9.0.0 (also unblocks #3399).
4. This PR and #3378 then become part of v9.1.0.
5. #3283 becomes v10.0.0.

martincostello:
Now that it's opt-in I'm kinda OK with this change, but I'd like a second opinion first: https://github.com/domaindrivendev/Swashbuckle.AspNetCore/pull/3428#issuecomment-2925042966

PR #3434, opened 2025-06-02
Fix HumanizeHrefTags not working when see tag spans over multiple lines

Focus1337:
# Pull Request

Updated pattern
```csharp
private const string HrefPattern = @"<see\s+href=\""([^""]*)\"">\s*(.*?)\s*<\/see>";
```

`\s+href=`  to account for optional whitespace (e.g., <see    href=)
non-greedy `.*?` and `\s*` trims leading/trailing whitespace (e.g., newlines)
`RegexOptions.Singleline` to allow `.` to match line breaks.


Focus1337:
Closing because I opened the PR incorrectly and duplicated it by missclick. I apologize.

PR #3444, opened 2025-06-04
Avoid ArgumentNullException being thrown generating examples

### LINKED ISSUE #3442: [Bug]: v8 IncludeXmlComments will cause errors. Also not using Newtonsoft despite being configured
### Describe the bug

```
builder.Services.AddSwaggerGenNewtonsoftSupport();
builder.Services.AddSwaggerGen(c=>
{
    c.IncludeXmlComments(Path.Combine(AppContext.BaseDirectory, "WebApplication1.xml"));
});
```

`IncludeXmlComments(fileName)` causes plenty of `'System.ArgumentNullException' in System.Text.Json.dll` despite Newtonsoft being registered as default for the WebApi and for the SwaggerGen.
The swagger page loads 15 sec.

Per AI:

✅ Root Cause:
Swashbuckle (even with Newtonsoft registered) still uses System.Text.Json behind the scenes to process certain parts of the Swagger generation pipeline, particularly schema generation for complex types. If your models have null or malformed metadata, or attributes that System.Text.Json doesn't handle well (like [JsonConverter] for Newtonsoft), it can throw ArgumentNullException.

martincostello:
Resolves #3442.

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsExampleHelper.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsExampleHelper.cs
@@ -17,7 +17,6 @@ internal static class XmlCommentsExampleHelper
         var exampleAsJson = isStringType
             ? JsonSerializer.Serialize(exampleString)
             : exampleString;
-		
         return JsonModelFactory.CreateFromJson(exampleAsJson);
     }
 }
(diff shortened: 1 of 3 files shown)


src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsExampleHelper.cs
```diff
@@ -17,7 +17,7 @@ public static Microsoft.OpenApi.Any.IOpenApiAny Create(
         var exampleAsJson = isStringType
             ? JsonSerializer.Serialize(exampleString)
             : exampleString;
-
+
```
martincostello:
```suggestion

```

PR #3533, opened 2025-08-13
Remove unrequired public partial class Program

benhopkinstech:
@martincostello 

Off the back of https://github.com/dotnet/aspnetcore/pull/62726

These should no longer be required. However, I've not ran the tests locally - hoping they get run from the build off this PR

### DIFF
--- a/test/Swashbuckle.AspNetCore.IntegrationTests/SwaggerIntegrationTests.cs
+++ b/test/Swashbuckle.AspNetCore.IntegrationTests/SwaggerIntegrationTests.cs
@@ -176,7 +176,7 @@ public class SwaggerIntegrationTests(ITestOutputHelper outputHelper)
         var assembly = Assembly.Load(assemblyName);
         var entryPointType = assembly
             .GetTypes().FirstOrDefault(x => x.Name == "Program")
-            ?? throw new InvalidOperationException($"Could not find the Program entry point type on assembly {assemblyName}.");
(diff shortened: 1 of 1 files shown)


benhopkinstech:
There are actually some tests that look like they require these, I have managed to build locally now, let me look into this.

benhopkinstech:
@martincostello

I have pushed the fixes, let me know if you disagree with how this is implemented.

Test cases are all passing locally now.

martincostello:
> These should no longer be required.

I think the tests show they are required because they're being used to give the easy navigation and type safety between the tests and the code they're testing.

If there's a _different_ type in each assembly/project that can be used to provide the linkage then we can swap over to that then remove the relevant `Program` class, but otherwise I don't think swapping a type for a "magic" string that you can't F12 over to in the IDE is an improvement to save a few lines of code.

benhopkinstech:
> If there's a _different_ type in each assembly/project that can be used to provide the linkage then we can swap over to that then remove the relevant `Program` class

I did consider this, some of them are truly minimal with no other available types on the project. At this point we would need to create some type that was available within the namespace - essentially what we have already.

> I don't think swapping a type for a "magic" string that you can't F12 over to in the IDE is an improvement to save a few lines of code.

I would agree, `ASP0027` doesn't get fired in the way you have implemented anyways so probably best to leave it as is.

martincostello:
OK cool, I'm going to close this, but thanks for the PR anyway!

PR #3574, opened 2025-09-22
Use NuGet Trusted Publishing

### LINKED ISSUE #3566: Switch to NuGet trusted publishing
Once we have access (it's still being progressively rolled-out), we should switch from using API keys to publish packages to NuGet.org to the new [Trusted Publishing](https://learn.microsoft.com/nuget/nuget-org/trusted-publishing) feature.

This is more secure, and avoids the need to renew/rotate API keys. Last time we only found out that the API key had expired when a release failed.

martincostello:
Switch to using GitHub OIDC for pushing packages to NuGet.org with Trusted Publishing.

Resolves #3566.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -227,6 +227,9 @@ jobs:
       name: NuGet.org
       url: https://www.nuget.org/profiles/domaindrivendev
 
+    permissions:
+      id-token: write
+
     steps:
 
     - name: Download packages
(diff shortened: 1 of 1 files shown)


martincostello:
[API key has expired](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/actions/runs/19740911815/job/56565207504#step:4:20). Once Trusted Publishing has been configured can merge this and then release v10.0.2 again.

martincostello:
Relevant changes have been made in NuGet, so this _should_ work now. Will find out for sure when the next release happens.

PR #3662, opened 2025-11-18
Annotate TryLookupByType for nullability

### LINKED ISSUE #3659: Please annotate SchemaRepository.TryLookupByType
I was reviewing some code and noticed a hidden bug: `null` could be returned, but there was no compiler warning. This is because `TryLookupByType` lacks the `[NotNullWhen(true)]` annotation, which is used, for example, in `IPAddress.TryParse`.

It would be helpful if `SchemaRepository.TryLookupByType` were annotated for nullability, similar to my workaround below:

```c#
#nullable enable
public static bool TryLookupByTypeSafe(this SchemaRepository schemaRepository,
    Type type, [NotNullWhen(true)] out OpenApiSchemaReference? referenceSchema)
{
    bool result = schemaRepository.TryLookupByType(type,
        out OpenApiSchemaReference? obliviousReferenceSchema);
    referenceSchema = result ? obliviousReferenceSchema : null;
    return result;
}
#nullable restore
```

And then the hidden bug becomes a compiler warning:

Image

wajahati:
# Pull Request

Fixes #3659

- Annotated `SchemaRepository.TryLookupByType` with `[NotNullWhen(true)]` so callers get nullability warnings when the lookup fails.
- Added the `System.Diagnostics.CodeAnalysis` import and kept the signature otherwise unchanged to maintain API compatibility.
- Verified the entire solution with `dotnet build` and `dotnet test` (net8/net9/net10) to ensure no regressions.

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SchemaRepository.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SchemaRepository.cs
@@ -1,4 +1,6 @@
-﻿using Microsoft.OpenApi;
+﻿using System.Diagnostics.CodeAnalysis;
+
+using Microsoft.OpenApi;
 
 namespace Swashbuckle.AspNetCore.SwaggerGen;
 
@@ -24,7 +26,7 @@ public class SchemaRepository(string documentName = null)
         _reservedIds.Add(type, schemaId);
     }
 
-    public bool TryLookupByType(Type type, out OpenApiSchemaReference referenceSchema)
+    public bool TryLookupByType(Type type, [NotNullWhen(true)] out OpenApiSchemaReference referenceSchema)
     {
         if (_reservedIds.TryGetValue(type, out string schemaId))
         {


wajahati:
@bkoelman I saw the issue you raised and fixed the bug. Can you please review the change and give me a output.
@martincostello  Can you please review the change and check this PR.

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SchemaRepository.cs
```diff
@@ -24,7 +26,7 @@ public void RegisterType(Type type, string schemaId)
         _reservedIds.Add(type, schemaId);
     }
 
-    public bool TryLookupByType(Type type, out OpenApiSchemaReference referenceSchema)
+    public bool TryLookupByType(Type type, [NotNullWhen(true)] out OpenApiSchemaReference? referenceSchema)
```
bkoelman:
```suggestion
#nullable enable
    public bool TryLookupByType(Type type, [NotNullWhen(true)] out OpenApiSchemaReference? referenceSchema)
#nullable restore
```

You can verify manually by calling the method from a project containing `<Nullable>enable</Nullable>` in the csproj and check for a compiler warning, similar to the screenshot in the issue.

wajahati: Thanks for the suggestion! I experimented with wrapping `TryLookupByType` in `#nullable enable`, but PublicAPI analyzer then requires us to re-record every shipped API with nullability, which is a huge churn for a single method. For now I kept the file in its existing nullable-oblivious context and simply added `[NotNullWhen(true)]` the attribute still gives callers in `<Nullable>enable</Nullable>` projects the desired warning. Happy to help if we decide to convert the entire SwaggerGen assembly to nullable-aware APIs later on.

bkoelman: If that works, I'm fine with that. But according to a little experiment, the PublicAPI analyzer can deal with partial annotations. It needs `nullable enable` as the first line in the text file, and then generates signatures with `string!` (non-nullable), `string?` (nullable) and `string` (oblivious).

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SchemaRepository.cs
```diff
@@ -1,4 +1,6 @@
-﻿using Microsoft.OpenApi;
+﻿using System.Diagnostics.CodeAnalysis;
+
+using Microsoft.OpenApi;
```
martincostello, line 3: Please revert this and just fully-qualify the attribute usage as it's the only thing that needs it.

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SchemaRepository.cs
```diff
@@ -24,7 +26,7 @@ public void RegisterType(Type type, string schemaId)
         _reservedIds.Add(type, schemaId);
     }
 
-    public bool TryLookupByType(Type type, out OpenApiSchemaReference referenceSchema)
+    public bool TryLookupByType(Type type, [NotNullWhen(true)] out OpenApiSchemaReference referenceSchema)
```
martincostello, line 29:
```suggestion
    public bool TryLookupByType(Type type, [System.Diagnostics.CodeAnalysis.NotNullWhen(true)] out OpenApiSchemaReference referenceSchema)
```

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SchemaRepository.cs
martincostello: I would prefer not introducing any nullability at this time.

#2374 tracks (maybe, at some point) adding nullability to the NuGet-published Swashbuckle libraries (which is frankly, a non-trivial task), and we can sort it out then. I'd rather not have a patchy implementation.

martincostello:
I'd like to include this change in the next release but as this PR has stalled, I've picked it up in #3719.

Issue #3649, opened 2025-11-12
[Bug]: Issue/Regression related to nullable enums and reference types in v10


zotanmew:
### Describe the bug

Before v10, I was configuring nullability as follows:

```csharp
builder.Services.AddSwaggerGen(o =>
{
    options.SupportNonNullableReferenceTypes();
    options.UseAllOfToExtendReferenceSchemas();
    options.UseAllOfForInheritance();
}
```

This produced correct nullability annotations for enums as well as reference types (including other object schemas) using `nullable: true`.

v10 seems to have changed both how nullability is annotated, as well as how references work. The above setup now produces the nonsensical combination of `type: "null"` with `allOf: [...]`, causing type definition source generators to go haywire (e.g. the typescript one produces an intersection of `null` and the referenced object type definition, which collapses into `never`).

I tried to fix this with a schema filter, however I came to the conclusion that this is not possible to model with the new way references are represented (as the `Type` property is passed through to the reference target, which is obviously not nullable in this context).

The `Microsoft.AspNetCore.OpenApi` generator does not have this issue, though it comes with a myriad of other problems for our codebase, so I'd love for this to get addressed / to get any leads for a possible workaround.

Thank you for this awesome project, and should I have missed anything obvious, apologies in advance.

martincostello:
Thanks for reporting this issue.

Could you provide a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) as a GitHub repository that demonstrates the issue you're experiencing so we can look into this further?

From the description it _sounds_ like you're generating an OpenAPI 3.1 document (which is not the intended default, which is 3.0). If you've opted-in to 3.1, then it might not be an issue with Swashbuckle, but with the tooling you're using being able to understand 3.1.

If you're using 3.0, then it sounds like there's a bug, but I'll need a repro to debug to be able to see where the problem might be (it _could_ be a bug in Microsoft.OpenApi, for example).

zotanmew:
Heya & thanks for the quick reply, I am indeed generating a 3.1 document. Minimal repro:

```csharp
using Microsoft.AspNetCore.Mvc;
using Microsoft.OpenApi;

var builder = WebApplication.CreateBuilder(args);

// Add services to the container.
builder.Services.AddEndpointsApiExplorer();
builder.Services.AddSwaggerGen(o =>
{
	o.SupportNonNullableReferenceTypes();
[...]

PR #3680, opened 2025-11-27
Fix AllOf behaviour when nullable

martincostello:
When a property is using `AnyOf` and is nullable, convert it to use `OneOf` with `null` instead, as it's not possible to otherwise satisfy the constraint of being a schema and being null simultaneously.

Also:

- Remove redundant project references.
- Remove unused `Base64StringAttribute` value.

Relates to #3649.


martincostello:
Verified by the issue reporter: https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/3649#issuecomment-3585672975

Issue #3683, opened 2025-11-27
[Bug]: Broken all/nullable handling


bkoelman:
### Describe the bug

The merge of #3680 is very breaking. For example, it changes the OAS 3.0 component schema structure from:
```json
"links": {
  "allOf": [
    {
      "$ref": "#/components/schemas/error-links"
    }
  ],
  "nullable": true
}
```
to:
```json
"links": {
  "oneOf": [
    {
      "$ref": "#/components/schemas/error-links"
    },
    {
      "nullable": true
    }
  ]
}
```
It looks like the v10.0.2 release containing this change was retracted, but we build/test against the MyGet feed, which still contains this change. As a result, lots of our tests that assert on the output OAS structure are currently broken. Ignoring that, generating a C# client using Kiota against the changed files results in code that doesn't compile. Can the commit be reverted in the main branch as well?

martincostello:
D'oh, I remember now I meant to check this for 3.0 but forgot.

Luckily the NuGet API key had expired so the release never went out.

I won't be reverting it as there's basically just me doing releases anyway, so it'll just be fix forward as I know not to release 10.0.2 until this is sorted.

bkoelman:
As far as I'm aware, the original OAS 3.0 structure is correct. Because title/description/nullable can't be combined with a $ref, the common workaround is to wrap the $ref in an allOf. Swashbuckle provides the `UseAllOfToExtendReferenceSchemas()` method to do that automatically.

For OAS 3.1, the story is different. The allOf workaround is no longer needed (but still allowed). Then my example from above can be reduced to:
```json
"links": {
  "type": "null",
  "allOf": [
    {
      "$ref": "#/components/schemas/error-links"
    }
  ]
}
```
And that works with both NSwag and kiota. What _may_ look even better is to use: `"type": ["null", "object"]`. Unfortunately, doing that breaks the entire ecosystem. It was initially tried in Microsoft.OpenAPI, but reverted later.

Aside from the above, there are two _additional_ legal syntaxes to express the combination of nullable and something else. See https://github.com/OAI/OpenAPI-Specification/issues/3148 for details. Because OAS 3.1 is hardly supported by client generators, it's hard to pick one. 

I found that [openapigenerator](https://github.com/OpenAPITools/openapi-generator) currently only supports the following style in OAS 3.1:
```yaml
RelationshipToOne:
  type: object
  properties:
    data:
      anyOf:
       - type: 'null'
[...]

PR #3686, opened 2025-11-28
Validate OpenAPI documents create valid C# clients

martincostello:
To try and avoid issues such as #3683 in the future, these changes extend the integration tests to validate that the OpenAPI documents produced by the test assets generate valid clients using Kiota and NSwag.

Note that this validation is restricted to compile-time validation only - runtime errors from any generated client code are not detected by these changes.

## TODO

- [x] ~~Iterate over the snapshot files and generate OpenAPI documents that are found~~
- [x] ~~Where relevant, generate OpenAPI documents dynamically for different OpenAPI versions and then use relevant tool(s) to generate clients from them~~
- [x] ~~Remove validation of `petstore3.swagger.io` OpenAPI documents~~
- [ ] Detect the breakage from #3683


martincostello:
Going to merge this now without the validation for the instigating issue present, and tackle that with a separate PR once there's a concrete repro for the specific compilation breakage.

PR #3687, opened 2025-11-29
End-to-end client validation tests

martincostello:
Add a working Minimal API (the venerable "Todo" app) that is used to:

- generate an OpenAPI document
- generate a Kiota client from that OpenAPI document
- verify functional interactions with the API using that generated client code

This was then used to create a repro for #3683, which is then fixed.

This also identified, via the use of <https://apinotes.io/openapi-validator>, that `nullable` shouldn't be emitted for OpenAPI 3.0 if `type` isn't, so that has also been resolved. Unfortunately I haven't yet found a validator that flags that that can be used in CI.

Also:

- extends the tests for client generation with Kiota and NSwag to add code snapshots
- adds a lint step using Spectral to validate any OpenAPI schema snapshots have no errors
- validates the OpenAPI schema snapshots against their relevant OAS JSON schemas

## TODO

- [x] ~~Investigate adding snapshot tests for the NSwag and Kiota clients added by #3686~~
- [x] ~~Investigate using [IBM/openapi-validator](https://github.com/IBM/openapi-validator) or spectral to validate the OpenAPI snapshots~~


test/Swashbuckle.AspNetCore.IntegrationTests/CodeGenerationTests.cs
```diff
@@ -77,11 +86,234 @@ public async Task OpenApiDocument_Generates_Valid_Client_Code_From_Snapshot(Clie
...
+        {
+            var cancellationToken = TestContext.Current.CancellationToken;
+
+            // Act - Get all the items
+            var items = await client.Api.Items.GetAsync(cancellationToken: cancellationToken);
+
+            // Assert - There should be no items
+            Assert.NotNull(items);
+            Assert.NotNull(items.Items);
+            Assert.Empty(items.Items);
```
martincostello:
```suggestion
```

PR #3692, opened 2025-12-01
fix(SchemaGenerator): exclude inherited properties only when base add…

### LINKED ISSUE #3201: [Bug]: Behavior change with UseOneOfForPolymorphism and abstract classes from version 7.0.0 and version 7.2.0
### Describe the bug

I notice a change in the generation of the open API in this situation:

public abstract class A { 
    public string? PropA { get; set; }
}

public class B: A
{
    public string? PropB { get; set; }
}

public class C: B
{
    public string? PropC { get; set; }
}

public class TestController : ControllerBase
{
    [HttpPost]
    public ActionResult Post(A model)
    {
        return Ok();
    }
}

builder.Services.AddSwaggerGen(c =>
{
    c.SwaggerDoc("v1", new OpenApiInfo { Title = "My API", Version = "v1" });
    c.UseOneOfForPolymorphism();
    c.SelectSubTypesUsing(t => typeof(A) == t ? [typeof(C)] : []);
});

### Expected behavior

With version 7.0.0, the behavior is correct.

Immagine 2024-12-16 174955

### Actual behavior

With version 7.2.0, the properties of class B are missing.

Immagine 2024-12-16 174944

### Steps to reproduce

I am attaching the test solution, which currently references version 7.0.0 to test the correct case. To test the incorrect case with the new version, just update the references in the csproj to 7.2.0

[TestSwagger.zip](https://github.com/user-attachments/files/18154019/TestSwagger.zip)

### Exception(s) (if any)

_No response_

7.0.0 and 7.2.0

### .NET Version

net9.0

### Anything else?

_No response_

John-Paul-R:
fix(SchemaGenerator): exclude inherited properties only when base added to AllOf

resolves #3201

---

I've outlined the theory for this change in this comment: https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/3201#issuecomment-3598141153

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/SchemaGenerator.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/SchemaGenerator.cs
@@ -405,7 +405,8 @@ public class SchemaGenerator(
                 }
 
                 applicableDataProperties = applicableDataProperties
-                    .Where(dataProperty => dataProperty.MemberInfo.DeclaringType != baseTypeDataContract.UnderlyingType);
+                    // if the property is declared on a type other than (the one we just added as a base or one of its parents)
+                    .Where(dataProperty => !baseTypeDataContract.UnderlyingType.IsAssignableTo(dataProperty.MemberInfo.DeclaringType));
             }
 
             if (IsBaseTypeWithKnownTypesDefined(dataContract, out var knownTypesDataContracts))
--- a/test/Swashbuckle.AspNetCore.SwaggerGen.Test/snapshots/VerifyTests.GenerateSchema_PreservesMultiLevelInheritance.DotNet10_0.verified.txt
+++ b/test/Swashbuckle.AspNetCore.SwaggerGen.Test/snapshots/VerifyTests.GenerateSchema_PreservesMultiLevelInheritance.DotNet10_0.verified.txt
@@ -115,10 +115,6 @@
           }
         ],
         "properties": {
-          "PropA": {
-            "type": "string",
-            "nullable": true
-          },
           "PropC": {
             "type": "string",
             "nullable": true
--- a/test/Swashbuckle.AspNetCore.SwaggerGen.Test/snapshots/VerifyTests.GenerateSchema_PreservesMultiLevelInheritance.DotNet8_0.verified.txt
+++ b/test/Swashbuckle.AspNetCore.SwaggerGen.Test/snapshots/VerifyTests.GenerateSchema_PreservesMultiLevelInheritance.DotNet8_0.verified.txt
@@ -115,10 +115,6 @@
           }
         ],
         "properties": {
-          "PropA": {
-            "type": "string",
-            "nullable": true
-          },
           "PropC": {
             "type": "string",
             "nullable": true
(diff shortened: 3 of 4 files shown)


martincostello:
Could you also add a test somewhere in the existing integration test project that would produce outputs in a snapshot file please? It's much easier to verify the behaviour in those kinds of tests where the rendered OpenAPI document is visible.

John-Paul-R:
@martincostello Attempted a snapshot test, let me know if that looks as you expect!

test/Swashbuckle.AspNetCore.SwaggerGen.Test/snapshots/VerifyTests.GenerateSchema_PreservesIntermediateBaseProperties_WhenUsingOneOfPolymorphism.DotNet10_0.verified.txt
```diff
@@ -0,0 +1,62 @@
...
+          "content": { }
+        },
+        "responses": {
+          "200": {
+            "description": "OK"
+          }
+        }
+      }
+    }
+  },
```
martincostello, line 72: Could you add a new controller with an action that uses the type? Otherwise it looks a bit odd as the types are "just there", rather than there because a path uses them. That way it's a bit more realistic.

Otherwise the schemas _look_ correct, and some of the other validation we have should flush out any problems if there's any.

John-Paul-R: I don't quite understand, as I did add an action that uses the type (it not showing up here is... odd... to me)

But I can add a new controller that uses both the base and the leaf, sure. 

And I'll try to figure what's off about that existing configuration.

John-Paul-R: Alright, found the issue and pushed a fix with more complete examples in a new controller!

test/Swashbuckle.AspNetCore.SwaggerGen.Test/SchemaGenerator/JsonSerializerSchemaGeneratorTests.cs
```diff
@@ -717,6 +717,60 @@ public void GenerateSchema_SupportsOption_UseAllOfForPolymorphism()
         Assert.Equal(["Property2"], subType2Schema.Properties.Keys);
     }
 
+    [Fact]
+    public void GenerateSchema_PreservesIntermediateBaseProperties_WhenUsingOneOfPolymorphism()
+    {
+        // Arrange - define a type hierarchy A <- B <- C where only A and C are selected as known subtypes
```
martincostello, line 723: Could we also get a test where A, B and C are present to verify the full chain when one isn't skipped? There might be another somewhere in the test case that is already that, but would be nice to have two tests next to each other showing the comparative behaviour?

John-Paul-R: Oh, interesting, I had not considered this case.

I'll make a test with the following expectation:

- A and C continue to behave as they currently do:
  - C is marked as extending A in the schema
- B is a completely flat schema with all properties from B and ~C~ A, no inheritance, as it has not been configured for this.

John-Paul-R: Alright, in addition to the above, I added a test where all of A, B, and C are configured to be included in the inheritance mapping, which revealed a problem in my approach (good call!)

I've updated the code change from

> exclude properties declared on the base type we just added to AllOf

to 

> exclude properties declared on the base type we just added to AllOf, or on any of that type's parents

and the new test's snapshot now looks as I expect

martincostello:
Thanks!

PR #3708, opened 2025-12-06
Add public method SchemaRepository.ReplaceSchemaId

### LINKED ISSUE #3707: [Feature request]: Add API to change schema ID

In our codebase, in some scenarios, a constructed `System.Type` is not sufficient to determine a unique schema ID. To make that fit the existing Swashbuckle `SchemaGenerator` API, we:
1. Let Swashbuckle generate the schema using `SchemaGenerator`
2. Rename the generated schema ID, based on an additional string value

The rename happens by obtaining the generated schema from `SchemaRepository`, removing it and then adding it under a different ID. However, we also need to update the private `_reservedIds` field for that to work, which we currently do using reflection.

I'd like to propose a new method on `SchemaRepository` to rename the schema ID for an existing `System.Type`.
```c#
public void ReplaceSchemaId(Type oldSchemaType, string newSchemaId);
```
The implementation would resemble what we currently use:
```c#
private static readonly FieldInfo ReservedIdsField; // loaded via reflection

public static void ReplaceSchemaId(this SchemaRepository schemaRepository,
    Type oldSchemaType, string newSchemaId)
{
    ArgumentNullException.ThrowIfNull(schemaRepository);
    ArgumentNullException.ThrowIfNull(oldSchemaType);
    ArgumentException.ThrowIfNullOrEmpty(newSchemaId);

    if (schemaRepository.TryLookupByType(oldSchemaType, out OpenApiSchemaReference? referenceSchema))
    {
        string oldSchemaId = referenceSchema.Reference.Id!;

        IOpenApiSchema targetSchema = schemaRepository.Schemas[oldSchemaId];

        schemaRepository.Schemas.Remove(oldSchemaId);
        schemaRepository.Schemas.Add(newSchemaId, targetSchema);

        var reservedIds = (Dictionary<Type, string>)ReservedIdsField.GetValue(schemaRepository)!;
        reservedIds.Remove(oldSchemaType);
    }
}
```
Although we don't need it, it may be better to return a boolean indicating whether the replacement has happened.

Once approved, I'm willing to open a PR to implement this.

### Additional context

_No response_

bkoelman:
Adds a public method to change the ID of a schema in the repository after generation.

Closes #3707.

### DIFF
--- a/test/Swashbuckle.AspNetCore.SwaggerGen.Test/SchemaRepositoryTests.cs
+++ b/test/Swashbuckle.AspNetCore.SwaggerGen.Test/SchemaRepositoryTests.cs
@@ -20,6 +20,24 @@ public static class SchemaRepositoryTests
     }
 
     [Fact]
+    public static void ReplaceSchemaId_MultipleSchemas_ReplacesSchemaId()
+    {
+        var repository = new SchemaRepository();
+        (string oldSchemaId, IOpenApiSchema exampleSchema) = GenerateSchemaForType(typeof(Example), repository);
+        (string otherSchemaId, IOpenApiSchema otherSchema) = GenerateSchemaForType(typeof(Exception), repository);
+
+        string newSchemaId = "alternateId";
+        bool result = repository.ReplaceSchemaId(typeof(Example), newSchemaId);
+
+        Assert.True(result);
+        Assert.DoesNotContain(oldSchemaId, repository.Schemas);
+        Assert.Contains(newSchemaId, repository.Schemas);
+        Assert.Same(exampleSchema, repository.Schemas[newSchemaId]);
+        Assert.Contains(otherSchemaId, repository.Schemas);
+        Assert.Same(otherSchema, repository.Schemas[otherSchemaId]);
+    }
+
+    [Fact]
     public static void ReplaceSchemaId_MissingSchema_DoesNotChange()
     {
         var repository = new SchemaRepository();


src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SchemaRepository.cs
```diff
@@ -42,4 +42,23 @@ public OpenApiSchemaReference AddDefinition(string schemaId, OpenApiSchema schem
 
         return new OpenApiSchemaReference(schemaId);
     }
+
+    public bool ReplaceSchemaId(Type oldSchemaType, string newSchemaId)
```
martincostello:
```suggestion
    public bool ReplaceSchemaId(Type schemaType, string replacementSchemaId)
```

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SchemaRepository.cs
```diff
@@ -42,4 +42,23 @@ public OpenApiSchemaReference AddDefinition(string schemaId, OpenApiSchema schem
...
+    {
+        ArgumentNullException.ThrowIfNull(oldSchemaType);
+        ArgumentException.ThrowIfNullOrEmpty(newSchemaId);
+
+        if (_reservedIds.TryGetValue(oldSchemaType, out string oldSchemaId) &&
+            oldSchemaId != newSchemaId && Schemas.TryGetValue(oldSchemaId, out var targetSchema))
```
martincostello:
```suggestion
            oldSchemaId != newSchemaId &&
            Schemas.TryGetValue(oldSchemaId, out var targetSchema))
```

test/Swashbuckle.AspNetCore.SwaggerGen.Test/SchemaRepositoryTests.cs
```diff
@@ -0,0 +1,101 @@
+﻿using Microsoft.OpenApi;
+
+namespace Swashbuckle.AspNetCore.SwaggerGen.Test;
+
+public class SchemaRepositoryTests
```
martincostello: As there's no state, the class and all the tests can be `static`.

martincostello: All the test cases only deal with a repository with a single associated `Type`, so to be more comprehensive there should be at least one test case where there's multiple types involved.

bkoelman: I don't understand how that is relevant. All branches are already covered by tests. Do you have a specific scenario in mind?

martincostello: If I just look at the public API and _not_ at the implementation, I think to myself _"if there are two types A and B, and I ask it to replace B, does it correctly replace B and not do anything with A?"_ - there doesn't appear to be a test case that validates that.

bkoelman: I can think of a thousand tests that validate something does *not* happen. :-)
But don't we write tests to verify the implementation is correct? I'm not sure what you want me to add or change.

martincostello: A test that adds two types to the repository and then edits one of them and validates the state after the edit has left one untouched and the other edited as expected.

bkoelman: Added

martincostello (approved):

PR #3719, opened 2025-12-10
Annotate TryLookupByType for nullability

### LINKED ISSUE #3659: Please annotate SchemaRepository.TryLookupByType

martincostello:
Supersedes #3662 to resolve #3659.

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SchemaRepository.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SchemaRepository.cs
@@ -27,7 +27,7 @@ public class SchemaRepository(string documentName = null)
     public bool TryLookupByType(Type type, [System.Diagnostics.CodeAnalysis.NotNullWhen(true)] out OpenApiSchemaReference referenceSchema)
     {
         referenceSchema = null;
-        bool result =_reservedIds.TryGetValue(type, out string schemaId);
+        bool result = _reservedIds.TryGetValue(type, out string schemaId);
 
         if (result)
         {


src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SchemaRepository.cs
```diff
@@ -24,16 +24,17 @@ public void RegisterType(Type type, string schemaId)
         _reservedIds.Add(type, schemaId);
     }
 
-    public bool TryLookupByType(Type type, out OpenApiSchemaReference referenceSchema)
+    public bool TryLookupByType(Type type, [System.Diagnostics.CodeAnalysis.NotNullWhen(true)] out OpenApiSchemaReference referenceSchema)
     {
-        if (_reservedIds.TryGetValue(type, out string schemaId))
+        referenceSchema = null;
+        bool result =_reservedIds.TryGetValue(type, out string schemaId);
```
martincostello:
```suggestion
        bool result = _reservedIds.TryGetValue(type, out string schemaId);
```

PR #3772, opened 2026-02-05
Fix browser caching behaviour

martincostello:
# Pull Request

https://github.com/domaindrivendev/Swashbuckle.AspNetCore/discussions/3709 and https://github.com/domaindrivendev/Swashbuckle.AspNetCore/discussions/3768.

- Fix ETags not being used due to non-zero `max-age` (#3709).
- Fix incorrect weak ETag type for hashed files.
- Consistently respond with HTTP 304 for unmodified static files.
- Fix document URL not responding with correct response if custom JSON serializer options are used (#3768).

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerUI/SwaggerUIMiddleware.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerUI/SwaggerUIMiddleware.cs
@@ -208,7 +208,7 @@ internal sealed partial class SwaggerUIMiddleware
 
             SetHeaders(response, _options, etag);
 
-            await response.WriteAsync(json, Encoding.UTF8);
+            await response.WriteAsync(json, Encoding.UTF8, context.RequestAborted);
         }
     }
 


src/Swashbuckle.AspNetCore.SwaggerUI/SwaggerUIMiddleware.cs
```diff
@@ -182,20 +192,32 @@ private async Task RespondWithDocumentUrls(HttpResponse response)
             json = JsonSerializer.Serialize(l, SwaggerUIOptionsJsonContext.Default.ListUrlDescriptor);
         }
 
-        json ??= JsonSerializer.Serialize(_options.ConfigObject, _jsonSerializerOptions);
+        json ??= JsonSerializer.Serialize(_options.ConfigObject.Urls, _jsonSerializerOptions);
```
pinkfloydx33, line 195: Doesn't this still suffer the issue? The json variable has a non-null default value, so if you pass json options it won't null coalesce

Or am I missing something?

martincostello: #3773

PR #3788, opened 2026-02-14
Fix null examples being represented as the string null

### LINKED ISSUE #3784: [Bug]: Need a way to represent null inside XML comments <example> tag.
### Describe the bug

I can't find a way to return an example value of null for a string property.  Our project does not have nullable reference types enabled, so by default a string variable/property is nullable.  I've tried a couple of options:

```
public class Response
{
    /// <example>null</example>
    public string Message{ get; set; }
}
```

I've also tried:

```
public class Response
{
    /// <example><see langword="null"/></example>
    public string Message{ get; set; }
}
```
And:

```
public class Response
{
    /// <example xsi:nil="true"/>
    public string Message{ get; set; }
}
```
but none of these end up in getting a null value into the example swagger documentation.  The generated documentation shows the example as:
`"message": "null",`

### Expected behavior

I would like this xml example comment
```
/// <example>null</example>
public string Message{ get; set; }
```

to generate an example like this:
`"message": null`

### Actual behavior

generates as if null is a literal string value:
`"message": "null",`

### Steps to reproduce

_No response_

### Exception(s) (if any)

_No response_

10.1.0

### .NET Version

<TargetFramework>net10.0</TargetFramework>

### Anything else?

Seems to be a regression of this issue that was resolved a long time ago:
https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/1768

And seems to have regressed here: https://github.com/domaindrivendev/Swashbuckle.AspNetCore/pull/3283

I can see there was a unit test for the earlier issue that was changed in a way to make the test pass, but the result is not correct/desired.  The change to the unit test is here: test/Swashbuckle.AspNetCore.SwaggerGen.Test/XmlComments/XmlCommentsSchemaFilterTests.cs

Changed from  
`{ typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.StringPropertyWithNullExample), JsonSchemaTypes.String, "null" },`
to
`{ typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.StringPropertyWithNullExample), JsonSchemaTypes.String, "\"null\"" },`

But that changes the whole purpose of the test, which is to be able to generate an example with a null string value.  Not a literal "null".

jgarciadelanoceda:
Fixes #3784.

I think I have found another thing derivated from the OpenAPI.NET upgrate from 1.X to 2.X, previously when there was declared the example as null that example could be shown on the OpenApi document. Now I cannot create a JsonNode from "null" string and the JsonNode not being null so the serializer does not serialize it(A minor thing I believe).

I tried creating a JsonNode from null string many ways but I think it's not possible by design JsonObject.Parse("null")

https://github.com/microsoft/OpenAPI.NET/blob/8bec88013cbb0f86f7f43dfdb039ce83bf83beaf/src/Microsoft.OpenApi/Models/OpenApiSchema.cs#L535

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsExampleHelper.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsExampleHelper.cs
@@ -19,7 +19,7 @@ internal static class XmlCommentsExampleHelper
 
         if (isStringType)
         {
-            return "null".Equals(exampleString) ? null : JsonValue.Create(exampleString);
+            return string.Equals(exampleString, "null") ? null : JsonValue.Create(exampleString);
         }
 
         // HACK If the value is a string, but we can't detect it as one, then


martincostello:
> I tried creating a JsonNode from null string many ways but I think it's not possible by design JsonObject.Parse("null")

I think that's basically why I ended up changing the code that caused the bug in the first place. I couldn't get it to work and assumed that it wouldn't matter.

PR #3793, opened 2026-02-17
Return null on example, as it was with OpenApiNull before

jgarciadelanoceda:
Ends fixing again #3784 .. I saw there was a thing a bit hidden on Microsoft.OpenApi that allows us to serialize the null as we were doing before with OpenApiNull


jgarciadelanoceda:
@martincostello what are your thoughts about this one? I found it while answering the comment.. it's a bit nitch but in previous versions <10 Swashbuckle was serializing null in the example

PR #3802, opened 2026-02-18
Fix null examples being wrongly converted

### LINKED ISSUE #3801: [Bug]: Example Null-Handling Seems Inverted
### Describe the bug

I was actually happy to see #3793 as this was a minor gap for us in the .NET 10 upgrade where we had some examples that were explicitly/intentionally written to the API Spec as `null`. 

However moving to the new package 10.1.4 seems to exhibit strange behavior. I am only allowed to write a `null` example to the document if the data type is `JsonSchemaType.String` but _not_  `JsonSchemaTypeString | JsonSchemaType.Null`.  

https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/8ad6682f15c4df3a3ad2ad8815b9af914af20b1f/src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsExampleHelper.cs#L16-L23

This seems inverted! I can set `example=null` on a **non-nullable** `string`, but _not_ on a **nullable** `string?`
It also seems a bit funny that it's limited to strings and not any nullable type

### Expected behavior

At the very least I would expect to be able to set the example value to `null` if the property itself supports being null i.e. with `JsonSchemaType.Null | JsonSchemaType.String`

Though I probably would not limit this to just strings at all! Perhaps something like this could suffice for any type of property:

```csharp
var type = schema?.ResolveType(schemaRepository);
if (type is { } value && value.HasFlag(JsonSchemaType.Null) && string.Equals(exampleString, "null"))
{
    return JsonNullSentinel.JsonNull;
}

// rest of method
```

### Actual behavior

Cannot write `"example": null` to document if schema resolves to `JsonSchemaTypeString | JsonSchemaType.Null`
Cannot write `"example": null` for numeric or other properties that have the `JsonSchemaType.Null` flag

### Steps to reproduce

Consider this data type: 

```csharp
public sealed class MyType
{
    /// <example>null</example>
    [Required]
    public string Ex0 { get; set; } = null!;  
    /// <example>null</example>
    public string? Ex1 { get; set; }
    /// <example>null</example>
    public Guid Ex2 { get; set; }
    /// <example>null</example>
    public Guid? Ex3 { get; set; }
    /// <example>null</example>
    public int Ex4 { get; set; }
    /// <example>null</example>
    public int? Ex5 { get; set; }
    /// <example>3</example>
    public int? Ex6 { get; set; }
}
```

The API Specification generated will produce

| Field | Expected Example | Actual Example | Correct |
| -- | -- | -- | -- |
| `Ex0` | omitted | `"example": null` |  :x: |
| `Ex1` |  `"example": null` | omitted |  :x: |
| `Ex2` |  omitted |  `"example": null` | :x: |
| `Ex3` |  `"example": null` | omitted |  :x: | 
| `Ex4` |  omitted |  omitted | :white_check_mark: | 
| `Ex5` |  `"example": null` | omitted |  :x: |
| `Ex6` |  `"example": 3` | `"example": 3` |  :white_check_mark: |

### Exception(s) (if any)

_No response_

10.1.4

### .NET Version

10

### Anything else?

As a work-around, you can write the literal value of `JsonNullSentinel.NullValue` as your example and it will get converted just fine:

```csharp
public sealed class MyType
{
    /// <example>openapi-json-null-sentinel-value-2BF93600-0FE4-4250-987A-E5DDB203E464</example>  
    public string? Ex1 { get; set; }
}
```

However that is a bit heavy handed and likely an implementation detail!

dldl-cmd:
Fixes #3801

Changes the JsonModelFactory, which is used in more cases than the XML comment examples to convert a json string to JsonNode with special handling for the null json string.

Moreover, adding a special rule for the null example on string which results in either the string null (`"example": "null"`) or the json value null (`"example": null`)


jgarciadelanoceda:
@dldl-cmd , let me know what you think about this other one #3803.. I marked my PR as draft to see if you agree to proceed

test/Swashbuckle.AspNetCore.IntegrationTests/snapshots/10_0/VerifyTests.SwaggerEndpoint_ReturnsValidSwaggerJson_startupType=Basic.Startup_swaggerRequestUri=v1.verified.txt
```diff
@@ -802,9 +802,9 @@
             "required": true,
             "schema": {
               "type": "string",
-              "example": null
+              "example": "null"
```
jgarciadelanoceda, line 805: For applications that do not have nullable enable this sounds weird

dldl-cmd: Why is the weird to applications which do not have nullable enabled? When the generated schema does indicate, that null isn't allowed having null in the example is wrong. For strings to use than just the value the developer has put there in the example would be a single a good understandable solution.

jgarciadelanoceda: But an application which has nullable enabled the desired behaviour I think it'll be to present that value as null(There is no possibility to use string?).. Indeed in the SwashBuckle versions prior to 10 that was the behaviour.. What I tried to do in #3803 is to put the null when the type is null or when is a string that is not DateTime/DateOnly/TimeOnly/Timespan (Because those had the possibility always to have the DateTime?/DateOnly?...)

Also think about the possibility that the client is using OpenApi 3.0.. To put the example as null seems fine.. Also I think that if the client put null we must never put the nullstring.. Because the client mean null not "null"(Which was the original issue #3784)

jgarciadelanoceda: If you want you can merge my changes (There are a couple of test more with DateTime/DateOnly).. And get to a consensus

dldl-cmd:
> @dldl-cmd , let me know what you think about this other one #3803.. I marked my PR as draft to see if you agree to proceed

1. Your PR doesn't fix the issue for the default value, where the same issue can occur.
2. For the differentiation between arrays and objects in the null handling I don't see a good reason to do this.
3. Your check on a string type for a format fixes just some cases, where null is the value. In general the whole example handling doesn't care about whether the value of the example or also default value matches the schema. Therefore adding here some special cases feels to me like mixing concerns.

   Alone with the [string formats (list could be extended at any time)](https://spec.openapis.org/registry/format/index.html) there are other cases like IP addresses where a string with the text null doesn't make sense. More over Pattern, MinLength, MaxLength are not considered. But then maybe also enum values need to be considered. On numbers minimum and maximum. On array minItems and maxItems and so on.
4. Regardless of our changes you can do the following, which shows how completely wrong an example can be:
    ```csharp
    /// <example>{ "property": "value", "complex": { "propA": 1, "propB": null} }</example>
    public int IntExample { get; set; }
    ```
    Results in:
    ```json
    {
      "type": "integer",
      "format": "int32",
      "example": {
        "property": "value",
        "complex": {
          "propA": 1,
          "propB": null
        }
      }
    }
    ```

test/Swashbuckle.AspNetCore.IntegrationTests/snapshots/10_0/VerifyTests.SwaggerEndpoint_ReturnsValidSwaggerJson_startupType=Basic.Startup_swaggerRequestUri=v1.verified.txt
dldl-cmd: The library already handles the nullable type indication correctly. So when nullable is enabled

```csharp
public string Prop { get; set; };
```
```json
{
  "type": "string"
}
````
If nullable annotations are not enabled:
```csharp
public string Prop { get; set; };
```
```json
{
  "type": "string",
  "nullable": true
}
````
Also the difference between OpenApi 3.0 and 3.1 doesn't make any difference here as the OpenApi.NET library abstracts the difference away:


**OpenApi v3.0**
```json
{
  "type": "string",
  "nullable": true
}
```

**OpenApi v3.1**
```json
{
   "type": ["null","string"],
}
```

jgarciadelanoceda:
> > @dldl-cmd , let me know what you think about this other one #3803.. I marked my PR as draft to see if you agree to proceed
> 
> 1. Your PR doesn't fix the issue for the default value, where the same issue can occur.
> 2. For the differentiation between arrays and objects in the null handling I don't see a good reason to do this.
> 3. Your check on a string type for a format fixes just some cases, where null is the value. In general the whole example handling doesn't care about whether the value of the example or also default value matches the schema. Therefore adding here some special cases feels to me like mixing concerns.
>    Alone with the [string formats (list could be extended at any time)](https://spec.openapis.org/registry/format/index.html) there are other cases like IP addresses where a string with the text null doesn't make sense. More over Pattern, MinLength, MaxLength are not considered. But then maybe also enum values need to be considered. On numbers minimum and maximum. On array minItems and maxItems and so on.
> 4. Regardless of our changes you can do the following, which shows how completely wrong an example can be:
>    ```cs
>    /// <example>{ "property": "value", "complex": { "propA": 1, "propB": null} }</example>
>    public int IntExample { get; set; }
>    ```
>    
>    
>        
>          
>        
>    
>          
>        
>    
>        
>      
>    Results in:
>    ```json
>    {
>      "type": "integer",
>      "format": "int32",
>      "example": {
>        "property": "value",
>        "complex": {
>          "propA": 1,
>          "propB": null
>        }
>      }
>    }
>    ```

Could you talk me about the first point?, I don't get to see it
About the second/third point I just checked if the object is nullable.. If you see the issue #3801, most of the test cases are fine with the implementation.

I do not like to put "null" as an example because if the example was set as null we shouldn´t create the example as "null", if they want to set the example as "null", they could put <example>"null"</example>

dldl-cmd:
For the first point the `JsonModelFactory` is also used to convert a json string for default values.


Why I think putting `"null"` as an example instead of skipping the example or setting the example to `null`:

- When the schema doesn't allow null, the example shouldn't be `"example": null` because this would generate an invalid OpenApi document.
- Therefore two options remain. Skipping the example or having `"example": "null"`:
  - As an example was added in code we can expect, that as a result in the generated document is expected
  - As the schema indicates, the type is just a string. It makes sense to take what ever is inside of it. 


Furthermore considering these two cases:
```csharp
/// <example>null</example>
public required string NullExampleOnNotNullable { get; set; }

/// <example>null2</example>
public required string Null2ExampleOnNotNullable { get; set; }
````

My expectation would be that I see:
```json
{
    "NullExampleOnNotNullable": {
        "type": "string",
        "example": "null"
    },
    "Null2ExampleOnNotNullable": {
        "type": "string",
        "example": "null2"
    }
}
```

And for the nullable case:

```csharp
/// <example>null</example>
public string? NullExampleOnNullable { get; set; }

/// <example>null2</example>
public string? Null2ExampleOnNullable { get; set; }
````
```json
{
    "NullExampleOnNullable": {
        "type": "string",
        "nullable": true,
        "example": null
    },
    "Null2ExampleOnNullable": {
        "type": "string",
        "nullable": true,
        "example": "null2"
    }
}
```

jgarciadelanoceda:
With your branch for simple int values with example set to null it will get serialized to null which is incorrect right?.
If I just use the example on the SwashBuckle, it's a wrong example.
 Also the change you did on JsonModelFactory isn´t going to be hitted because on all the calls made to XmlCommentsExampleHelper.Create there is a check to do not pass null(I mean the  if (json is null))


From the table sent on the issue my branch does not fail at any case

jgarciadelanoceda:
@dldl-cmd  should I remove the draft from my PR? if so you can do the comments there ;)

PR #3803, opened 2026-02-18
Fix nullExamples

### LINKED ISSUE #3801: [Bug]: Example Null-Handling Seems Inverted

jgarciadelanoceda:
Fixes #3801.
As previously the null was being always serialized and also keep compatibility with not nullable references

As the integer/double could be always have the nullableReference because the objects where int?,double? for old versions seems fine.
I also tought about the object case, that for projects that do not have the nullable references enabled seems OK to me to have that case and also for the case of strings.. I do not really know if I should also place an specific case for arrays


src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsExampleHelper.cs
```diff
@@ -17,9 +17,15 @@ public static JsonNode Create(
...
+        {
+
+            return type is { } jsonSchema && jsonSchema.HasFlag(JsonSchemaType.Null) ? JsonNullSentinel.JsonNull : null;
+        }
+
         if (isStringType)
         {
-            return string.Equals(exampleString, "null") ? JsonNullSentinel.JsonNull : JsonValue.Create(exampleString);
+            return JsonValue.Create(exampleString);
         }
```
dldl-cmd: Changing the order makes this a bit better readable. The check `!value.HasFlag(JsonSchemaType.Null)` should able also redundant as this case is handled before.

```suggestion
        if (string.Equals(exampleString, "null"))
        {
            return type is { } jsonSchema && jsonSchema.HasFlag(JsonSchemaType.Null) ? JsonNullSentinel.JsonNull : null;
        }
        
   var isStringType = type is { } value &&
       value.HasFlag(JsonSchemaType.String) &&
       !value.HasFlag(JsonSchemaType.Null);
   if (isStringType)
   {
        return JsonValue.Create(exampleString);
   }
```

test/Swashbuckle.AspNetCore.SwaggerGen.Test/XmlComments/XmlCommentsSchemaFilterTests.cs
```diff
@@ -52,32 +52,64 @@ public void Apply_SetsDescription_FromPropertySummaryTag(
...
+        { typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.NullableGuidPropertyWithNullExample), JsonSchemaTypes.String | JsonSchemaType.Null, "null", "uuid" },
+        { typeof(XmlAnnotatedRecord), nameof(XmlAnnotatedRecord.NullableGuidPropertyWithNullExample), JsonSchemaTypes.String | JsonSchemaType.Null, "null", "uuid" },
+        { typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.GuidPropertyWithNullExample), JsonSchemaTypes.String, null,"uuid"  },
+        { typeof(XmlAnnotatedRecord), nameof(XmlAnnotatedRecord.GuidPropertyWithNullExample), JsonSchemaTypes.String, null,"uuid"  },
+        { typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.ObjectPropertyNullExample), JsonSchemaTypes.String, null,null },
+        { typeof(XmlAnnotatedRecord), nameof(XmlAnnotatedRecord.ObjectPropertyNullExample), JsonSchemaTypes.String, null,null },
+        { typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.NullableObjectPropertyNullExample), JsonSchemaTypes.String | JsonSchemaType.Null, "null",null},
+        { typeof(XmlAnnotatedRecord), nameof(XmlAnnotatedRecord.NullableObjectPropertyNullExample), JsonSchemaTypes.String | JsonSchemaType.Null, "null",null},
```
dldl-cmd:
```suggestion
        { typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.ObjectPropertyNullExample), JsonSchemaTypes.Object, null,null },
        { typeof(XmlAnnotatedRecord), nameof(XmlAnnotatedRecord.ObjectPropertyNullExample), JsonSchemaTypes.Object, null,null },
        { typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.NullableObjectPropertyNullExample), JsonSchemaTypes.Object | JsonSchemaType.Null, "null",null},
        { typeof(XmlAnnotatedRecord), nameof(XmlAnnotatedRecord.NullableObjectPropertyNullExample), JsonSchemaTypes.Object | JsonSchemaType.Null, "null",null},
```

jgarciadelanoceda:
@martincostello could you review this?, I think that there is an extra property that is the format that is right now no neccesary but I did at the first implementation needed(I could remove it by the wat)

martincostello:
> could you review this?

I see all issues, PRs and comments in this repo - I get to stuff when it fits into my day and my TODO/priority list 😃

src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsExampleHelper.cs
```diff
@@ -17,10 +17,27 @@ public static JsonNode Create(
...
+            return type is { } jsonSchema && jsonSchema.HasFlag(JsonSchemaType.Null) ? JsonNullSentinel.JsonNull : null;
         }
+        
+   var isStringType = type is { } value &&
+       value.HasFlag(JsonSchemaType.String) &&
+       !value.HasFlag(JsonSchemaType.Null);
+   if (isStringType)
+   {
+        return JsonValue.Create(exampleString);
+   }
```
jgarciadelanoceda:
```suggestion
```

src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsExampleHelper.cs
```diff
@@ -13,15 +13,19 @@ public static JsonNode Create(
...
 
-        var isStringType = type is { } value &&
-            value.HasFlag(JsonSchemaType.String) &&
-            !value.HasFlag(JsonSchemaType.Null);
 
-        if (isStringType)
+        if (string.Equals(exampleString, "null"))
         {
-            return string.Equals(exampleString, "null") ? JsonNullSentinel.JsonNull : JsonValue.Create(exampleString);
+            return (type?.HasFlag(JsonSchemaType.Null) ?? false) ? JsonNullSentinel.JsonNull : null;
```
jgarciadelanoceda: I was doubting too much here.
Because for applications with nullable enable seems right. But for applications that do not have it enabled in the versions previous to 10 Swashbuckle was serializing the value as null

It might need a new option on SwaggerOptions to keep compability.. also a think that I thought was to serialize it as null when is a string and it's not a Guid/Timespan/... I left here my doubt just to see what is the desired approach

src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsExampleHelper.cs
```diff
@@ -13,15 +13,19 @@ public static JsonNode Create(
...
         }
 
+        if (type is { } value
+        && value.HasFlag(JsonSchemaType.String))
+        {
+            return JsonValue.Create(exampleString);
+        }
+
+
```
martincostello:
```suggestion
        if (type is { } value && value.HasFlag(JsonSchemaType.String))
        {
            return JsonValue.Create(exampleString);
        }

```

test/Swashbuckle.AspNetCore.IntegrationTests/snapshots/10_0/VerifyTests.SwaggerEndpoint_ReturnsValidSwaggerJson_startupType=Basic.Startup_swaggerRequestUri=v1.verified.txt
martincostello, line 1: Just to clarify, why is it that we added this in v10.1.3 but now we're taking it away again?

test/Swashbuckle.AspNetCore.SwaggerGen.Test/XmlComments/XmlCommentsExampleHelperTests.cs
```diff
@@ -74,7 +74,7 @@ public void Create_Returns_Null_When_Value_And_Schema_Are_Null()
     [Fact]
     public void Create_Returns_Null_When_Type_String_And_Value_Null_String_Literal()
     {
-        var schema = new OpenApiSchema { Type = JsonSchemaTypes.String };
+        var schema = new OpenApiSchema { Type = JsonSchemaTypes.String | JsonSchemaType.Null };
```
martincostello: Maybe we should have tests for with and without nullable, rather than changing the existing tests to be nullable?

test/Swashbuckle.AspNetCore.SwaggerGen.Test/Fixtures/XmlAnnotatedType.cs
```diff
@@ -124,20 +142,109 @@ public void AcceptsConstructedOfConstructedGenericType(
...
+    /// Summary for NullableIntPropertyWithNullExample
+    /// </summary>
+    /// <example>null</example>
+    public int? NullableIntPropertyWithNullExample { get; set; }
+
+    /// <summary>
+    /// Summary for IntPropertyWithNullExample
+    /// </summary>
+    /// <example>null</example>
+
```
martincostello: Nit: comments should sit on the member, not have a blank line between them.

test/Swashbuckle.AspNetCore.IntegrationTests/snapshots/10_0/VerifyTests.SwaggerEndpoint_ReturnsValidSwaggerJson_startupType=Basic.Startup_swaggerRequestUri=v1.verified.txt
jgarciadelanoceda: This is exactly what I am concerned about if you see the issue #3784, you will see that the requirement was to put a null there.

If you see the #3801, you will see more examples of what is right/wrong 

I think that the problem behind is that one of them is having null references enabled while the other doesn't.

I thought about what should we do.. because in versions lower than 10 we just placed null in the examples, not caring about if the type is nullable or not. For the special case of strings the thing is that when you do not set the null reference enabled it's perfectly fine to put the example as null..

I do not know if we should check for the nullabilty or not.. and from my perspective it's good to look at it( we may need a new SwagerGeneration option to solve both issues), by default the behaviour that was in previous versions and do not check for nullabilty there and if defined then do not serialize the example and have the new behaviour

test/Swashbuckle.AspNetCore.SwaggerGen.Test/XmlComments/XmlCommentsExampleHelperTests.cs
jgarciadelanoceda: See the other answer.. I am not sure about a behaviour of null examples

test/Swashbuckle.AspNetCore.SwaggerGen.Test/Fixtures/XmlAnnotatedType.cs
jgarciadelanoceda: fixed

src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsExampleHelper.cs
jgarciadelanoceda: Fixed

test/Swashbuckle.AspNetCore.IntegrationTests/snapshots/10_0/VerifyTests.SwaggerEndpoint_ReturnsValidSwaggerJson_startupType=Basic.Startup_swaggerRequestUri=v1.verified.txt
martincostello: > not caring about if the type is nullable or not. For the special case of strings the thing is that when you do not set the null reference enabled it's perfectly fine to put the example as null.

I don't quite follow - with NRT enabled, `string` won't allow null while `string?` will.

jgarciadelanoceda: That's what I am concern about.. If you see the issue #3801 is
incompatible with the issue #3784.

I see more acurate to only put example = null for string?. Not for
string.. But this is a breaking change.. And I think we should do one
of these two options:

- Create a setting in SWaggerGeneration to allow the behaviour of
3801.. The default value will be to serialize always the value as
null.. If set to the oposite it will not be always be
serialized(Depending on the type allows null or not)
- Don´t care about if the type allows null or not and always serialize
the example as null. This is not so scary because you could even write
the example as an object when the type is string

I would do as you think is the path forward, because neither of the
options convince me

martincostello: #3801 looks correct to me so far that if you use `string?` then you should be able to set an example to `null`.

In theory we could just let users use `null` regardless of whether nullable is in use at all (i.e. not bother checking and just make `"null"` a special token), and/or add a special keyword to mean that.

src/Swashbuckle.AspNetCore.SwaggerGen/XmlComments/XmlCommentsExampleHelper.cs
```diff
@@ -13,15 +13,18 @@ public static JsonNode Create(
...
         {
-            return string.Equals(exampleString, "null") ? JsonNullSentinel.JsonNull : JsonValue.Create(exampleString);
+            return JsonNullSentinel.JsonNull;
         }
 
+        if (type is { } value && value.HasFlag(JsonSchemaType.String))
+        {
+            return JsonValue.Create(exampleString);
+        }
+
```
martincostello, line 26:
```suggestion
```

test/Swashbuckle.AspNetCore.SwaggerGen.Test/XmlComments/XmlCommentsSchemaFilterTests.cs
```diff
@@ -52,32 +52,38 @@ public void Apply_SetsDescription_FromPropertySummaryTag(
...
+        { typeof(XmlAnnotatedRecord), nameof(XmlAnnotatedRecord.StringPropertyWithUri), JsonSchemaTypes.String, "\"https://test.com/a?b=1\\u0026c=2\"", null },
+        { typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.StringPropertyWithNullExample), JsonSchemaTypes.String, "null", null },
+        { typeof(XmlAnnotatedRecord), nameof(XmlAnnotatedRecord.StringPropertyWithNullExample), JsonSchemaTypes.String, "null", null },
+        { typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.NullableStringPropertyWithNullExample), JsonSchemaTypes.String | JsonSchemaType.Null, "null",null },
+        { typeof(XmlAnnotatedRecord), nameof(XmlAnnotatedRecord.NullableStringPropertyWithNullExample), JsonSchemaTypes.String | JsonSchemaType.Null, "null",null },
+        { typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.NullableStringPropertyWithNotNullExample), JsonSchemaTypes.String | JsonSchemaType.Null, "\"example\"",null },
+        { typeof(XmlAnnotatedRecord), nameof(XmlAnnotatedRecord.NullableStringPropertyWithNotNullExample), JsonSchemaTypes.String | JsonSchemaType.Null, "\"example\"",null },
```
martincostello:
```suggestion
        { typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.BoolProperty), JsonSchemaTypes.Boolean, "true", null },
        { typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.IntProperty), JsonSchemaTypes.Integer, "10", "int32" },
        { typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.LongProperty), JsonSchemaTypes.Integer, "4294967295", "int64" },
        { typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.FloatProperty), JsonSchemaTypes.Number, "1.2", "float" },
        { typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.DoubleProperty), JsonSchemaTypes.Number, "1.25", "double" },
        { typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.DateTimeProperty), JsonSchemaTypes.String, "\"6/22/2022 12:00:00 AM\"", "date-time" },
        { typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.EnumProperty), JsonSchemaTypes.Integer, "2", "int32" },
        { typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.GuidProperty), JsonSchemaTypes.String, "\"d3966535-2637-48fa-b911-e3c27405ee09\"", "uuid" },
        { typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.StringProperty), JsonSchemaTypes.String, "\"Example for StringProperty\"", null },
        { typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.ObjectProperty), JsonSchemaTypes.Object, "{\n  \"prop1\": 1,\n  \"prop2\": \"foobar\"\n}", null },
        { typeof(XmlAnnotatedRecord), nameof(XmlAnnotatedRecord.BoolProperty), JsonSchemaTypes.Boolean, "true", null },
        { typeof(XmlAnnotatedRecord), nameof(XmlAnnotatedRecord.IntProperty), JsonSchemaTypes.Integer, "10", "int32"},
        { typeof(XmlAnnotatedRecord), nameof(XmlAnnotatedRecord.LongProperty), JsonSchemaTypes.Integer, "4294967295", "int64" },
        { typeof(XmlAnnotatedRecord), nameof(XmlAnnotatedRecord.FloatProperty), JsonSchemaTypes.Number, "1.2", "float" },
        { typeof(XmlAnnotatedRecord), nameof(XmlAnnotatedRecord.DoubleProperty), JsonSchemaTypes.Number, "1.25", "double" },
        { typeof(XmlAnnotatedRecord), nameof(XmlAnnotatedRecord.DateTimeProperty), JsonSchemaTypes.String, "\"6/22/2022 12:00:00 AM\"", "date-time" },
        { typeof(XmlAnnotatedRecord), nameof(XmlAnnotatedRecord.EnumProperty), JsonSchemaTypes.Integer, "2", "int32"},
        { typeof(XmlAnnotatedRecord), nameof(XmlAnnotatedRecord.GuidProperty), JsonSchemaTypes.String, "\"d3966535-2637-48fa-b911-e3c27405ee09\"", "uuid"},
        { typeof(XmlAnnotatedRecord), nameof(XmlAnnotatedRecord.StringProperty), JsonSchemaTypes.String, "\"Example for StringProperty\"", null },
        { typeof(XmlAnnotatedRecord), nameof(XmlAnnotatedRecord.ObjectProperty), JsonSchemaTypes.Object, "{\n  \"prop1\": 1,\n  \"prop2\": \"foobar\"\n}", null },
        { typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.StringPropertyWithUri), JsonSchemaTypes.String, "\"https://test.com/a?b=1\\u0026c=2\"", null },
        { typeof(XmlAnnotatedRecord), nameof(XmlAnnotatedRecord.StringPropertyWithUri), JsonSchemaTypes.String, "\"https://test.com/a?b=1\\u0026c=2\"", null },
        { typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.StringPropertyWithNullExample), JsonSchemaTypes.String, "null", null },
        { typeof(XmlAnnotatedRecord), nameof(XmlAnnotatedRecord.StringPropertyWithNullExample), JsonSchemaTypes.String, "null", null },
        { typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.NullableStringPropertyWithNullExample), JsonSchemaTypes.String | JsonSchemaType.Null, "null", null },
        { typeof(XmlAnnotatedRecord), nameof(XmlAnnotatedRecord.NullableStringPropertyWithNullExample), JsonSchemaTypes.String | JsonSchemaType.Null, "null", null },
        { typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.NullableStringPropertyWithNotNullExample), JsonSchemaTypes.String | JsonSchemaType.Null, "\"example\"", null },
        { typeof(XmlAnnotatedRecord), nameof(XmlAnnotatedRecord.NullableStringPropertyWithNotNullExample), JsonSchemaTypes.String | JsonSchemaType.Null, "\"example\"", null },
```

test/Swashbuckle.AspNetCore.SwaggerGen.Test/XmlComments/XmlCommentsSchemaFilterTests.cs
```diff
@@ -86,13 +92,14 @@ public void Apply_SetsExample_FromPropertyExampleTag(
         Type declaringType,
         string propertyName,
         JsonSchemaType schemaType,
-        string expectedExampleAsJson)
+        string expectedExampleAsJson,
+        string format)
```
martincostello: The inputs should be before the outputs (`expectedExampleAsJson`).

test/Swashbuckle.AspNetCore.SwaggerGen.Test/XmlComments/XmlCommentsSchemaFilterTests.cs
jgarciadelanoceda: The format was finally not needed see https://github.com/domaindrivendev/Swashbuckle.AspNetCore/pull/3803/changes/5e972dbabbaec1b57e6b1754f8595c3c1a20ce3e.

I had this first approach that was fine for almost every case there.. But the other solution suggested me to undo that thing:
https://github.com/domaindrivendev/Swashbuckle.AspNetCore/pull/3802#issuecomment-3925777799

test/Swashbuckle.AspNetCore.SwaggerGen.Test/XmlComments/XmlCommentsSchemaFilterTests.cs
```diff
@@ -55,29 +55,35 @@ public void Apply_SetsDescription_FromPropertySummaryTag(
...
         { typeof(XmlAnnotatedRecord), nameof(XmlAnnotatedRecord.StringPropertyWithNullExample), JsonSchemaTypes.String, "null" },
+        { typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.NullableStringPropertyWithNullExample), JsonSchemaTypes.String | JsonSchemaType.Null, "null" },
+        { typeof(XmlAnnotatedRecord), nameof(XmlAnnotatedRecord.NullableStringPropertyWithNullExample), JsonSchemaTypes.String | JsonSchemaType.Null, "null" },
+        { typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.NullableStringPropertyWithNotNullExample), JsonSchemaTypes.String | JsonSchemaType.Null, "\"example\""},
+        { typeof(XmlAnnotatedRecord), nameof(XmlAnnotatedRecord.NullableStringPropertyWithNotNullExample), JsonSchemaTypes.String | JsonSchemaType.Null, "\"example\""},
+        { typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.NullableIntPropertyWithNotNullExample), JsonSchemaTypes.Integer | JsonSchemaType.Null, "3" },
+        { typeof(XmlAnnotatedRecord), nameof(XmlAnnotatedRecord.NullableIntPropertyWithNotNullExample), JsonSchemaTypes.Integer | JsonSchemaType.Null, "3" }
```
martincostello:
```suggestion
        { typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.NullableStringPropertyWithNotNullExample), JsonSchemaTypes.String | JsonSchemaType.Null, "\"example\"" },
        { typeof(XmlAnnotatedRecord), nameof(XmlAnnotatedRecord.NullableStringPropertyWithNotNullExample), JsonSchemaTypes.String | JsonSchemaType.Null, "\"example\"" },
        { typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.NullableIntPropertyWithNotNullExample), JsonSchemaTypes.Integer | JsonSchemaType.Null, "3" },
        { typeof(XmlAnnotatedRecord), nameof(XmlAnnotatedRecord.NullableIntPropertyWithNotNullExample), JsonSchemaTypes.Integer | JsonSchemaType.Null, "3" },
```

test/Swashbuckle.AspNetCore.SwaggerGen.Test/XmlComments/XmlCommentsSchemaFilterTests.cs
```diff
@@ -55,29 +55,35 @@ public void Apply_SetsDescription_FromPropertySummaryTag(
...
+        { typeof(XmlAnnotatedRecord), nameof(XmlAnnotatedRecord.IntProperty), JsonSchemaTypes.Integer, "10"},
         { typeof(XmlAnnotatedRecord), nameof(XmlAnnotatedRecord.LongProperty), JsonSchemaTypes.Integer, "4294967295" },
         { typeof(XmlAnnotatedRecord), nameof(XmlAnnotatedRecord.FloatProperty), JsonSchemaTypes.Number, "1.2" },
         { typeof(XmlAnnotatedRecord), nameof(XmlAnnotatedRecord.DoubleProperty), JsonSchemaTypes.Number, "1.25" },
-        { typeof(XmlAnnotatedRecord), nameof(XmlAnnotatedRecord.DateTimeProperty), JsonSchemaTypes.String, "\"6/22/2022 12:00:00 AM\"" },
+        { typeof(XmlAnnotatedRecord), nameof(XmlAnnotatedRecord.DateTimeProperty), JsonSchemaTypes.String, "\"6/22/2022 12:00:00 AM\""},
         { typeof(XmlAnnotatedRecord), nameof(XmlAnnotatedRecord.EnumProperty), JsonSchemaTypes.Integer, "2" },
-        { typeof(XmlAnnotatedRecord), nameof(XmlAnnotatedRecord.GuidProperty), JsonSchemaTypes.String, "\"d3966535-2637-48fa-b911-e3c27405ee09\"" },
+        { typeof(XmlAnnotatedRecord), nameof(XmlAnnotatedRecord.GuidProperty), JsonSchemaTypes.String, "\"d3966535-2637-48fa-b911-e3c27405ee09\""},
```
martincostello:
```suggestion
        { typeof(XmlAnnotatedRecord), nameof(XmlAnnotatedRecord.GuidProperty), JsonSchemaTypes.String, "\"d3966535-2637-48fa-b911-e3c27405ee09\"" },
```

test/Swashbuckle.AspNetCore.SwaggerGen.Test/XmlComments/XmlCommentsSchemaFilterTests.cs
```diff
@@ -55,29 +55,35 @@ public void Apply_SetsDescription_FromPropertySummaryTag(
...
+        { typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.ObjectProperty), JsonSchemaTypes.Object, "{\n  \"prop1\": 1,\n  \"prop2\": \"foobar\"\n}"},
         { typeof(XmlAnnotatedRecord), nameof(XmlAnnotatedRecord.BoolProperty), JsonSchemaTypes.Boolean, "true" },
-        { typeof(XmlAnnotatedRecord), nameof(XmlAnnotatedRecord.IntProperty), JsonSchemaTypes.Integer, "10" },
+        { typeof(XmlAnnotatedRecord), nameof(XmlAnnotatedRecord.IntProperty), JsonSchemaTypes.Integer, "10"},
         { typeof(XmlAnnotatedRecord), nameof(XmlAnnotatedRecord.LongProperty), JsonSchemaTypes.Integer, "4294967295" },
         { typeof(XmlAnnotatedRecord), nameof(XmlAnnotatedRecord.FloatProperty), JsonSchemaTypes.Number, "1.2" },
         { typeof(XmlAnnotatedRecord), nameof(XmlAnnotatedRecord.DoubleProperty), JsonSchemaTypes.Number, "1.25" },
-        { typeof(XmlAnnotatedRecord), nameof(XmlAnnotatedRecord.DateTimeProperty), JsonSchemaTypes.String, "\"6/22/2022 12:00:00 AM\"" },
+        { typeof(XmlAnnotatedRecord), nameof(XmlAnnotatedRecord.DateTimeProperty), JsonSchemaTypes.String, "\"6/22/2022 12:00:00 AM\""},
```
martincostello:
```suggestion
        { typeof(XmlAnnotatedRecord), nameof(XmlAnnotatedRecord.DateTimeProperty), JsonSchemaTypes.String, "\"6/22/2022 12:00:00 AM\"" },
```

test/Swashbuckle.AspNetCore.SwaggerGen.Test/XmlComments/XmlCommentsSchemaFilterTests.cs
```diff
@@ -55,29 +55,35 @@ public void Apply_SetsDescription_FromPropertySummaryTag(
...
-        { typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.StringProperty), JsonSchemaTypes.String, "\"Example for StringProperty\"" },
-        { typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.ObjectProperty), JsonSchemaTypes.Object, "{\n  \"prop1\": 1,\n  \"prop2\": \"foobar\"\n}" },
+        { typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.GuidProperty), JsonSchemaTypes.String, "\"d3966535-2637-48fa-b911-e3c27405ee09\""},
+        { typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.StringProperty), JsonSchemaTypes.String, "\"Example for StringProperty\""},
+        { typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.ObjectProperty), JsonSchemaTypes.Object, "{\n  \"prop1\": 1,\n  \"prop2\": \"foobar\"\n}"},
         { typeof(XmlAnnotatedRecord), nameof(XmlAnnotatedRecord.BoolProperty), JsonSchemaTypes.Boolean, "true" },
-        { typeof(XmlAnnotatedRecord), nameof(XmlAnnotatedRecord.IntProperty), JsonSchemaTypes.Integer, "10" },
+        { typeof(XmlAnnotatedRecord), nameof(XmlAnnotatedRecord.IntProperty), JsonSchemaTypes.Integer, "10"},
```
martincostello:
```suggestion
        { typeof(XmlAnnotatedRecord), nameof(XmlAnnotatedRecord.IntProperty), JsonSchemaTypes.Integer, "10" },
```

test/Swashbuckle.AspNetCore.SwaggerGen.Test/XmlComments/XmlCommentsSchemaFilterTests.cs
```diff
@@ -55,29 +55,35 @@ public void Apply_SetsDescription_FromPropertySummaryTag(
...
+        { typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.DateTimeProperty), JsonSchemaTypes.String, "\"6/22/2022 12:00:00 AM\""},
         { typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.EnumProperty), JsonSchemaTypes.Integer, "2" },
-        { typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.GuidProperty), JsonSchemaTypes.String, "\"d3966535-2637-48fa-b911-e3c27405ee09\"" },
-        { typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.StringProperty), JsonSchemaTypes.String, "\"Example for StringProperty\"" },
-        { typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.ObjectProperty), JsonSchemaTypes.Object, "{\n  \"prop1\": 1,\n  \"prop2\": \"foobar\"\n}" },
+        { typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.GuidProperty), JsonSchemaTypes.String, "\"d3966535-2637-48fa-b911-e3c27405ee09\""},
+        { typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.StringProperty), JsonSchemaTypes.String, "\"Example for StringProperty\""},
+        { typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.ObjectProperty), JsonSchemaTypes.Object, "{\n  \"prop1\": 1,\n  \"prop2\": \"foobar\"\n}"},
```
martincostello:
```suggestion
        { typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.GuidProperty), JsonSchemaTypes.String, "\"d3966535-2637-48fa-b911-e3c27405ee09\"" },
        { typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.StringProperty), JsonSchemaTypes.String, "\"Example for StringProperty\"" },
        { typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.ObjectProperty), JsonSchemaTypes.Object, "{\n  \"prop1\": 1,\n  \"prop2\": \"foobar\"\n}" },
```

test/Swashbuckle.AspNetCore.SwaggerGen.Test/XmlComments/XmlCommentsSchemaFilterTests.cs
```diff
@@ -55,29 +55,35 @@ public void Apply_SetsDescription_FromPropertySummaryTag(
...
     {
         { typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.BoolProperty), JsonSchemaTypes.Boolean, "true" },
-        { typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.IntProperty), JsonSchemaTypes.Integer, "10" },
+        { typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.IntProperty), JsonSchemaTypes.Integer, "10"},
         { typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.LongProperty), JsonSchemaTypes.Integer, "4294967295" },
         { typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.FloatProperty), JsonSchemaTypes.Number, "1.2" },
-        { typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.DoubleProperty), JsonSchemaTypes.Number, "1.25" },
-        { typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.DateTimeProperty), JsonSchemaTypes.String, "\"6/22/2022 12:00:00 AM\"" },
+        { typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.DoubleProperty), JsonSchemaTypes.Number, "1.25"},
+        { typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.DateTimeProperty), JsonSchemaTypes.String, "\"6/22/2022 12:00:00 AM\""},
```
martincostello:
```suggestion
        { typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.DoubleProperty), JsonSchemaTypes.Number, "1.25" },
        { typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.DateTimeProperty), JsonSchemaTypes.String, "\"6/22/2022 12:00:00 AM\"" },
```

test/Swashbuckle.AspNetCore.SwaggerGen.Test/XmlComments/XmlCommentsSchemaFilterTests.cs
```diff
@@ -55,29 +55,35 @@ public void Apply_SetsDescription_FromPropertySummaryTag(
     public static TheoryData<Type, string, JsonSchemaType, string> Apply_SetsExample_FromPropertyExampleTag_Data() => new()
     {
         { typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.BoolProperty), JsonSchemaTypes.Boolean, "true" },
-        { typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.IntProperty), JsonSchemaTypes.Integer, "10" },
+        { typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.IntProperty), JsonSchemaTypes.Integer, "10"},
```
martincostello:
```suggestion
        { typeof(XmlAnnotatedType), nameof(XmlAnnotatedType.IntProperty), JsonSchemaTypes.Integer, "10" },
```

test/Swashbuckle.AspNetCore.SwaggerGen.Test/XmlComments/XmlCommentsSchemaFilterTests.cs
jgarciadelanoceda: Not needed

martincostello:
Once there's some user feedback that this _looks_ right (just to avoid too many releases trying to fix this bug) then can look at merging.

Seems reasonable to me.

pinkfloydx33:
@martincostello  

I think it's reasonable. If a user puts `null` as the example:
- they likely did it on purpose
- presumably didn't mean the _string_ `"null"`
  - If they did, there's (hopefully) billions of other words they can put there instead

(Anecdotally I once had a boss whose last name was "Null", so it's not totally unheard of outside of programming contexts.... but I'd personally avoid using common programming language terms in the example texts)

The only downside is that it _does_ allow you to put `null` on a clearly non-nullable schema. That could also creep in when changing something to non-nullable _later_ as its easy to accidently miss the example. Both of these would leave you with a slightly incorrect specification... But perhaps those are such just giant edge cases that it doesn't matter and/or are not worth the complexity against guarding. 

**I'd be A-OK if it was merged in this state.**

--- 

That said, I still don't quite understand why checking for `.HasFlag(JsonSchemaType.Null)` as a guard to prevent `null` for non-nullable values presented any issues. 

IOW why doesn't the following address all cases?

```diff
- if (string.Equals(exampleString, "null"))
+ if (string.Equals(exampleString, "null") && type is { } value && value.HasFlag(JsonSchemaType.Null))
```

Someone mentioned NRT and various options during schema-generation that you'd have to account for. But wouldn't all of those have been handled properly during  `GenerateSchemaForType` /`GenerateSchemaForMember`, leaving the schema with the appropriate type/null-flag based on project configuration / compiler options / etc. _before_ being processed by the XmlComment-related filters? 

Or am I overlooking something about that? I'm genuinely curious at this point.

But like I said above, I'd be fine if the PR was merged as-is given the edge-case nature of what we're discussing.

martincostello:
It's mainly that anecdotally IIRC we get users who don't use NRTs, but then we get users either side of the fence who complain that the nullability is wrong (_"it's not optional so it shouldn't be nullable"_, _"strings are technically reference types so it should be nullable"_, ...).

If we can, I'd rather avoid that can of worms by just making it _"null means `null`"_.

pinkfloydx33:
I think that's reasonable. I am good with as proposed

martincostello (approved):

PR #3887, opened 2026-04-09
Support HEAD requests

### LINKED ISSUE #2174: HEAD request to /swagger returns a 404
I'm using v6.1.4.

I have noticed in our monitoring platform that we are receiving a large number of `HEAD` requests to the API Explorer which are  being served a 404 response.

I can recreate this both in my current project or just by spinning up a webapi project, adding the swashbuckle nuget package and the standard code to `ConfigureServices`.

Is this a bug or am I missing some configuration to handle `HEAD` requests to the swagger UI?

TIA

snebjorn:
# Pull Request

Fixes #2174 

HEAD requests now return resource metadata
```http
HTTP/1.1 200 OK
Content-Length: 1234567
NO BODY
```

### DIFF
--- a/src/Swashbuckle.AspNetCore.ReDoc/ReDocMiddleware.cs
+++ b/src/Swashbuckle.AspNetCore.ReDoc/ReDocMiddleware.cs
@@ -10,6 +10,7 @@ namespace Swashbuckle.AspNetCore.ReDoc;
 
 internal sealed class ReDocMiddleware
 {
+    private static readonly string[] AllowedHttpMethods = [HttpMethods.Get, HttpMethods.Head];
     private static readonly string ReDocVersion = GetReDocVersion();
 
     private readonly RequestDelegate _next;
@@ -37,7 +38,7 @@ internal sealed class ReDocMiddleware
 
     public async Task Invoke(HttpContext httpContext)
     {
-        if (HttpMethods.IsGet(httpContext.Request.Method))
+        if (AllowedHttpMethods.Contains(httpContext.Request.Method, StringComparer.OrdinalIgnoreCase))
         {
             var path = httpContext.Request.Path.Value;
 
@@ -165,7 +166,12 @@ internal sealed class ReDocMiddleware
 
             SetHeaders(response, _options, etag);
 
-            await response.WriteAsync(text, Encoding.UTF8, cancellationToken);
+            response.ContentLength = Encoding.UTF8.GetByteCount(text);
+
+            if (!HttpMethods.IsHead(context.Request.Method))
+            {
+                await response.WriteAsync(text, Encoding.UTF8, cancellationToken);
+            }
         }
 
         static string GetETag(string text)
--- a/src/Swashbuckle.AspNetCore.ReDoc/ReDocMiddleware.cs
+++ b/src/Swashbuckle.AspNetCore.ReDoc/ReDocMiddleware.cs
@@ -10,7 +10,7 @@ namespace Swashbuckle.AspNetCore.ReDoc;
 
 internal sealed class ReDocMiddleware
 {
-    private static readonly string[] AllowedHttpMethods = [HttpMethods.Get, HttpMethods.Head];
+    private static readonly HashSet<string> AllowedHttpMethods = new(StringComparer.OrdinalIgnoreCase) { HttpMethods.Get, HttpMethods.Head };
     private static readonly string ReDocVersion = GetReDocVersion();
 
     private readonly RequestDelegate _next;
@@ -38,7 +38,7 @@ internal sealed class ReDocMiddleware
 
     public async Task Invoke(HttpContext httpContext)
     {
-        if (AllowedHttpMethods.Contains(httpContext.Request.Method, StringComparer.OrdinalIgnoreCase))
+        if (AllowedHttpMethods.Contains(httpContext.Request.Method))
         {
             var path = httpContext.Request.Path.Value;
 
@@ -166,11 +166,14 @@ internal sealed class ReDocMiddleware
 
             SetHeaders(response, _options, etag);
 
-            response.ContentLength = Encoding.UTF8.GetByteCount(text);
-
-            if (!HttpMethods.IsHead(context.Request.Method))
+            if (HttpMethods.IsGet(context.Request.Method))
             {
                 await response.WriteAsync(text, Encoding.UTF8, cancellationToken);
+            }
+            else if (HttpMethods.IsHead(context.Request.Method))
+            {
+                // HEAD response must have an empty body, but have correct Content-Length header
+                response.ContentLength = Encoding.UTF8.GetByteCount(text);
             }
         }
 
--- a/src/Swashbuckle.AspNetCore.Swagger/SwaggerMiddleware.cs
+++ b/src/Swashbuckle.AspNetCore.Swagger/SwaggerMiddleware.cs
@@ -12,6 +12,7 @@ namespace Swashbuckle.AspNetCore.Swagger;
 internal sealed class SwaggerMiddleware
 {
     private static readonly Encoding UTF8WithoutBom = new UTF8Encoding(false);
+    private static readonly string[] AllowedHttpMethods = [HttpMethods.Get, HttpMethods.Head];
 
     private readonly RequestDelegate _next;
     private readonly SwaggerOptions _options;
@@ -74,13 +75,15 @@ internal sealed class SwaggerMiddleware
                 filter(swagger, httpContext.Request);
             }
 
+            var isHeadRequest = HttpMethods.IsHead(httpContext.Request.Method);
+
             if (extension is ".yaml" or ".yml")
             {
-                await RespondWithSwaggerYaml(httpContext.Response, swagger);
+                await RespondWithSwaggerYaml(httpContext.Response, swagger, isHeadRequest);
             }
             else
             {
-                await RespondWithSwaggerJson(httpContext.Response, swagger);
+                await RespondWithSwaggerJson(httpContext.Response, swagger, isHeadRequest);
             }
         }
         catch (UnknownSwaggerDocument)
@@ -94,7 +97,7 @@ internal sealed class SwaggerMiddleware
         documentName = null;
         extension = null;
 
-        if (!HttpMethods.IsGet(request.Method))
+        if (!AllowedHttpMethods.Contains(request.Method, StringComparer.OrdinalIgnoreCase))
         {
             return false;
         }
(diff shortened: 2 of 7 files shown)


martincostello:
For now please remove the changes for OPTIONS and just scope this to HEAD.

We need to make sure any support for OPTIONS works correctly with CORS.

src/Swashbuckle.AspNetCore.ReDoc/ReDocMiddleware.cs
```diff
@@ -165,7 +166,12 @@ private async Task RespondWithFile(HttpContext context, string fileName)
 
             SetHeaders(response, _options, etag);
 
-            await response.WriteAsync(text, Encoding.UTF8, cancellationToken);
+            response.ContentLength = Encoding.UTF8.GetByteCount(text);
+
+            if (!HttpMethods.IsHead(context.Request.Method))
```
martincostello:
```suggestion
            if (HttpMethods.IsGet(context.Request.Method))
```

src/Swashbuckle.AspNetCore.ReDoc/ReDocMiddleware.cs
```diff
@@ -10,6 +10,7 @@ namespace Swashbuckle.AspNetCore.ReDoc;
 
 internal sealed class ReDocMiddleware
 {
+    private static readonly string[] AllowedHttpMethods = [HttpMethods.Get, HttpMethods.Head];
```
martincostello: We could make this an invariant `HashSet<string>`, then line 41 would just be `Contains()`.

src/Swashbuckle.AspNetCore.Swagger/SwaggerMiddleware.cs
```diff
@@ -12,6 +12,7 @@ namespace Swashbuckle.AspNetCore.Swagger;
 internal sealed class SwaggerMiddleware
 {
     private static readonly Encoding UTF8WithoutBom = new UTF8Encoding(false);
+    private static readonly string[] AllowedHttpMethods = [HttpMethods.Get, HttpMethods.Head];
```
martincostello: Same here with the `HashSet<T>`.

src/Swashbuckle.AspNetCore.ReDoc/ReDocMiddleware.cs
```diff
@@ -165,7 +166,12 @@ private async Task RespondWithFile(HttpContext context, string fileName)
 
             SetHeaders(response, _options, etag);
 
-            await response.WriteAsync(text, Encoding.UTF8, cancellationToken);
+            response.ContentLength = Encoding.UTF8.GetByteCount(text);
```
martincostello: Maybe only do this for HEAD so we don't need to compute the length for GETs too?

src/Swashbuckle.AspNetCore.Swagger/SwaggerMiddleware.cs
```diff
@@ -140,11 +143,15 @@ private async Task RespondWithSwaggerJson(HttpResponse response, OpenApiDocument
 
         response.StatusCode = 200;
         response.ContentType = "application/json;charset=utf-8";
+        response.ContentLength = UTF8WithoutBom.GetByteCount(json);
```
martincostello: Maybe only do this for HEAD so we don't need to compute the length for GETs too?

src/Swashbuckle.AspNetCore.Swagger/SwaggerMiddleware.cs
```diff
@@ -159,8 +166,12 @@ private async Task RespondWithSwaggerYaml(HttpResponse response, OpenApiDocument
 
         response.StatusCode = 200;
         response.ContentType = "text/yaml;charset=utf-8";
+        response.ContentLength = UTF8WithoutBom.GetByteCount(yaml);
```
martincostello: +1

src/Swashbuckle.AspNetCore.SwaggerUI/SwaggerUIMiddleware.cs
martincostello, line 1: Same comments here.

test/Swashbuckle.AspNetCore.IntegrationTests/ReDocIntegrationTests.cs
```diff
@@ -26,6 +26,19 @@ public async Task RoutePrefix_RedirectsToIndexUrl()
...
+    [Fact]
+    public async Task IndexUrl_HeadRequest_ReturnsMetadata()
+    {
+        var site = new TestSite(typeof(ReDocApp.Startup), outputHelper);
+        using var client = site.BuildClient();
+        using var request = new HttpRequestMessage(HttpMethod.Head, "/api-docs/index.html");
+        using var response = await client.SendAsync(request, TestContext.Current.CancellationToken);
+
+        Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+        Assert.True(response.Content.Headers.ContentLength > 0, $"Content-Length should be greater than 0, but was {response.Content.Headers.ContentLength}");
```
martincostello:
```suggestion
        Assert.True(response.Content.Headers.ContentLength > 0, "Content-Length should not be zero.");
```

test/Swashbuckle.AspNetCore.IntegrationTests/SwaggerIntegrationTests.cs
```diff
@@ -51,7 +53,20 @@ public async Task SwaggerEndpoint_ReturnsNotFound_IfUnknownSwaggerDocument()
...
+    [Fact]
+    public async Task SwaggerEndpoint_ReturnsMetadata_ForHeadRequest()
+    {
+        var testSite = new TestSite(typeof(Basic.Startup), outputHelper);
+        using var client = testSite.BuildClient();
+        using var request = new HttpRequestMessage(HttpMethod.Head, "/swagger/v1/swagger.json");
+        using var response = await client.SendAsync(request, TestContext.Current.CancellationToken);
+
+        Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+        Assert.True(response.Content.Headers.ContentLength > 0, $"{HeaderNames.ContentLength} should be greater than 0, but was {response.Content.Headers.ContentLength}");
```
martincostello:
```suggestion
        Assert.True(response.Content.Headers.ContentLength > 0, "Content-Length should not be zero.");
```

test/Swashbuckle.AspNetCore.IntegrationTests/SwaggerUIIntegrationTests.cs
```diff
@@ -1,6 +1,7 @@
 ﻿using System.IO.Compression;
 using System.Net;
 using System.Security.Cryptography;
+using Microsoft.Net.Http.Headers;
```
martincostello:
```suggestion
```

test/Swashbuckle.AspNetCore.IntegrationTests/SwaggerUIIntegrationTests.cs
```diff
@@ -53,6 +54,23 @@ public async Task RoutePrefix_RedirectsToPathRelativeIndexUrl(
...
+        Type startupType,
+        string requestPath)
+    {
+        var site = new TestSite(startupType, outputHelper);
+        using var client = site.BuildClient();
+        using var request = new HttpRequestMessage(HttpMethod.Head, requestPath);
+        using var response = await client.SendAsync(request, TestContext.Current.CancellationToken);
+
+        Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+        Assert.True(response.Content.Headers.ContentLength > 0, $"{HeaderNames.ContentLength} should be greater than 0, but was {response.Content.Headers.ContentLength}");
```
martincostello:
```suggestion
        Assert.True(response.Content.Headers.ContentLength > 0, "Content-Length should not be be 0.");
```

martincostello (commented):
Looks good overall - just some minor suggestions.

src/Swashbuckle.AspNetCore.ReDoc/ReDocMiddleware.cs
snebjorn: GET needs the length as well. Wouldn't it just be calculated later in the pipeline anyway?

src/Swashbuckle.AspNetCore.ReDoc/ReDocMiddleware.cs
snebjorn: can you elaborate. It doesn't seem to change line 41?

```cs
private static readonly HashSet<string> AllowedHttpMethods = [HttpMethods.Get, HttpMethods.Head];

if (AllowedHttpMethods.Contains(httpContext.Request.Method, StringComparer.OrdinalIgnoreCase))
```

martincostello: I was thinking something similar to:

```csharp
HashSet<string> AllowedHttpMethods = new(StringComparer.OrdinalIgnoreCase) { HttpMethods.Get, HttpMethods.Head };

if (AllowedHttpMethods.Contains(httpContext.Request.Method))
```

src/Swashbuckle.AspNetCore.ReDoc/ReDocMiddleware.cs
martincostello: Either it doesn't _need_ it, or it's already being computed anyway, as it works today.

If it's already being calculated, then we should avoid doing it again, unless whatever does calculate it already is smart enough to not re-calculate it if it's already set.

src/Swashbuckle.AspNetCore.ReDoc/ReDocMiddleware.cs
snebjorn: Didn't know that trick. That's neat!

src/Swashbuckle.AspNetCore.ReDoc/ReDocMiddleware.cs
snebjorn: I'm assuming the asp net pipeline sets the content length if it's null. But I'll change it

test/Swashbuckle.AspNetCore.IntegrationTests/SwaggerIntegrationTests.cs
```diff
@@ -1,7 +1,9 @@
 using System.Globalization;
+using System.Net;
 using System.Reflection;
 using System.Text;
 using System.Text.Json;
+using Microsoft.Net.Http.Headers;
```
martincostello: I think this is redundant now?

```suggestion
```

martincostello (approved):

martincostello:
Thanks!

PR #3912, opened 2026-04-18
Fix conflicting required+nullable schema when only NonNullableReferen…

### LINKED ISSUE #3297: [Bug]: When using NonNullableReferenceTypesAsRequired, both required and nullable exist
### Describe the bug

Image

### Expected behavior

Remove the summary field "nullable": true configuration

### Actual behavior

_No response_

### Steps to reproduce

_No response_

### Exception(s) (if any)

Use the net9 default webapi template
Add or modify the following configuration
```
builder.Services.AddSwaggerGen(c => c.NonNullableReferenceTypesAsRequired());

app.UseSwagger();
app.UseSwaggerUI();
```

Remove the ? in front of the Summary attribute.
```
public class WeatherForecast
{
}
```

7.3.1

### .NET Version

net9

### Anything else?

_No response_

KitKeen:
## Fix conflicting `required` + `nullable` schema when only `NonNullableReferenceTypesAsRequired` is enabled

Fixes #3297

### Problem

When `NonNullableReferenceTypesAsRequired()` is enabled **without** `SupportNonNullableReferenceTypes()`, non-nullable reference-type properties end up with both `"required": true` and `"nullable": true` in the generated schema.

### Root cause

`IsNullable` only checks `SupportNonNullableReferenceTypes`:

```csharp
if (_generatorOptions.SupportNonNullableReferenceTypes)
{
    nullable &= !memberInfo.IsNonNullableReferenceType();
}
```

So when only `-AsRequired` is set, the property gets added to `required` (via `GenerateSchemaForMembers`) but `nullable` is never cleared.

### Fix

```csharp
if (_generatorOptions.SupportNonNullableReferenceTypes || _generatorOptions.NonNullableReferenceTypesAsRequired)
{
    nullable &= !memberInfo.IsNonNullableReferenceType();
}
```

### Before / after

`builder.Services.AddSwaggerGen(c => c.NonNullableReferenceTypesAsRequired());`

```csharp
public class WeatherForecast
{
}
```

**Before:**
```json
"summary": { "type": "string", "nullable": true }
```

**After:**
```json
"summary": { "type": "string" }
```

### Tests

Added `GenerateSchema_NonNullableReferenceTypesAsRequired_DoesNotMarkPropertyAsNullable` for both `TypeWithNullableContextAnnotated` and `TypeWithNullableContextNotAnnotated`. Fails on `main`, passes after the fix. Full suite 720/720 on `net10.0`, clean build on `net8.0`/`net9.0`/`net10.0`.

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/SchemaGenerator.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/SchemaGenerator.cs
@@ -139,7 +139,7 @@ public class SchemaGenerator(
     {
         var nullable = dataProperty.IsNullable && requiredAttribute == null;
 
-        if (_generatorOptions.SupportNonNullableReferenceTypes)
+        if (_generatorOptions.SupportNonNullableReferenceTypes || _generatorOptions.NonNullableReferenceTypesAsRequired)
         {
             nullable &= !memberInfo.IsNonNullableReferenceType();
         }
--- a/test/Swashbuckle.AspNetCore.SwaggerGen.Test/SchemaGenerator/JsonSerializerSchemaGeneratorTests.cs
+++ b/test/Swashbuckle.AspNetCore.SwaggerGen.Test/SchemaGenerator/JsonSerializerSchemaGeneratorTests.cs
@@ -1127,6 +1127,23 @@ public class JsonSerializerSchemaGeneratorTests
     }
 
     [Theory]
+    [InlineData(typeof(TypeWithNullableContextAnnotated), nameof(TypeWithNullableContextAnnotated.SubTypeWithOneNonNullableContent), nameof(TypeWithNullableContextAnnotated.NonNullableString))]
+    [InlineData(typeof(TypeWithNullableContextNotAnnotated), nameof(TypeWithNullableContextNotAnnotated.SubTypeWithOneNonNullableContent), nameof(TypeWithNullableContextNotAnnotated.NonNullableString))]
+    public void GenerateSchema_NonNullableReferenceTypesAsRequired_DoesNotMarkPropertyAsNullable(
+        Type declaringType,
+        string subType,
+        string propertyName)
+    {
+        var subject = Subject(c => c.NonNullableReferenceTypesAsRequired = true);
+        var schemaRepository = new SchemaRepository();
+
+        subject.GenerateSchema(declaringType, schemaRepository);
+
+        var propertySchema = schemaRepository.Schemas[subType].Properties[propertyName];
+        AssertIsNullable(propertySchema, expected: false);
+    }
+
+    [Theory]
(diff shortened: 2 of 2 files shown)


martincostello:
I need to think about this more before doing a proper review on this PR.

Members can still be _required_ yet allow `null`s (i.e. property must be present, may not have a value).

For example:

```csharp
public class SomeType
{
    public required string? A { get; set; }

    public required string? C { get; set; }
}
```

```jsonc
// Bad payload, c is missing
{
  "a": "b"
}
```

```jsonc
// Good payload, c is specified as explicitly null
{
  "a": "b",
  "c": null
}
```

```jsonc
// Good payload, c is specified with a value
{
  "a": "b",
  "c": "d"
}
```

Any changes here need to support different scenarios people may have when nullable reference types are and aren't enabled and when members are required or not.

The defaults should where possible match expectations, but at the same type we don't want to introduce any breaking changes or make different opinions users may have on how to treat required and/or nullable reference types together.

KitKeen:
Martin, I appreciate your review!

  You're right that **required + nullable: true** is a valid OpenAPI combination
  and should be preserved. The fix handles this correctly because the guard
  `memberInfo.IsNonNullableReferenceType()` distinguishes between the two cases:

  public class SomeType
  {
      public string  A { get; set; }   // non-nullable → IsNonNullableReferenceType() = true
      public string? B { get; set; }   // nullable     → IsNonNullableReferenceType() = false
  }

  With NonNullableReferenceTypesAsRequired enabled:

   property required  nullable correctness
   string A    true       false    ✅ non-nullable, must be present and have a value
   string? B   false      true     ✅ nullable, not required

  Your required string? example is also covered:

  public class SomeType
  {
      public required string? A { get; set; }
  }

  Here IsNonNullableReferenceType() returns false for string?, so nullable
  is not cleared- the schema correctly reflects required: true, nullable: true.

  The fix only clears nullable when both conditions are true:
  а) NonNullableReferenceTypesAsRequired is enabled
  b) The member is genuinely non-nullable (IsNonNullableReferenceType() = true)


I've also added a test
GenerateSchema_NonNullableReferenceTypesAsRequired_PreservesNullable_IfPropertyHasRequiredKeywordAndIsNullable covering exactly this scenario required string? stays required + nullable, required string stays required + not nullable.

KitKeen:
take your time reviewing it, I'd rather listen to all of the conserns before merging.

KitKeen:
Hi @martincostello- CI hasn't run on this PR yet (the "first-time contributor needs workflow approval" gate). Could you kick it off when you have a moment? Happy to address whatever the checks surface.

KitKeen:
@martincostello just a kind reminder!

martincostello:
Test are failing.

KitKeen:
@martincostello pushed 78f5b52 - added the missing `Assert.NotNull(schema.Required);` before the `.ToArray()` call so the new test compiles under `#nullable enable` on all three target frameworks. Could you re-approve the CI workflow when you have a moment? Thanks!

martincostello (approved):

PR #3976, opened 2026-06-01
Change nullability handling for schemas with anyOf, allOf, oneOf or type = null

### LINKED ISSUE #3936: [Bug]: Dictionary with values of type object generates invalid open api schema
### Describe the bug

The schema generation for a value of type `IDictionary<XXX, object>` or `IImmutableDictionary<XXX, object>` puts the wrong schema type on the property values. The type of the properties of the dictionary is just 'null' instead of a list with object and 'null'.

### Expected behavior

The property type of a `IDictionary<XXX, object>` or `IImmutableDictionary<XXX, object>` should be object or at worst object and 'null'

### Actual behavior

System.Object data contract is dynamic. A dynamic data contract as json schema type "unknown". An unknown json schema type within a data contract will produce an open api schema whose type is null. Within the dictionary property open api schema generator, you attempt to add nullable to the types of the open api schema, and you replace null with Nullable if the value is null.

As a result, the type of the properties of the dictionary is just 'null' instead of a list with object and 'null'.

### Steps to reproduce

_No response_

### Exception(s) (if any)

_No response_

10.1.7

### .NET Version

10

### Anything else?

I think this is a regression. Before, `IDictionary<XXX, object>` would have had non nullable properties and as such the issue didn't appear

ldeluigi:
# Pull Request

Fixes #3936 

In order to fix #3936 I needed to check whethere the nullability was being applied to a schema of one of these kinds:
* a schema with an `allOf` property configured: I restored the previous behaviour of having type: Null instead of type: null. This is a niche scenario.
* a schema with an `anyOf` or `oneOf` property configured: in this case, according to the open api specification, nullability should be expressed with an additional anyOf/oneOf type with type: Null
* a schema with a type which is not null. In this case I restored the previous behaviour of or-ing the type with Null
*   a schema with a type which is null. The previous behaviour made it be just Null, which is not valid. Instead, I translate the type: null meaning of AnyType to the more explicit type: Boolean | Integer | Number | String | Array | Object befor or-ing Null


ldeluigi:
@martincostello could you enable the CI again please?

Timovzl:
@ldeluigi What would be the most straightforward way for me to check if your PR fixes the issue for me?

ldeluigi:
> @ldeluigi What would be the most straightforward way for me to check if your PR fixes the issue for me?

Either clone it and add your case as a unit test for swagger gen and see the result, or debug from your side where is the problem in the current version of the library and then look at the changed lines and see if I changed those lines related to your issue

martincostello:
Sorry for the delay in reviewing - been busy with other things.

As this area is quite sensitive in terms of potential breaking changes (we've accidentally broken things in this area in the past) I asked Claude Code to do a review of the changes for possible issues or conflicts with the specification.

Please review the output below and see if any of the findings are relevant to these changes. The "Concerns" section is the most relevant, but I've included all the output for context.

---

## Review of PR #3976 — Nullability handling for `anyOf`/`allOf`/`oneOf`/`type:null` schemas

### Verdict
The fix is **valid and an improvement**. It correctly resolves #3936 (the `IDictionary<…, object?>` → `type: 'null'` bug) and the new `anyOf`/`oneOf` handling is the idiomatic JSON Schema / OpenAPI 3.1 form. CI is green on all platforms. There are some **coverage gaps and one pre-existing semantic wart** worth flagging, but no regressions.

### Is the fix valid?

The core change in `SchemaGenerator.SetNullable` replaces the broken `schema.Type ??= Null; schema.Type |= Null` with branch-by-shape handling. I validated each branch by actually serializing the resulting schemas with Microsoft.OpenApi 2.7.6 (the pinned version) in **both** 3.0 and 3.1:

| Case | Old 3.1 output | New 3.1 output | New 3.0 output |
|------|----------------|----------------|----------------|
| `object?` (no type) | `type: "null"` ❌ (only null allowed — the bug) | `type: ["null","boolean","integer",…,"array"]` ✅ | `nullable: true` ✅ |
| nullable `anyOf` | `type: "null"` + anyOf | `anyOf: [..., {type:null}]` ✅ | `nullable: true` (+ flattened anyOf) ✅ |
| nullable `oneOf` | same as anyOf | `oneOf: [..., {type:null}]` ✅ | ✅ |
| `int?` etc. | `integer`+`null` | `integer`+`null` (unchanged) ✅ | `integer`+`nullable` ✅ |

So the headline bug is genuinely fixed, and the union approach matches exactly the "Any Type" definition the issue author cited from the OpenAPI spec. Importantly, **the default serialization version in Swashbuckle is OpenAPI 3.0** (`SwaggerOptions.OpenApiVersion = OpenApi3_0`), and I confirmed the verbose union cleanly downconverts to `nullable: true` with no type — so 3.0 documents are unaffected/correct.

The normal primitive path (`int?` → `integer | null`) is provably unchanged: those types skip all the `anyOf/oneOf/allOf` branches, fall into the `else` (`??=` is a no-op since `Type` is already set), then `|= Null`. ✅

### Concerns

1. **`allOf` branch produces a contradictory 3.1 schema.** `schema.Type ??= JsonSchemaType.Null` followed by `|= Null` yields `type: "null"` **alongside** `allOf: [...]` — i.e. "must be null AND must satisfy the subschemas," which is unsatisfiable in 3.1. (In 3.0 it downconverts harmlessly to `allOf` + `nullable: true`.) The PR explicitly says this *preserves pre-existing behavior* for a niche case, and I confirmed the old code produced the same thing — so it's **not a regression**, but it's not actually "fixed" either. The correct 3.1 form would wrap in `anyOf: [{allOf:[…]}, {type:null}]`. Acceptable to defer, but worth calling out.

2. **The `allOf` branch has no dedicated test.** The two new `[Fact]` tests cover `anyOf` and `oneOf`; `DictionaryWrappersData` covers the `object?` union. There is no test that drives `SetNullable` on a schema with `AllOf.Count > 0`. Codecov reports patch coverage green, so an existing test likely grazes it, but there's no explicit assertion of the (admittedly questionable) `type:null + allOf` output. Minor: the `allOf` pattern binds `{ Count: > 0 } allOf` but never uses `allOf` (the `anyOf`/`oneOf` bindings are used).

3. **`AnnotationsSchemaFilter` was only half-aligned.** It got the union-type fix but **not** the `anyOf`/`oneOf`/`allOf` handling, so `[SwaggerSchema(Nullable = true)]` on an `anyOf` custom schema would OR the union into `Type` rather than adding `{type:null}` to the `anyOf` list — inconsistent with `SchemaGenerator`. No test covers this path either. Edge case, but the two code paths now diverge.

4. **No test asserts the serialized document.** All tests assert in-memory `JsonSchemaType` flags / `anyOf` counts. Given the whole point is spec-valid output, a snapshot test of the emitted 3.0 *and* 3.1 JSON would be more convincing. (I verified this manually above; it's fine — just not locked in by a test.)

### Breaking changes
No API breaks. There are **document-shape changes** for 3.1 consumers: `object?`, nullable `anyOf`, and nullable `oneOf` members now emit different (correct) schemas than before. These are bug fixes replacing invalid/misleading output, and 3.0 output is unchanged — so I'd classify this as a non-breaking correctness fix, though anyone snapshotting 3.1 docs will see diffs.

The test refactor (moving `ModelOfA/B/C`, `TypeWithRequiredProperties`, etc. into `TestSupport/Fixtures`) is clean — no other references to the old nested types and no name collisions (CI passes).

ldeluigi:
@martincostello I updated the code with copilot

src/Swashbuckle.AspNetCore.Annotations/AnnotationsSchemaFilter.cs
```diff
@@ -102,10 +102,36 @@ private static void ApplySchemaAttribute(IOpenApiSchema schema, SwaggerSchemaAtt
...
+                {
+                    anyOf.Add(new OpenApiSchema { Type = JsonSchemaType.Null });
+                }
+                else if (concrete.OneOf is { Count: > 0 } oneOf)
+                {
+                    oneOf.Add(new OpenApiSchema { Type = JsonSchemaType.Null });
+                }
+                else
+                {
+                    concrete.Type ??= JsonSchemaType.Boolean | JsonSchemaType.Integer | JsonSchemaType.Number | JsonSchemaType.String | JsonSchemaType.Object | JsonSchemaType.Array;
```
martincostello: This doesn't seem right - making it be every possible type?

src/Swashbuckle.AspNetCore.Annotations/AnnotationsSchemaFilter.cs
```diff
@@ -102,10 +102,36 @@ private static void ApplySchemaAttribute(IOpenApiSchema schema, SwaggerSchemaAtt
         if (schemaAttribute.NullableFlag is { } nullable)
         {
             // See https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/3387
+            // See https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/3936
             if (nullable)
             {
-                concrete.Type ??= JsonSchemaType.Null;
-                concrete.Type |= JsonSchemaType.Null;
+                if (concrete.AllOf is { Count: > 0 } allOf)
```
martincostello: I'm not sure this logic is correct for all OpenAPI document versions.

src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/SchemaGenerator.cs
```diff
@@ -632,10 +632,35 @@ private System.Text.Json.Nodes.JsonNode GenerateDefaultValue(
     private static void SetNullable(OpenApiSchema schema, bool nullable)
     {
         // See https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/3387
+        // See https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/3936
         if (nullable)
         {
-            schema.Type ??= JsonSchemaType.Null;
-            schema.Type |= JsonSchemaType.Null;
+            if (schema.AllOf is { Count: > 0 } allOf)
```
martincostello: Same comments here.

test/Swashbuckle.AspNetCore.TestSupport/Fixtures/JsonSerializerSchemaGeneratorTestTypes.cs
martincostello, line 1: There's a lot of types being declared here that the tests don't seem to be actually using?

test/Swashbuckle.AspNetCore.SwaggerGen.Test/SchemaGenerator/JsonSerializerSchemaGeneratorTests.cs
```diff
@@ -1537,4 +1600,33 @@ private static void AssertIsNullable(IOpenApiSchema schema, bool expected = true
...
+        };
+
+        if (version == OpenApiSpecVersion.OpenApi3_1)
+        {
+            document.SerializeAsV31(jsonWriter);
+        }
+        else
+        {
+            document.SerializeAsV3(jsonWriter);
+        }
```
martincostello: Use the `SerializeAs(OpenApiSpecVersion version, IOpenApiWriter writer)` overload.

test/Swashbuckle.AspNetCore.SwaggerGen.Test/SchemaGenerator/JsonSerializerSchemaGeneratorTests.cs
```diff
@@ -585,6 +585,69 @@ public void GenerateSchema_SupportsOption_CustomTypeMappings(
...
+        Assert.Equal(2, anyOf3_1.GetArrayLength());
+        Assert.Contains(anyOf3_1.EnumerateArray(), element => element.TryGetProperty("type", out var type) && type.GetString() == "null");
+        Assert.Contains(anyOf3_1.EnumerateArray(), element => element.TryGetProperty("allOf", out _));
+
+        var serializedAs3_0 = SerializeSchemaInDocument(propertySchema, OpenApiSpecVersion.OpenApi3_0);
+        using var json3_0 = JsonDocument.Parse(serializedAs3_0);
+        var schema3_0 = json3_0.RootElement.GetProperty("components").GetProperty("schemas").GetProperty("Test");
+        Assert.True(schema3_0.TryGetProperty("nullable", out var nullable3_0));
+        Assert.True(nullable3_0.GetBoolean());
+        Assert.True(schema3_0.TryGetProperty("allOf", out _) || schema3_0.TryGetProperty("anyOf", out _));
```
martincostello: Don't test multiple versions in the same test, split it out.

martincostello:
> @martincostello I updated the code with copilot

And do you understand and vouch for its output?

src/Swashbuckle.AspNetCore.Annotations/AnnotationsSchemaFilter.cs
ldeluigi: Having null in the type field is equivalent to being of any type EXCEPT 'Null', so if we want to have a schema that can be of any type AND 'Null', we need to list them all and add 'Null' to the list. The old code was missing this in particular. It was putting 'Null' alone, creating schemas that can only be of type 'Null' (not object not arrays...).

ldeluigi:
> > @martincostello I updated the code with copilot
> 
> And do you understand and vouch for its output?

Today I should have the time to address all the issues, your doubts and to assess the quality of its output. I just didn't have it before this weekend.

src/Swashbuckle.AspNetCore.Annotations/AnnotationsSchemaFilter.cs
ldeluigi: https://swagger.io/docs/specification/v3_0/data-models/data-types/#any-type

ldeluigi: Nevermind, apparently type: null allows for the value to be null, I'm updating the code accordingly

src/Swashbuckle.AspNetCore.Annotations/AnnotationsSchemaFilter.cs
ldeluigi: The in-memory model stays version-agnostic; Microsoft.OpenApi does the down-conversion. I'm adding a regression test.

test/Swashbuckle.AspNetCore.TestSupport/Fixtures/JsonSerializerSchemaGeneratorTestTypes.cs
ldeluigi: I'm removing the unused ones

ldeluigi:
@martincostello Pushed. Summary of the changes and one correction.

**"Making it be every possible type?"** — you were right. My earlier reply on that thread was wrong: I claimed an absent `type` means "any type *except* null", so nullability had to be expressed by enumerating every type. Per JSON Schema 2020-12 §6.1.1 a schema with no `type` already validates every type *including* `null`, so the union was redundant. Now the null flag is only OR-ed in when a concrete type is present, which still fixes #3936 without inventing a verbose schema.

**"Not sure this logic is correct for all OpenAPI document versions"** — this one caught a real bug, via CI. The `allOf` → `anyOf: [{allOf:[…]}, {type:null}]` rewrite (added in response to the review you pasted) breaks client generators: Kiota can no longer resolve the `$ref` through the extra nesting and degrades the property to an untyped value, in **both** 3.0 and 3.1:

```diff
- public Models.TodoPriority? Priority { get; set; }   // nullable enum ref
+ public UntypedNode? Priority { get; set; }

- public Models.Child? Child { get; set; }             // nullable model ref
+ public UntypedNode? Child { get; set; }
```

So I've reverted that rewrite — `allOf` is left intact with the null type flag, as this PR originally had it. The 3.1 `allOf` + `type: "null"` wart therefore remains, which the review you pasted classified as pre-existing and acceptable to defer. The `anyOf`/`oneOf` branches only append a null member to an existing composition, so their output is unaffected either way. The re-blessed snapshots are reverted with them.

Also addressed: `SerializeAs(version, writer)` overload, split the multi-version test, removed the unused fixture types (and added the missing `anyOf`/`oneOf` coverage so the composition fixtures are actually used). Added a regression test for the `IDictionary<string, object?>` case from #3936, plus a guard asserting the serialized 3.0 output keeps `allOf` unwrapped so this can't regress silently.

test/Swashbuckle.AspNetCore.SwaggerGen.Test/SchemaGenerator/JsonSerializerSchemaGeneratorTests.cs
```diff
@@ -585,6 +585,144 @@ public void GenerateSchema_SupportsOption_CustomTypeMappings(
...
+    }
+
+    // The allOf must survive down-conversion unwrapped, otherwise client generators cannot
+    // resolve the referenced schema and degrade the property to an untyped value.
+    [Fact]
+    public void GenerateSchema_SerializesNullableAllOfSchema_AsNullableAllOf_ForOpenApi3_0()
+    {
+        var propertySchema = GenerateNullableAllOfPropertySchema();
+
+        var serialized = SerializeSchemaInDocument(propertySchema, OpenApiSpecVersion.OpenApi3_0);
```
martincostello, line 767: Could you add any tests as appropriate for Swagger 2.0 as well please?

src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/SchemaGenerator.cs
```diff
@@ -632,10 +632,31 @@ private System.Text.Json.Nodes.JsonNode GenerateDefaultValue(
...
+                schema.Type |= JsonSchemaType.Null;
+            }
+            else if (schema.AnyOf is { Count: > 0 } anyOf)
+            {
+                anyOf.Add(new OpenApiSchema { Type = JsonSchemaType.Null });
+            }
+            else if (schema.OneOf is { Count: > 0 } oneOf)
+            {
+                oneOf.Add(new OpenApiSchema { Type = JsonSchemaType.Null });
+            }
```
martincostello, line 653: Only add if not already present?

A property like this might otherwise end up with it being added twice:

```csharp
[SwaggerSchema(Nullable = true)]
public CustomAnyOfType? Property { get; set; }
```

src/Swashbuckle.AspNetCore.Annotations/AnnotationsSchemaFilter.cs
```diff
@@ -102,10 +102,31 @@ private static void ApplySchemaAttribute(IOpenApiSchema schema, SwaggerSchemaAtt
...
+                    concrete.Type |= JsonSchemaType.Null;
+                }
+                else if (concrete.AnyOf is { Count: > 0 } anyOf)
+                {
+                    anyOf.Add(new OpenApiSchema { Type = JsonSchemaType.Null });
+                }
+                else if (concrete.OneOf is { Count: > 0 } oneOf)
+                {
+                    oneOf.Add(new OpenApiSchema { Type = JsonSchemaType.Null });
+                }
```
martincostello, line 123: Same here.

src/Swashbuckle.AspNetCore.Annotations/AnnotationsSchemaFilter.cs
```diff
@@ -144,4 +144,16 @@ private static void ApplySchemaAttribute(IOpenApiSchema schema, SwaggerSchemaAtt
             concrete.Title = title;
         }
     }
+
+    private static void AddNullSchema(IList<IOpenApiSchema> schemas)
+    {
+        // Only add a null member if the composition does not already allow null,
+        // otherwise a schema that is nullable through multiple mechanisms (e.g. a
+        // nullable reference type also annotated with [SwaggerSchema(Nullable = true)])
+        // would end up with the null type listed more than once.
```
martincostello:
```suggestion
```

src/Swashbuckle.AspNetCore.Annotations/AnnotationsSchemaFilter.cs
```diff
@@ -144,4 +144,16 @@ private static void ApplySchemaAttribute(IOpenApiSchema schema, SwaggerSchemaAtt
             concrete.Title = title;
         }
     }
+
+    private static void AddNullSchema(IList<IOpenApiSchema> schemas)
```
martincostello: Rename to `TryAddNullSchema`

src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/SchemaGenerator.cs
```diff
@@ -664,6 +664,18 @@ private static void SetNullable(OpenApiSchema schema, bool nullable)
...
+    {
+        // Only add a null member if the composition does not already allow null,
+        // otherwise a schema that is nullable through multiple mechanisms (e.g. a
+        // nullable reference type also annotated with [SwaggerSchema(Nullable = true)])
+        // would end up with the null type listed more than once.
+        if (!schemas.Any(static s => s.Type is { } type && type.HasFlag(JsonSchemaType.Null)))
+        {
+            schemas.Add(new OpenApiSchema { Type = JsonSchemaType.Null });
+        }
+    }
```
martincostello: Same comments here.

test/Swashbuckle.AspNetCore.SwaggerGen.Test/SchemaGenerator/JsonSerializerSchemaGeneratorTests.cs
```diff
@@ -723,6 +775,23 @@ public void GenerateSchema_SerializesNullableAllOfSchema_AsNullableAllOf_ForOpen
         Assert.False(schema.TryGetProperty("anyOf", out _));
     }
 
+    // Swagger 2.0 has no way to express nullability, so the null member is dropped, but the
+    // allOf must still survive unwrapped so client generators can resolve the referenced schema.
```
martincostello:
```suggestion
```

src/Swashbuckle.AspNetCore.Annotations/AnnotationsSchemaFilter.cs
```diff
@@ -102,10 +102,31 @@ private static void ApplySchemaAttribute(IOpenApiSchema schema, SwaggerSchemaAtt
...
+                    TryAddNullSchema(anyOf);
+                }
+                else if (concrete.OneOf is { Count: > 0 } oneOf)
+                {
+                    TryAddNullSchema(oneOf);
+                }
+                else if (concrete.Type.HasValue)
+                {
+                    // A schema without a "type" already validates every JSON type, including null,
+                    // so the null flag is only added when a concrete type is present.
```
Youssef1313, line 127: Why does the same principle isn't applied when `AllOf`/`AnyOf`/`OneOf` is/are present?

Issue #2359, opened 2022-02-27
How to map Optional<T>


KillerBoogie:
Please excuse, if this is not the right spot to post. It is unclear if I just don't know how to use the api correctly, if swashbuckle is missing a feature or if it is a bug. I made a detailed post on Stackoverflow, but haven't received any response. Therefore I'm thinking that this might be a missing feature or a bug:

https://stackoverflow.com/questions/71265953/net-core-web-api-how-to-map-optionalt-for-swaggerdoc

KillerBoogie:
Is there anyone who can help?

momt99:
Same problem here for a custom `PaginatedList<T>` class.

momt99:
> Is there anyone who can help?

I managed to handle it using a `ISchemaFilter`. Implementing one should be straightforward. Also, [this question](https://stackoverflow.com/questions/51499445/openapi-3-0-generic-data-types) on StackOverflow may be helpful for your implementation.

maganuk:
@KillerBoogie were you able to resolve this? I'm trying to use Optional for a patch request too.

maganuk:
@momt99 any chance you could share the snippet here? thanks

momt99:
@maganuk 
This is a simplified version of what I have done for my `PaginatedList`.
```csharp
public class PaginatedListSchemaFilter : ISchemaFilter
{
    public void Apply(OpenApiSchema schema, SchemaFilterContext context)
    {
        var type = context.Type;
        if (!(type.IsGenericType
            && type.GetGenericTypeDefinition() == typeof(PaginatedList<>)))
        {
            return;
        }

        schema.Type = string.Empty;
        schema.Items = null;

        // For properties: PageIndex, PageSize, ...
        var paginatedListSchema = context.SchemaGenerator.GenerateSchema(
            typeof(PaginatedList), context.SchemaRepository);
        schema.AllOf.Add(paginatedListSchema);

        var itemType = type.GetGenericArguments()[0];
        var genericPartType = new OpenApiSchema();
        genericPartType.Properties.Add(
            ToCamelCase(nameof(PaginatedList.Items)),
            new OpenApiSchema
            {
                Type = "array",
                Items = context.SchemaGenerator.GenerateSchema(
                    itemType, context.SchemaRepository),
            });
        schema.AllOf.Add(genericPartType);
    }

    private static string ToCamelCase(string name) =>
        name.Length > 1
[...]

PR #3998, opened 2026-06-14
Move NRT handling to DataContractResolver

### LINKED ISSUE #3977: [Question]: Move nullable logic to JsonSerializerDataContractResolver
### What are you wanting to achieve?

I am trying to set the nullability of a wrapper type via a custom data contract resolver (https://github.com/desjoerd/OptionalValues/blob/main/src/OptionalValues.Swashbuckle/OptionalValueDataContractResolver.cs). The `DataProperty` has a `IsNullable` property but this is being overwritten when `SupportNonNullableReferenceTypes = true`.
This is overwritten in the `SchemaGenerator` class (https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/62b8cbcc959fb3e2570711f7ce175af2f8dff9bf/src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/SchemaGenerator.cs#L138-L148).

### What code or approach do you have so far?

I tried to handle the nullability of a property within a custom data contract resolver, but it's being overwritten.

### Additional context

Related issues: https://github.com/desjoerd/OptionalValues/issues/36, https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/2359.

I am open to moving this logic when accepted.

desjoerd:
### Problem
When `SupportNonNullableReferenceTypes` is enabled, the `SchemaGenerator` was overriding the `DataProperty.IsNullable` values provided by custom data contract resolvers. This prevented libraries which give an `OptionalValue<T>` Type from correctly marking struct-based nullable wrapper properties as nullable, as described in #2359.

### Solution
1. **Move NRT logic to the data contract resolvers** - Both `JsonSerializerDataContractResolver` and `NewtonsoftDataContractResolver` now handle NRT nullability checks internally
2. **Remove NRT override from SchemaGenerator** - The `SchemaGenerator.IsNullable()` method no longer applies non-nullable reference type metadata, allowing it to respect the resolver-provided `DataProperty.IsNullable` values

These changes make the DataContractResolver the `IsNullable` authority.

### Related Issues
- Fixes #3977 (custom resolver nullability being overwritten)
- Addresses nullability challenges described in #2359 (mapping `Optional<T>` wrapper types)

### DIFF
--- a/src/Swashbuckle.AspNetCore.Newtonsoft/SchemaGenerator/NewtonsoftDataContractResolver.cs
+++ b/src/Swashbuckle.AspNetCore.Newtonsoft/SchemaGenerator/NewtonsoftDataContractResolver.cs
@@ -219,7 +219,8 @@ public class NewtonsoftDataContractResolver : ISerializerDataContractResolver
     {
         var nullable = propertyType.IsReferenceOrNullableType();
 
-        if (_generatorOptions.SupportNonNullableReferenceTypes
+        if ((_generatorOptions.SupportNonNullableReferenceTypes
+                || _generatorOptions.NonNullableReferenceTypesAsRequired)
             && memberInfo != null)
         {
             nullable &= !memberInfo.IsNonNullableReferenceType();
--- a/src/Swashbuckle.AspNetCore.Newtonsoft/SchemaGenerator/NewtonsoftDataContractResolver.cs
+++ b/src/Swashbuckle.AspNetCore.Newtonsoft/SchemaGenerator/NewtonsoftDataContractResolver.cs
@@ -219,8 +219,7 @@ public class NewtonsoftDataContractResolver : ISerializerDataContractResolver
     {
         var nullable = propertyType.IsReferenceOrNullableType();
 
-        if ((_generatorOptions.SupportNonNullableReferenceTypes
-                || _generatorOptions.NonNullableReferenceTypesAsRequired)
+        if ((_generatorOptions.SupportNonNullableReferenceTypes || _generatorOptions.NonNullableReferenceTypesAsRequired)
             && memberInfo != null)
         {
             nullable &= !memberInfo.IsNonNullableReferenceType();
--- a/src/Swashbuckle.AspNetCore.Newtonsoft/SchemaGenerator/NewtonsoftDataContractResolver.cs
+++ b/src/Swashbuckle.AspNetCore.Newtonsoft/SchemaGenerator/NewtonsoftDataContractResolver.cs
@@ -10,11 +10,32 @@ namespace Swashbuckle.AspNetCore.Newtonsoft;
 /// <summary>
 /// A class representing an <see cref="ISerializerDataContractResolver"/> implementation that uses Newtonsoft.Json.
 /// </summary>
-/// <param name="serializerSettings">The <see cref="JsonSerializerSettings"/> to use.</param>
-public class NewtonsoftDataContractResolver(JsonSerializerSettings serializerSettings) : ISerializerDataContractResolver
+public class NewtonsoftDataContractResolver : ISerializerDataContractResolver
 {
-    private readonly JsonSerializerSettings _serializerSettings = serializerSettings;
-    private readonly IContractResolver _contractResolver = serializerSettings.ContractResolver ?? new DefaultContractResolver();
+    private readonly JsonSerializerSettings _serializerSettings;
+    private readonly SchemaGeneratorOptions _generatorOptions;
+    private readonly IContractResolver _contractResolver;
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="NewtonsoftDataContractResolver"/> class.
+    /// </summary>
+    /// <param name="serializerSettings">The <see cref="JsonSerializerSettings"/> to use.</param>
+    public NewtonsoftDataContractResolver(JsonSerializerSettings serializerSettings)
+        : this(serializerSettings, new SchemaGeneratorOptions())
+    {
+    }
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="NewtonsoftDataContractResolver"/> class.
+    /// </summary>
+    /// <param name="serializerSettings">The <see cref="JsonSerializerSettings"/> to use.</param>
+    /// <param name="generatorOptions">The <see cref="SchemaGeneratorOptions"/> to use.</param>
+    public NewtonsoftDataContractResolver(JsonSerializerSettings serializerSettings, SchemaGeneratorOptions generatorOptions)
+    {
+        _serializerSettings = serializerSettings;
+        _generatorOptions = generatorOptions ?? new SchemaGeneratorOptions();
+        _contractResolver = serializerSettings.ContractResolver ?? new DefaultContractResolver();
+    }
 
     /// <inheritdoc/>
     public DataContract GetDataContractForType(Type type)
@@ -164,14 +185,16 @@ public class NewtonsoftDataContractResolver(JsonSerializerSettings serializerSet
                         string.Equals(p.Name, jsonProperty.PropertyName, StringComparison.OrdinalIgnoreCase);
                 });
 
+            var propertyType = jsonProperty.PropertyType;
+
             dataProperties.Add(
                 new DataProperty(
                     name: jsonProperty.PropertyName,
                     isRequired: required == Required.Always || required == Required.AllowNull,
-                    isNullable: (required == Required.AllowNull || required == Required.Default) && jsonProperty.PropertyType.IsReferenceOrNullableType(),
+                    isNullable: (required == Required.AllowNull || required == Required.Default) && IsNullable(memberInfo, propertyType),
                     isReadOnly: jsonProperty.Readable && !jsonProperty.Writable && !isSetViaConstructor,
                     isWriteOnly: jsonProperty.Writable && !jsonProperty.Readable,
-                    memberType: jsonProperty.PropertyType,
+                    memberType: propertyType,
                     memberInfo: memberInfo));
         }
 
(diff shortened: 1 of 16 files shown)


martincostello (commented):
Could you add/update at least one snapshot test for this change please? It's easier to visualise the changes and protect against regressions with those.

desjoerd:
Will do, it requires a custom resolver. Shall I do it with an Optional mock?

martincostello:
> Shall I do it with an Optional mock?

Whatever you think best validates the scenario 👍

desjoerd:
Ok, because I make it completely opague and make the underlying type the member type, this patch does not work. 

I think NRT needs to be handled in the `JsonSerializerDataContractResolver` and the `NewtonsoftDataContractResolver` to make them authoritive around the nullability of properties.

Working on it, but will take more time!

desjoerd:
I've moved the logic, a few open questions:

Seems that references do not map nullable to an oneOf ref, but maybe this is another issue.

The changes give a change in the public api, I've made it backwards compatible, but having the SchemaGeneratorOptions as a argument gives more than the DataContractResolvers requires as we only use one setting. But this should not directly cause a maintainability issue I think.

The nullability for the JsonSerializer has the moved logic which is custom logic, the JsonPropertyInfo could also be used for the nullability as it handles NRT as well. Maybe a future optional improvement. 

(sorry for brevity and not adding backlinks from my phone 🙂)

martincostello:
Please fix the merge conflicts, then I'll know I'm reviewing the right set of changes.

Lots of issues come up in this area and it's a complex topic.

src/Swashbuckle.AspNetCore.Newtonsoft/SchemaGenerator/NewtonsoftDataContractResolver.cs
```diff
@@ -192,6 +215,20 @@ private List<DataProperty> GetDataPropertiesFor(JsonObjectContract jsonObjectCon
         return dataProperties;
     }
 
+    private bool IsNullable(MemberInfo memberInfo, Type propertyType)
+    {
+        var nullable = propertyType.IsReferenceOrNullableType();
+
+        if ((_generatorOptions.SupportNonNullableReferenceTypes
```
desjoerd: @martincostello 

This adds NRT support for Newtonsoft json. In theory it's a breaking change when it's not added here (as we moved it from the SchemaGenerator). But I do not know if it makes sense, because Newtonsoft Json does not have NRT enforcement during Serialization/Deserialization.

I am happy with both options. Please let me know, as I need to add a test covering the backwards compatible constructor and dependency injection registration to meet the CodeCov check in the pipeline.

desjoerd:
> Please fix the merge conflicts, then I'll know I'm reviewing the right set of changes.
> 
> Lots of issues come up in this area and it's a complex topic.

Done and I know, this area is so much fun /s. Especially OpenAPI 3.0 which has some fun recomendations/clarifications around nullability 😜.

On the codecov check, let me know what you want.

martincostello:
Don't worry about the codecov check, that's just advisory rather than a hard-block.

test/Swashbuckle.AspNetCore.SwaggerGen.Test/Fixtures/OptionalValue.cs
```diff
@@ -0,0 +1,18 @@
+namespace Swashbuckle.AspNetCore.SwaggerGen.Test;
+
+public readonly struct OptionalValue<T>
+{
+    public OptionalValue()
+    {
+        HasValue = false;
+    }
+    public OptionalValue(T value)
```
martincostello, line 10:
```suggestion

    public OptionalValue(T value)
```

src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/JsonSerializerDataContractResolver.cs
```diff
@@ -230,20 +242,35 @@ private List<DataProperty> GetDataPropertiesFor(Type objectType, out Type extens
...
 
         return dataProperties;
     }
 
+    private bool IsNullable(PropertyInfo propertyInfo, Type propertyType)
+    {
+        var nullable = propertyType.IsReferenceOrNullableType();
+
+        if (_generatorOptions.SupportNonNullableReferenceTypes
+            || _generatorOptions.NonNullableReferenceTypesAsRequired)
```
martincostello:
```suggestion
        if (_generatorOptions.SupportNonNullableReferenceTypes ||
            _generatorOptions.NonNullableReferenceTypesAsRequired)
```

src/Swashbuckle.AspNetCore.Newtonsoft/SchemaGenerator/NewtonsoftDataContractResolver.cs
```diff
@@ -192,6 +215,20 @@ private List<DataProperty> GetDataPropertiesFor(JsonObjectContract jsonObjectCon
...
+                || _generatorOptions.NonNullableReferenceTypesAsRequired)
```
martincostello:
```suggestion
        if ((_generatorOptions.SupportNonNullableReferenceTypes ||
             _generatorOptions.NonNullableReferenceTypesAsRequired)
```

test/Swashbuckle.AspNetCore.SwaggerGen.Test/Fixtures/OptionalValueDataContractResolver.cs
```diff
@@ -0,0 +1,90 @@
...
+
+                effectiveProperties.Add(effectiveProperty);
+            }
+
+            dataContract = DataContract.ForObject(
+                dataContract.UnderlyingType,
+                effectiveProperties,
+                dataContract.ObjectExtensionDataType,
+                dataContract.ObjectTypeNameProperty,
+                dataContract.ObjectTypeNameProperty,
```
martincostello: Is it right that these parameters both have the same value?

test/Swashbuckle.AspNetCore.SwaggerGen.Test/Fixtures/OptionalValueDataContractResolver.cs
```diff
@@ -0,0 +1,90 @@
+using System.Reflection;
+
+namespace Swashbuckle.AspNetCore.SwaggerGen.Test;
+
+public sealed class OptionalValueDataContractResolver(
+    ISerializerDataContractResolver innerResolver,
+    SchemaGeneratorOptions generatorOptions
+) : ISerializerDataContractResolver
```
martincostello: Style nit throughout the PR: last parenthesis shouldn't be on a new line.

e.g.:

```csharp
SomeMethod(
    first,
    second);
```

not:

```csharp
SomeMethod(
    first,
    second
);
```

test/Swashbuckle.AspNetCore.SwaggerGen.Test/VerifyTests.cs
```diff
@@ -1690,12 +1720,28 @@ private static string ToJson(OpenApiDocument document)
         return NormalizeLineBreaks(stringWriter.ToString());
     }
 
+    private static string ToJsonV31(OpenApiDocument document)
+    {
+        using var stringWriter = new StringWriter();
+        var jsonWriter = new OpenApiJsonWriter(stringWriter);
+
+        document.SerializeAsV31(jsonWriter);
```
martincostello: Use the `SerializeAs(OpenApiSpecVersion version, IOpenApiWriter writer)` overload and introduce a new optional parameter to `ToJson()` and `Verify()` to specify the version that defaults to `OpenApiSpecVersion.OpenApi3_0` rather than duplicating code.

martincostello (commented):
Sorry for the delay in reviewing properly - just some minor comments.

martincostello:
@desjoerd LGTM - have you done any manual validation that this generates valid OpenAPI 3.0/3.1 documents in a real application with NRT enabled?

desjoerd:
I did not package a temporary version locally to test it. Can still do that. I need to do it in between working 😉. Also the fun of a holiday period 😁.

desjoerd:
The fixture with the `DataContractResolver` is taken from https://github.com/desjoerd/OptionalValues/blob/main/src/OptionalValues.Swashbuckle/OptionalValueDataContractResolver.cs :).

martincostello:
There's no rush - but want to check nothing gets unintentionally broken.

desjoerd:
I did some manual validation using a locally built package. I compared the generated OpenAPI 3.0 and 3.1 output against the previous version, and the diff showed no changes for:

- The ToDo sample in this repository, both with and without `Option<T>` properties.
- Nullable and non-nullable properties with NRT enabled using:
  - schema references;
  - `int` and enums;
  - strings.
- My private Minesweeper application.

In my OptionalValues package, this PR fixes the original issue: NRT nullability now works correctly without requiring any changes to the package itself.

I did find one compatibility consideration. My unit tests construct `JsonSerializerDataContractResolver` directly. With this PR, the resolver must receive the same `SchemaGeneratorOptions` instance as the `SchemaGenerator`; otherwise, it does not know that NRT support is enabled:

```diff
var schemaGeneratorOptions = new SchemaGeneratorOptions
{
    SupportNonNullableReferenceTypes = true,
};

var schemaGenerator = new SchemaGenerator(
    schemaGeneratorOptions,
-   new JsonSerializerDataContractResolver(JsonSerializerOptions.Default)
+   new JsonSerializerDataContractResolver(
+       JsonSerializerOptions.Default,
+       schemaGeneratorOptions)
);
```

This affected [[this test setup in OptionalValues](https://github.com/desjoerd/OptionalValues/blob/41e4686c8acface0350e04eb15558935576ae775/test/OptionalValues.Swashbuckle.V10.Tests/SchemaGeneratorTest.cs#L18)].

The normal DI-based setup passes the options automatically, so decorators such as the one used by OptionalValues continue to work without modification.

However, this is a behavioral compatibility change for consumers who:

- construct `JsonSerializerDataContractResolver` or `NewtonsoftDataContractResolver` directly; or
- derive from either resolver and call the original single-argument constructor.

Those consumers must use the new constructor and pass the generator options; otherwise, `SupportNonNullableReferenceTypes` and `NonNullableReferenceTypesAsRequired` will no longer affect the resolver’s nullability results.

I expect this to affect relatively few consumers, but it seems worth documenting.

desjoerd:
Another note, while reviewing I noticed this behavior as well -> https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/3649. Which got me confused at first, but the diff did not show changes and I saw it's an existing issue. 

It's an issue to really love /s 😅, so much pain around this AllOf, OneOf especially with the difference between OpenApi 3.0 and 3.1 and OpenApi.NET.

martincostello (approved):

PR #4024, opened 2026-06-27
Fix #3978: add UseAnnotatedEnumValues option to generate oneOf/const/description enum schemas

### LINKED ISSUE #3978: [Feature request]: Add comments on enum members to OpenAPI document

I have enums with comments on their enum members. It would be nice to see this comments in the OpenAPI document.

Example enum:
```cs
public enum MyStatus
{
    /// <summary>
    /// Comment on value
    /// </summary>
    Value,

    /// <summary>
    /// Other comment
    /// </summary>
    OtherValue
}
```

According to the [spec](https://spec.openapis.org/oas/v3.1.1.html#annotated-enumerations) this is not possible with how enums are currently generated (with an `enum` array). But the spec describes an alternative way that allows enum member descriptions (`oneOf` or `anyOf` with `const` values). Here is also an example: https://github.com/OAI/OpenAPI-Specification/issues/348#issuecomment-336194030
As I understand it this would only be allowed in OpenAPI v3.1 mode. So the older modes would have to stay as they are.

The chosen solution should also be supported by popular OpenAPI UIs (Swagger UI, Scalar, ...)

### Additional context

Maybe the `Description` attribute should be supported in addition to XML doc comments.

I know that `Unchase.Swashbuckle.AspNetCore.Extensions` exists (or rather `CoreDX.Unchase.Swashbuckle.AspNetCore.Extensions` for a Swashbuckle 10 compatible package). That is better than nothing, but not optimal. It generates properties which are not in the OpenAPI schema (with an `x-` prefix).

Related request for Microsoft.AspNetCore.OpenApi: https://github.com/dotnet/aspnetcore/issues/66062

danjourno-dev:
# Pull Request

Fixes #3978

Adds a new UseAnnotatedEnumValues option to SchemaGeneratorOptions that, when enabled, generates richer enum schemas using oneOf with per-value const and description instead of a flat enum array.

How it works:

When `UseAnnotatedEnumValues = true` and at least one enum member carries a `[Description]` or `[Display(Description = "...")]` attribute, CreatePrimitiveSchema switches 
from:

`{ "enum": ["Zero", "One", "Two"] }`

to:

```
{
  "oneOf": [
    { "const": "Zero", "description": "Value zero description" },
    { "const": "One",  "description": "Value one description" },
    { "const": "Two" }
  ]
}
```

This allows OpenAPI consumers (e.g. Swagger UI, code generators) to surface human-readable descriptions per enum value, which is not possible with the plain enum keyword.

Changes:
- SchemaGeneratorOptions - new bool UseAnnotatedEnumValues property (defaults to false, fully opt-in, no breaking change).
- SchemaGenerator.CreatePrimitiveSchema - when the option is set and the enum type has any annotated members, builds OneOf sub-schemas with Const (unwrapped from JSON encoding via JsonModelFactory.CreateFromJson) and Description sourced from [Description] or [Display(Description)].
- Duplicate JSON values (e.g. aliased enum members) are collapsed via GroupBy before building the oneOf list.
- Tests added in JsonSerializerSchemaGeneratorTests covering [Description] and [Display] attributes, unannotated fallback, and JsonStringEnumConverter interaction.

### DIFF
--- a/docs/configure-and-customize-swaggergen.md
+++ b/docs/configure-and-customize-swaggergen.md
@@ -854,6 +854,60 @@ services.AddSwaggerGen(options =>
 > Filter pipelines are DI-aware. That is, you can create filters with constructor parameters and if the parameter types
 > are registered with the DI framework, they'll be automatically injected when the filters are instantiated.
 
+### Describe Enum Members with Annotated Enum Values
+
+By default, Swashbuckle generates enum schemas as a flat array of values:
+
+```yaml
+CvvCheckResult:
+  type: string
+  enum:
+    - D
+    - M
+    - N
+```
+
+If your enum members are annotated with `[Description]` or `[Display(Description = "...")]`, you can opt in to generating per-value descriptions using the [OpenAPI annotated enumeration pattern](https://spec.openapis.org/oas/v3.1.1.html#fixed-fields-20) (`oneOf` with `const` + `description` sub-schemas):
+
+```cs
+services.AddSwaggerGen(options =>
+{
+    options.UseAnnotatedEnumValues();
+});
+```
+
+With this option enabled, enum members decorated with `[Description]` or `[Display(Description = "...")]`:
+
+```cs
+public enum CvvCheckResult
+{
+    [Description("Suspicious transaction")]
+    D,
+
+    [Description("Match")]
+    M,
+
+    [Description("No Match")]
+    N,
+}
+```
+
+…are rendered as:
+
+```yaml
+CvvCheckResult:
+  type: string
+  oneOf:
+    - const: '"D"'
+      description: Suspicious transaction
+    - const: '"M"'
+      description: Match
+    - const: '"N"'
+      description: No Match
+```
+
+Members without a `[Description]` still appear in the `oneOf` list; their `description` field is simply omitted.
+
 ### Schema Filters
 
 Swashbuckle.AspnetCore generates an OpenAPI-flavored [JSONSchema](https://swagger.io/specification/#schema-object) for every parameter, response
--- a/docs/configure-and-customize-swaggergen.md
+++ b/docs/configure-and-customize-swaggergen.md
@@ -867,7 +867,7 @@ CvvCheckResult:
     - N
 ```
 
-If your enum members are annotated with `[Description]` or `[Display(Description = "...")]`, you can opt in to generating per-value descriptions using the [OpenAPI annotated enumeration pattern](https://spec.openapis.org/oas/v3.1.1.html#fixed-fields-20) (`oneOf` with `const` + `description` sub-schemas):
+If your enum members are annotated with `[Description]`, `[Display(Description = "...")]`, or XML doc comments (`/// <summary>`), you can opt in to generating per-value descriptions using the [OpenAPI annotated enumeration pattern](https://spec.openapis.org/oas/v3.1.1.html#fixed-fields-20) (`oneOf` with `const` + `description` sub-schemas):
 
 ```cs
 services.AddSwaggerGen(options =>
@@ -907,6 +907,34 @@ CvvCheckResult:
 ```
 
 Members without a `[Description]` still appear in the `oneOf` list; their `description` field is simply omitted.
+
+#### XML Doc Comments as Descriptions
+
+If you have called `IncludeXmlComments`, XML `<summary>` comments on enum members are also used as descriptions:
+
+```cs
+services.AddSwaggerGen(options =>
+{
+    options.UseAnnotatedEnumValues();
+    options.IncludeXmlComments("MyApi.xml");
+});
+```
+
+```cs
+public enum CvvCheckResult
+{
+    /// <summary>Suspicious transaction</summary>
+    D,
+
+    /// <summary>Match</summary>
+    M,
+
+    /// <summary>No Match</summary>
+    N,
+}
+```
+
+When a member has both an attribute and an XML doc comment, the attribute takes priority and the XML comment is ignored for that member.
 
 ### Schema Filters
 
--- a/docs/configure-and-customize-swaggergen.md
+++ b/docs/configure-and-customize-swaggergen.md
@@ -892,7 +892,7 @@ public enum CvvCheckResult
 }
 ```
 
-…are rendered as:
+...are rendered as:
 
 ```yaml
 CvvCheckResult:
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenOptionsExtensions.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenOptionsExtensions.cs
@@ -823,6 +823,18 @@ public static class SwaggerGenOptionsExtensions
         swaggerGenOptions.AddOperationFilterInstance(new XmlCommentsOperationFilter(xmlDocMembers, swaggerGenOptions.SwaggerGeneratorOptions));
         swaggerGenOptions.AddSchemaFilterInstance(new XmlCommentsSchemaFilter(xmlDocMembers, swaggerGenOptions.SwaggerGeneratorOptions));
 
+        swaggerGenOptions.SchemaGeneratorOptions.EnumMemberDescriptionProviders.Add(field =>
+        {
+            var memberName = XmlCommentsNodeNameHelper.GetMemberNameForFieldOrProperty(field);
+            if (xmlDocMembers.TryGetValue(memberName, out var memberNode))
+            {
+                var summaryNode = memberNode.SelectFirstChild("summary");
+                if (summaryNode != null)
+                    return XmlCommentsTextHelper.Humanize(summaryNode.InnerXml, swaggerGenOptions.SwaggerGeneratorOptions.XmlCommentEndOfLine);
+            }
+            return null;
+        });
+
         if (includeControllerXmlComments)
         {
             swaggerGenOptions.AddDocumentFilterInstance(new XmlCommentsDocumentFilter(xmlDocMembers, swaggerGenOptions.SwaggerGeneratorOptions));
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenOptionsExtensions.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenOptionsExtensions.cs
@@ -189,14 +189,25 @@ public static class SwaggerGenOptionsExtensions
     }
 
     /// <summary>
-    /// Generate enum schemas as annotated enumerations when enum members have
+    /// Generates enum schemas as <c>oneOf/const</c> enumerations (OpenAPI 3.1+) or as
+    /// flat <c>enum</c> arrays (&lt;3.1) when enum members have
     /// <see cref="System.ComponentModel.DescriptionAttribute"/> or
     /// <see cref="System.ComponentModel.DataAnnotations.DisplayAttribute"/> descriptions.
     /// </summary>
     /// <param name="swaggerGenOptions"></param>
-    public static void UseAnnotatedEnumValues(this SwaggerGenOptions swaggerGenOptions)
+    /// <param name="openApiVersion">
+    /// Target OpenAPI version. Use <see cref="OpenApiSpecVersion.OpenApi3_1"/> (default) to emit
+    /// <c>oneOf/const/description</c> per enum member. Use <see cref="OpenApiSpecVersion.OpenApi3_0"/>
+    /// or <see cref="OpenApiSpecVersion.OpenApi2_0"/> to fall back to a flat <c>enum</c> array
+    /// (per-value descriptions are dropped as they cannot be expressed in those versions).
+    /// This must match the <c>OpenApiVersion</c> configured on <c>UseSwagger</c>.
+    /// </param>
+    public static void UseAnnotatedEnumValues(
+        this SwaggerGenOptions swaggerGenOptions,
+        OpenApiSpecVersion openApiVersion = OpenApiSpecVersion.OpenApi3_1)
     {
         swaggerGenOptions.SchemaGeneratorOptions.UseAnnotatedEnumValues = true;
+        swaggerGenOptions.SchemaGeneratorOptions.AnnotatedEnumOpenApiVersion = openApiVersion;
     }
 
     /// <summary>
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenOptionsExtensions.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenOptionsExtensions.cs
@@ -189,8 +189,8 @@ public static class SwaggerGenOptionsExtensions
     }
 
     /// <summary>
-    /// Generate enum schemas as annotated enumerations (oneOf with const + description sub-schemas)
-    /// when enum members have <see cref="System.ComponentModel.DescriptionAttribute"/> or
+    /// Generate enum schemas as annotated enumerations when enum members have
+    /// <see cref="System.ComponentModel.DescriptionAttribute"/> or
     /// <see cref="System.ComponentModel.DataAnnotations.DisplayAttribute"/> descriptions.
     /// </summary>
     /// <param name="swaggerGenOptions"></param>
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenOptionsExtensions.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenOptionsExtensions.cs
@@ -189,6 +189,17 @@ public static class SwaggerGenOptionsExtensions
     }
 
     /// <summary>
+    /// Generate enum schemas as annotated enumerations (oneOf with const + description sub-schemas)
+    /// when enum members have <see cref="System.ComponentModel.DescriptionAttribute"/> or
+    /// <see cref="System.ComponentModel.DataAnnotations.DisplayAttribute"/> descriptions.
+    /// </summary>
+    /// <param name="swaggerGenOptions"></param>
+    public static void UseAnnotatedEnumValues(this SwaggerGenOptions swaggerGenOptions)
+    {
+        swaggerGenOptions.SchemaGeneratorOptions.UseAnnotatedEnumValues = true;
+    }
+
+    /// <summary>
     /// Provide a custom strategy for generating the unique Ids that are used to reference object Schemas
     /// </summary>
     /// <param name="swaggerGenOptions"></param>
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/SchemaGenerator.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/SchemaGenerator.cs
@@ -6,6 +6,7 @@ using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.Mvc.ApiExplorer;
 using Microsoft.OpenApi;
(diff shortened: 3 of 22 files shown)


docs/configure-and-customize-swaggergen.md
```diff
@@ -854,6 +854,60 @@ services.AddSwaggerGen(options =>
...
+
+    [Description("Match")]
+    M,
+
+    [Description("No Match")]
+    N,
+}
+```
+
+…are rendered as:
```
martincostello:
```suggestion
...are rendered as:
```

src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenOptionsExtensions.cs
```diff
@@ -188,6 +188,17 @@ public static void UseInlineDefinitionsForEnums(this SwaggerGenOptions swaggerGe
         swaggerGenOptions.SchemaGeneratorOptions.UseInlineDefinitionsForEnums = true;
     }
 
+    /// <summary>
+    /// Generate enum schemas as annotated enumerations (oneOf with const + description sub-schemas)
+    /// when enum members have <see cref="System.ComponentModel.DescriptionAttribute"/> or
+    /// <see cref="System.ComponentModel.DataAnnotations.DisplayAttribute"/> descriptions.
```
martincostello:
```suggestion
    /// Generate enum schemas as annotated enumerations when enum members have
    /// <see cref="System.ComponentModel.DescriptionAttribute"/> or
    /// <see cref="System.ComponentModel.DataAnnotations.DisplayAttribute"/> descriptions.
```

src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/SchemaGenerator.cs
```diff
@@ -360,15 +361,44 @@ private static OpenApiSchema CreatePrimitiveSchema(DataContract dataContract)
                 enumValues = enumValues.Append(null);
             }*/
 
-            schema.Enum = [.. enumValues
-                .Select(value => dataContract.JsonConverter(value))
-                .Distinct()
-                .Select(JsonModelFactory.CreateFromJson)];
+            if (options?.UseAnnotatedEnumValues == true && HasEnumMemberDescriptions(underlyingType))
+            {
+                schema.OneOf = [.. enumValues
```
martincostello, line 368: What happens to the enum if the user is generating Swagger 2.0 or OpenAPI 3.0 documents and not 3.1?

src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/SchemaGenerator.cs
```diff
@@ -360,15 +361,44 @@ private static OpenApiSchema CreatePrimitiveSchema(DataContract dataContract)
...
+                    .Select(value => (Raw: value, Json: dataContract.JsonConverter(value)))
+                    .GroupBy(x => x.Json)
+                    .Select(g => g.First())
+                    .Select(x => (IOpenApiSchema)new OpenApiSchema
+                    {
+                        Const = JsonModelFactory.CreateFromJson(x.Json)?.ToString(),
```
martincostello, line 374: Will this work correctly if the user's enum is returning the values rather than the names?

src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/SchemaGenerator.cs
```diff
@@ -360,15 +361,44 @@ private static OpenApiSchema CreatePrimitiveSchema(DataContract dataContract)
...
+    private static bool HasEnumMemberDescriptions(Type enumType) =>
+        enumType.GetFields(BindingFlags.Public | BindingFlags.Static)
+            .Any(f => f.GetCustomAttribute<DescriptionAttribute>() != null
+                   || f.GetCustomAttribute<DisplayAttribute>()?.GetDescription() != null);
+
+    private static string GetEnumMemberDescription(Type enumType, object value)
+    {
+        var name = Enum.GetName(enumType, value);
+        if (name is null) return null;
+        var field = enumType.GetField(name);
```
martincostello, line 403:
```suggestion
        var name = Enum.GetName(enumType, value);

        if (name is null)
        {
            return null;
        }

        var field = enumType.GetField(name);
```

test/Swashbuckle.AspNetCore.SwaggerGen.Test/SchemaGenerator/JsonSerializerSchemaGeneratorTests.cs
martincostello, line 1: Could you also add some tests using Verify snapshots following the existing patterns?

src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/SchemaGenerator.cs
danjourno-dev: Const is always a string according to this 
See https://learn.microsoft.com/en-us/dotnet/api/microsoft.openapi.openapischema.const
I've updated unit tests to demonstrate functionality for both values and names.

src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/SchemaGenerator.cs
danjourno-dev: Fair point. I've added a gate on the version. 
https://github.com/domaindrivendev/Swashbuckle.AspNetCore/pull/4024/commits/c518d00847aca85ba631b2cd734637f0466f6f8f
However it does lead to config duplication since Dependency direction blocks it. SwaggerGen project doesn't reference Swagger project - Swagger depends on SwaggerGen, not the reverse. SwaggerOptions lives in Swagger; SchemaGenerator lives in SwaggerGen. Can't import across that boundary.
@martincostello If that approach is not acceptable, I have an idea for another way.

test/Swashbuckle.AspNetCore.SwaggerGen.Test/SchemaGenerator/JsonSerializerSchemaGeneratorTests.cs
danjourno-dev: Added https://github.com/domaindrivendev/Swashbuckle.AspNetCore/pull/4024/commits/0153f998b4fe292ed6b1fbc0df0df5e7f0512dfc

danjourno-dev:
@martincostello It's ready for you now

docs/configure-and-customize-swaggergen.md
```diff
@@ -854,6 +854,88 @@ services.AddSwaggerGen(options =>
...
+```yaml
+CvvCheckResult:
+  type: string
+  enum:
+    - D
+    - M
+    - N
+```
+
+If your enum members are annotated with `[Description]`, `[Display(Description = "...")]`, or XML doc comments (`/// <summary>`), you can opt in to generating per-value descriptions using the [OpenAPI annotated enumeration pattern](https://spec.openapis.org/oas/v3.1.1.html#fixed-fields-20) (`oneOf` with `const` + `description` sub-schemas):
```
martincostello, line 870:
```suggestion
If your enum members are annotated with `[Description]`, `[Display(Description = "...")]`, or XML doc comments (`/// <summary>`),
you can opt in to generating per-value descriptions using the [OpenAPI annotated enumeration pattern](https://spec.openapis.org/oas/v3.1.1.html#fixed-fields-20) (`oneOf` with `const` and `description` sub-schemas):
```

docs/configure-and-customize-swaggergen.md
```diff
@@ -854,6 +854,88 @@ services.AddSwaggerGen(options =>
...
+
+```yaml
+CvvCheckResult:
+  type: string
+  oneOf:
+    - const: '"D"'
+      description: Suspicious transaction
+    - const: '"M"'
+      description: Match
+    - const: '"N"'
```
martincostello, line 905: Is it correct that the `const` values include an inner `"` pair?

docs/configure-and-customize-swaggergen.md
```diff
@@ -854,6 +854,88 @@ services.AddSwaggerGen(options =>
...
+      description: No Match
+```
+
+Members without a `[Description]` still appear in the `oneOf` list; their `description` field is simply omitted.
+
+#### XML Doc Comments as Descriptions
```
martincostello, line 911:
```suggestion
#### XML Documentation Comments as Descriptions
```

src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenOptionsExtensions.cs
```diff
@@ -188,6 +188,28 @@ public static void UseInlineDefinitionsForEnums(this SwaggerGenOptions swaggerGe
         swaggerGenOptions.SchemaGeneratorOptions.UseInlineDefinitionsForEnums = true;
     }
 
+    /// <summary>
+    /// Generates enum schemas as <c>oneOf/const</c> enumerations (OpenAPI 3.1+) or as
+    /// flat <c>enum</c> arrays (&lt;3.1) when enum members have
```
martincostello, line 193:
```suggestion
    /// flat <c>enum</c> arrays (OpenAPI version &lt;3.1) when enum members have
```

src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenOptionsExtensions.cs
```diff
@@ -188,6 +188,28 @@ public static void UseInlineDefinitionsForEnums(this SwaggerGenOptions swaggerGe
...
+    /// <see cref="System.ComponentModel.DescriptionAttribute"/> or
+    /// <see cref="System.ComponentModel.DataAnnotations.DisplayAttribute"/> descriptions.
+    /// </summary>
+    /// <param name="swaggerGenOptions"></param>
```
martincostello, line 197: No inner text.

src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenOptionsExtensions.cs
```diff
@@ -188,6 +188,28 @@ public static void UseInlineDefinitionsForEnums(this SwaggerGenOptions swaggerGe
...
+    /// <param name="openApiVersion">
+    /// Target OpenAPI version. Use <see cref="OpenApiSpecVersion.OpenApi3_1"/> (default) to emit
```
martincostello, line 199: `3_0` should be the default to match the overall default.

src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenOptionsExtensions.cs
```diff
@@ -188,6 +188,28 @@ public static void UseInlineDefinitionsForEnums(this SwaggerGenOptions swaggerGe
...
+    /// <c>oneOf/const/description</c> per enum member. Use <see cref="OpenApiSpecVersion.OpenApi3_0"/>
+    /// or <see cref="OpenApiSpecVersion.OpenApi2_0"/> to fall back to a flat <c>enum</c> array
+    /// (per-value descriptions are dropped as they cannot be expressed in those versions).
+    /// This must match the <c>OpenApiVersion</c> configured on <c>UseSwagger</c>.
+    /// </param>
+    public static void UseAnnotatedEnumValues(
+        this SwaggerGenOptions swaggerGenOptions,
+        OpenApiSpecVersion openApiVersion = OpenApiSpecVersion.OpenApi3_1)
```
martincostello, line 207: I'm not convinced this is a good idea - having an option to mix and match versions within a document sounds wrong and seems to just invite users to generate non-compliant OpenAPI documents.

src/Swashbuckle.AspNetCore.SwaggerGen/DependencyInjection/SwaggerGenOptionsExtensions.cs
```diff
@@ -801,6 +823,18 @@ public static void IncludeXmlComments(
...
+        swaggerGenOptions.SchemaGeneratorOptions.EnumMemberDescriptionProviders.Add(field =>
+        {
+            var memberName = XmlCommentsNodeNameHelper.GetMemberNameForFieldOrProperty(field);
+            if (xmlDocMembers.TryGetValue(memberName, out var memberNode))
+            {
+                var summaryNode = memberNode.SelectFirstChild("summary");
+                if (summaryNode != null)
+                    return XmlCommentsTextHelper.Humanize(summaryNode.InnerXml, swaggerGenOptions.SwaggerGeneratorOptions.XmlCommentEndOfLine);
+            }
+            return null;
```
martincostello, line 835:
```suggestion
            var memberName = XmlCommentsNodeNameHelper.GetMemberNameForFieldOrProperty(field);

            if (xmlDocMembers.TryGetValue(memberName, out var memberNode))
            {
                var summaryNode = memberNode.SelectFirstChild("summary");
                if (summaryNode != null)
                {
                    return XmlCommentsTextHelper.Humanize(summaryNode.InnerXml, swaggerGenOptions.SwaggerGeneratorOptions.XmlCommentEndOfLine);
                }
            }

            return null;
```

src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/SchemaGenerator.cs
```diff
@@ -360,15 +361,76 @@ private static OpenApiSchema CreatePrimitiveSchema(DataContract dataContract)
                 enumValues = enumValues.Append(null);
             }*/
 
-            schema.Enum = [.. enumValues
-                .Select(value => dataContract.JsonConverter(value))
-                .Distinct()
-                .Select(JsonModelFactory.CreateFromJson)];
+            if (options?.UseAnnotatedEnumValues == true
+                && options.AnnotatedEnumOpenApiVersion >= OpenApiSpecVersion.OpenApi3_1
```
martincostello, line 365: This could potentially introduce a bug when/if OpenAPI 3.2 support is added - it won't necessarily be correct behaviour because OpenAPI doesn't follow SemVer.

src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/SchemaGenerator.cs
martincostello: This doesn't look correct as it will generate strings of values that happen to be integers, not actual integers?

```json
"IntEnumWithDescriptions": {
  "type": "integer",
  "oneOf": [ { "const": "2", ... }, { "const": "4", ... }, { "const": "8" } ],
  "format": "int32"
}
```

"2" is a string not an integer, so this wouldn't be valid?

src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/SchemaGenerator.cs
```diff
@@ -360,15 +361,76 @@ private static OpenApiSchema CreatePrimitiveSchema(DataContract dataContract)
...
+        {
+            return description;
+        }
+
+        if (options?.EnumMemberDescriptionProviders is { Count: > 0 } providers)
+        {
+            foreach (var provider in providers)
+            {
+                var result = provider(field);
+                if (result != null) return result;
```
martincostello, line 427:
```suggestion
                if (result != null)
                {
                    return result;
                }
```

src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/SchemaGeneratorOptions.cs
```diff
@@ -18,6 +19,12 @@ public SchemaGeneratorOptions()
 
     public bool UseInlineDefinitionsForEnums { get; set; }
 
+    public bool UseAnnotatedEnumValues { get; set; }
+
+    public OpenApiSpecVersion AnnotatedEnumOpenApiVersion { get; set; } = OpenApiSpecVersion.OpenApi3_1;
```
martincostello, line 24: Again, doesn't match Swashbuckle's actual default.

test/Swashbuckle.AspNetCore.SwaggerGen.Test/snapshots/10_0/VerifyTests.GenerateSchema_AnnotatedEnum_WhenDescriptionAttribute.verified.txt
```diff
@@ -0,0 +1,56 @@
+﻿{
+  "openapi": "3.1.1",
```
martincostello, line 2: All the snapshots are 3.1 - there's no coverage for other versions.

test/Swashbuckle.AspNetCore.TestSupport/Fixtures/Enums.cs
```diff
@@ -38,3 +41,48 @@ public enum IntEnumWithDuplicateValues : int
     PreferredName = 1,
     OldNameForBackwardsCompatibility = PreferredName,
 }
+
+public enum IntEnumWithDescriptions : int
```
martincostello, line 45: Can omit the `: int` for all of these because that's the default for enums.

test/Swashbuckle.AspNetCore.SwaggerGen.Test/VerifyTests.cs
```diff
@@ -1696,6 +1840,22 @@ await Verifier.Verify(ToJson(document))
...
+        document.SerializeAsV31(jsonWriter);
+
+        return NormalizeLineBreaks(stringWriter.ToString());
+    }
+
+    private static async Task VerifyV31(OpenApiDocument document)
+    {
+        await Verifier.Verify(ToJsonV31(document))
+                      .UseDirectory($"snapshots/{Environment.Version.Major}_{Environment.Version.Minor}");
+    }
```
martincostello, line 1857: Use the `SerializeAs(OpenApiSpecVersion version, IOpenApiWriter writer)` overload and introduce a new optional parameter to `ToJson()` and `Verify()` to specify the version that defaults to `OpenApiSpecVersion.OpenApi3_0` rather than duplicating code.

test/Swashbuckle.AspNetCore.SwaggerGen.Test/VerifyTests.cs
```diff
@@ -1654,6 +1654,150 @@ public async Task GetSwagger_GenerateConsumesSchemas_ForProvidedOpenApiOperation
...
+                c.UseAnnotatedEnumValues = true;
+                c.EnumMemberDescriptionProviders.Add(field =>
+                    field.DeclaringType == typeof(IntEnumWithXmlComments)
+                        ? field.Name switch
+                        {
+                            "Two" => "Value two xml comment",
+                            "Four" => "Value four xml comment",
+                            _ => null,
+                        }
+                        : null);
```
martincostello, line 1750: There's no tests that use the built-in XML documentation resolution.

src/Swashbuckle.AspNetCore.SwaggerGen/SchemaGenerator/SchemaGenerator.cs
```diff
@@ -360,15 +361,76 @@ private static OpenApiSchema CreatePrimitiveSchema(DataContract dataContract)
...
-                .Distinct()
-                .Select(JsonModelFactory.CreateFromJson)];
+            if (options?.UseAnnotatedEnumValues == true
+                && options.AnnotatedEnumOpenApiVersion >= OpenApiSpecVersion.OpenApi3_1
+                && HasEnumMemberDescriptions(underlyingType, options))
+            {
+                schema.OneOf = [.. enumValues
+                    .Select(value => (Raw: value, Json: dataContract.JsonConverter(value)))
+                    .GroupBy(x => x.Json)
+                    .Select(g => g.First())
```
martincostello, line 371: Users might not get the value they want described if there are duplicates.

PR #4063, opened 2026-07-16
Do not mark nested form properties as required when their container property is optional

### LINKED ISSUE #3534: [Bug]: Required nested properties are marked required even if outer property is optional
### Describe the bug

Consider the following model:

```
public record Model ( Inner? inner, [Required] string other );
public record Inner ( [Required] string prop );
```

The model is used in an ASP.NET Core action like this:

```
public IActionResult MyAction ( [FromForm, Required] Model model )
```

It translates the nested models to a flattened representation of dot-concatenated properties, i.e., `inner.prop` - unfortunately, this property is set to required which is not true.

### Expected behavior

The property should be optional.

### Actual behavior

The property is marked as required.

### Steps to reproduce

See description.

### Exception(s) (if any)

_No response_

8

### .NET Version

8

### Anything else?

_No response_

KitKeen:
Fixes #3534

## Summary

When a `[FromForm]` model is flattened into dot-concatenated properties (e.g. `Inner.Prop`), a nested property marked `[Required]` ends up in the top-level `required` list of the form schema even when its container property is optional:

```csharp
public record Model(Inner? Inner, [Required] string Other);
public record Inner([Required] string Prop);

public IActionResult MyAction([FromForm, Required] Model model) => ...
```

Here `Inner.Prop` is generated as required, but the whole `Inner` subobject may legitimately be absent, so the field must be optional.

## Root cause

`ApiParameterDescriptionExtensions.IsRequiredParameter` only inspects the leaf member (via its `PropertyInfo` custom attributes or the `IsRequired` flag). The container chain of a flattened member is never considered.

## Fix

`IsRequiredParameter` now additionally verifies the container chain for flattened members: for a dotted name, every intermediate property (resolved by walking from `ParameterDescriptor.ParameterType`) must itself carry one of the required markers (`RequiredAttribute`, `BindRequiredAttribute`, `RequiredMemberAttribute` - the same set the leaf check uses). If any container is optional, the member is not required.

Deliberately conservative bail-outs (behavior stays exactly as before):

- name has no dots, or the member has no container metadata (regular top-level parameters)
- the chain cannot be resolved: custom binding names, collection indexers, ambiguous or missing properties
- the root parameter type is unavailable

No public API changes.

## Tests

TDD: the new test reproduces the issue first (`Inner.Prop` appeared in `required`), then the fix turns it green. A test infrastructure note: `ApiDescriptionFactory` reassigns `ParameterDescriptor` by name matching, which cannot express flattened members (real ApiExplorer assigns the root parameter's descriptor to every flattened member), so the test assigns descriptors after the factory call.

Full Swashbuckle.AspNetCore.SwaggerGen.Test suite: 740/740 passing on net8.0, net9.0, and net10.0.

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/ApiParameterDescriptionExtensions.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/ApiParameterDescriptionExtensions.cs
@@ -36,14 +36,60 @@ public static class ApiParameterDescriptionExtensions
         // This is the default logic for IsRequired
         bool IsRequired() => apiParameter.CustomAttributes().Any(attr => RequiredAttributeTypes.Contains(attr.GetType()));
 
-        // This is to keep compatibility with MVC controller logic that has existed in the past
-        if (apiParameter.ParameterDescriptor is ControllerParameterDescriptor)
+        // This is to keep compatibility with MVC controller logic that has existed in the past.
+        // For non-controllers, prefer the IsRequired flag if we're not on netstandard 2.0, otherwise fallback to the default logic.
+        var isRequired = apiParameter.ParameterDescriptor is ControllerParameterDescriptor
+            ? IsRequired()
+            : apiParameter.IsRequired;
+
+        // A member flattened from a nested model (e.g. "Inner.Prop") can only be required
+        // if every property in its container chain is required as well
+        return isRequired && HasRequiredContainerChain(apiParameter);
+    }
+
+    private static bool HasRequiredContainerChain(ApiParameterDescription apiParameter)
+    {
+        var name = apiParameter.Name;
+        if (name is null || !name.Contains('.') || apiParameter.ModelMetadata?.ContainerType is null)
         {
-            return IsRequired();
+            return true;
         }
 
-        // For non-controllers, prefer the IsRequired flag if we're not on netstandard 2.0, otherwise fallback to the default logic.
-        return apiParameter.IsRequired;
+        var containerType =
+            apiParameter.ParameterDescriptor?.ParameterType ??
+            apiParameter.ParameterInfo()?.ParameterType;
+
+        var segments = name.Split('.');
+
+        // Walk the container chain, i.e. every segment except the last one, which is the
+        // member itself. If the chain cannot be resolved (custom binding names, collection
+        // indexers, etc.), assume the container is required to keep the previous behavior.
+        for (var i = 0; i < segments.Length - 1 && containerType is not null; i++)
+        {
+            PropertyInfo containerProperty;
+            try
+            {
+                containerProperty = containerType.GetProperty(segments[i], BindingFlags.Public | BindingFlags.Instance | BindingFlags.IgnoreCase);
+            }
+            catch (AmbiguousMatchException)
+            {
+                return true;
+            }
+
+            if (containerProperty is null)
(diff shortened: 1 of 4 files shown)


src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/ApiParameterDescriptionExtensions.cs
```diff
@@ -36,14 +36,60 @@ public static bool IsRequiredParameter(this ApiParameterDescription apiParameter
...
+        // indexers, etc.), assume the container is required to keep the previous behavior.
+        for (var i = 0; i < segments.Length - 1 && containerType is not null; i++)
         {
-            return IsRequired();
+            PropertyInfo containerProperty;
+            try
+            {
+                containerProperty = containerType.GetProperty(segments[i], BindingFlags.Public | BindingFlags.Instance | BindingFlags.IgnoreCase);
+            }
+            catch (AmbiguousMatchException)
```
martincostello, line 74: This feels like a workaround to an issue with the reflection code.

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/ApiParameterDescriptionExtensions.cs
```diff
@@ -36,14 +36,60 @@ public static bool IsRequiredParameter(this ApiParameterDescription apiParameter
...
-        if (apiParameter.ParameterDescriptor is ControllerParameterDescriptor)
+        // This is to keep compatibility with MVC controller logic that has existed in the past.
+        // For non-controllers, prefer the IsRequired flag if we're not on netstandard 2.0, otherwise fallback to the default logic.
+        var isRequired = apiParameter.ParameterDescriptor is ControllerParameterDescriptor
+            ? IsRequired()
+            : apiParameter.IsRequired;
+
+        // A member flattened from a nested model (e.g. "Inner.Prop") can only be required
+        // if every property in its container chain is required as well
+        return isRequired && HasRequiredContainerChain(apiParameter);
```
martincostello, line 47: I'm not convinced this is correct that having an optional member in a required object makes the whole thing non-required.

src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/ApiParameterDescriptionExtensions.cs
```diff
@@ -36,14 +36,60 @@ public static bool IsRequiredParameter(this ApiParameterDescription apiParameter
...
+    }
+
+    private static bool HasRequiredContainerChain(ApiParameterDescription apiParameter)
```
martincostello, line 50: This doesn't have any support for nullable reference types - it only works for explicit attributes.

test/Swashbuckle.AspNetCore.SwaggerGen.Test/SwaggerGenerator/SwaggerGeneratorTests.cs
martincostello, line 1: Missing any snapshot tests that show the shape of the generated OpenAPI documents.

cremor:
You might want to check out the PR that I linked in https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/3534#issuecomment-5010705775

PR #3786, opened 2026-02-13
Support ASP.NET Core 11

martincostello:
- Add support for .NET 11.
- Drop support for .NET 8 and 9.

### TODO

Before merge:

- [ ] Use stable version of .NET 11 SDK
- [ ] Use stable versions of Microsoft.AspNetCore NuGet packages
- [x] Use Microsoft.OpenApi 3.x.x for `net11.0` ([![Latest][openapi-version]][openapi-download])
- [ ] Use Microsoft.OpenApi 3.x.x for all TFMs ([![Latest][openapi-version]][openapi-download])
- [ ] Support OpenAPI 3.2
- [ ] Remove `NU5104` suppression
- [ ] Check `VersionPrefix` is correct for the next major version at the time of merge ([![Next][next-xml]][next-code])
- [ ] Remove any members decorated with `[Obsolete]` that are appropriate ([code search][obsolete])
- [ ] Update documentation samples
- [ ] Create migration guide

---

Prerelease builds from this PR can be consumed from MyGet:

```xml
<PackageReference Include="Swashbuckle.AspNetCore" Version="11.0.0-pr.3786.*" />
```

```xml
<?xml version="1.0" encoding="utf-8"?>
<configuration>
  <packageSources>
    <clear />
    <add key="domaindrivendev" value="https://www.myget.org/F/domaindrivendev/api/v3/index.json" />
    <add key="NuGet" value="https://api.nuget.org/v3/index.json" />
  </packageSources>
</configuration>
```

[next-code]: https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/HEAD/Directory.Build.props#L40
[next-xml]: https://img.shields.io/badge/dynamic/xml?url=https%3A%2F%2Fraw.githubusercontent.com%2Fdomaindrivendev%2FSwashbuckle.AspNetCore%2FHEAD%2FDirectory.Build.props&query=(%2F%2FProject%2FPropertyGroup%2FVersionPrefix)%5B1%5D&logo=github&label=CI
[obsolete]: https://github.com/search?q=repo%3Adomaindrivendev%2FSwashbuckle.AspNetCore+language%3Acsharp+%22%5BObsolete%22+NOT+path%3Atest&type=code
[openapi-version]: https://img.shields.io/nuget/v/Microsoft.OpenApi?logo=nuget&label=Latest&color=blue
[openapi-download]: https://www.nuget.org/packages/Microsoft.OpenApi/

PR #4067, opened 2026-07-18
Downgrade Microsoft.OpenApi

### LINKED ISSUE #4065: [Bug]: Broken nullability in .NET 11
### Describe the bug

Building against #3786, currently at 11.0.0-pr.3786.2844, I notice that `"nullable": true` disappears next to `allOf` in many `openapi.json` files. In other cases, it is replaced by `"enum": [ null ]` (while the type is not an enum).

This is a regression.

Image

Image

### Expected behavior

No changes in `openapi.json` files.

### Actual behavior

_No response_

### Steps to reproduce

_No response_

### Exception(s) (if any)

_No response_

_No response_

### .NET Version

_No response_

### Anything else?

_No response_

bkoelman:
# Pull Request

Workaround for https://github.com/microsoft/OpenAPI.NET/issues/2967

Fixes #4065.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -17,8 +17,8 @@
     <PackageVersion Include="Microsoft.CodeAnalysis.PublicApiAnalyzers" Version="5.6.0" />
     <PackageVersion Include="Microsoft.EntityFrameworkCore.InMemory" Version="11.0.0-preview.6.26359.118" />
     <PackageVersion Include="Microsoft.Kiota.Bundle" Version="2.0.0" />
-    <PackageVersion Include="Microsoft.OpenApi" Version="3.9.0" />
-    <PackageVersion Include="Microsoft.OpenApi.YamlReader" Version="3.9.0" />
+    <PackageVersion Include="Microsoft.OpenApi" Version="3.7.0" />
+    <PackageVersion Include="Microsoft.OpenApi.YamlReader" Version="3.7.0" />
     <PackageVersion Include="Microsoft.Playwright" Version="1.61.0" />
     <PackageVersion Include="Microsoft.NET.Test.Sdk" Version="18.8.1" />
     <PackageVersion Include="MSBuild.ProjectCreation" Version="19.1.4" />
--- a/src/Swashbuckle.AspNetCore.ApiTesting/Swashbuckle.AspNetCore.ApiTesting.csproj
+++ b/src/Swashbuckle.AspNetCore.ApiTesting/Swashbuckle.AspNetCore.ApiTesting.csproj
@@ -10,7 +10,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.OpenApi" VersionOverride="2.11.0" />
+    <PackageReference Include="Microsoft.OpenApi" VersionOverride="2.9.0" />
     <PackageReference Include="Newtonsoft.Json" />
   </ItemGroup>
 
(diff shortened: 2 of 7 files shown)


bkoelman:
This should be good to go now. The workaround to unblock my testing today is [too ugly](https://github.com/json-api-dotnet/JsonApiDotNetCore/pull/1929/changes/1ea731e91350a7a4eba463b1c6056af50608ac65) to share with our downstream consumers, who would need to apply it as well.

bkoelman:
Can this be merged? Going back one version is insufficient.

martincostello:
First I'm curious as to what the difference is between here and your project is, because our tests only showed any change in behaviour when going from 2.10/3.8 to 2.11/3.9 (https://github.com/domaindrivendev/Swashbuckle.AspNetCore/pull/3786/commits/889b987a8cbd5e613f2aeec1fd7118ee04bb9ae6), not earlier when upgrading to 2.9/3.7 to 2.10/3.8 (https://github.com/domaindrivendev/Swashbuckle.AspNetCore/pull/3786/commits/5e565177358c7a225339f7ba27639b309e47e621).

Suggests that either there are two distinct issues going on, and we only have test coverage that finds one of them, or that the one issue somehow got worse.

bkoelman:
It's about the broken `meta` case in the upstream issue I reported. While Swashbuckle itself may be agnostic about it, or lack test coverage, it breaks consumers that rely on logic provided by MS.OpenApi, for example when generating their own component schemas.

martincostello:
So does that mean that the changes I previously made in https://github.com/domaindrivendev/Swashbuckle.AspNetCore/pull/3786/commits/cc099e8fb7379ed5388107e7f207b93ccfc74784 to get the CI in #3786 green after 889b987a8cbd5e613f2aeec1fd7118ee04bb9ae6 have nothing to do with the issue you've reported in Microsoft.OpenApi?

I don't see the word `meta` anywhere in the updated snapshots, so if it's the same issue there's more to it than just that. Otherwise it's either a different unreported issue, or it's not actually a bug and a fix to an incorrect behaviour.

bkoelman:
See https://github.com/microsoft/OpenAPI.NET/issues/2967, first part.

martincostello:
Sorry, I still don't follow. Your issue is 3.7 ➡️ 3.9, `meta` loses information about nullability.

The commit I referenced is 3.8 ➡️ 3.9 also loses some information about nullability, but meta isn't involved.

So if they're the same underlying issue, what changed in 3.8 to make it appear here when it would have already an issue for you since https://github.com/domaindrivendev/Swashbuckle.AspNetCore/commit/5e565177358c7a225339f7ba27639b309e47e621 was committed on the 7th of July? Otherwise they're different issues with the same net effect, so there's an issue or additional repro missing.

bkoelman:
In short: meta broke in 3.8, then 3.9 broke more. Only the latter is observed in Swashbuckle tests. Same story for 2.x.

martincostello:
So the second option here I already said?

> Suggests that either there are two distinct issues going on, and we only have test coverage that finds one of them, **or that the one issue somehow got worse**.

So that suggests that the issue needs some extra context so that whoever picks it up and fixes it handles both manifestations.

I wouldn't want your specific issue to be fixed in a 2.12/3.9 but not the other one, because ultimately you'd be in the same position with broken nullability.

That extra detail is also useful to Microsoft.OpenApi, because not only did they miss the first regression, but they made it worst and still didn't know they'd broken anything.

bkoelman:
> I wouldn't want your specific issue to be fixed in a 2.12/3.9 but not the other one, because ultimately you'd be in the same position with broken nullability.

Yes, that is why I created a single issue describing both bugs.

martincostello:
That's not clear to me from your previous replies, nor in the issue description.

martincostello:
https://github.com/domaindrivendev/Swashbuckle.AspNetCore/pull/3786/commits/a748bf89a5c5f71786c824ae398d8d0ee719e98c

bkoelman:
Thanks

Issue #4083, opened 2026-07-22
[Question]: How to customize index.html location


superstator:
### What are you wanting to achieve?

I am trying to customize the endpoint for accessing Swagger and Swagger UI. I've tried this both via the `UseSwagger` methods and the endpoint routing `MapSwagger` methods. The JSON OAS document is available at the new URL, and fetching `.../swagger/` returns a 301 to `.../swagger/index.html`, but `.../swagger/index.html` itself returns a 404. What could I be missing?

### What code or approach do you have so far?

```csharp
const string routePrefix = "/api";

// snip lots of other setup

app.MapSwagger($"{routePrefix}/swagger/{{documentName}}.json");
app.MapSwaggerUI($"{routePrefix}/swagger/",
        options => options.SwaggerEndpoint($"{routePrefix}/swagger/v1.json", "Cool API V1"));
```

```
$ curl -v http://localhost:5068/api/swagger/v1.json
* Host localhost:5068 was resolved.
* IPv6: ::1
* IPv4: 127.0.0.1
*   Trying 127.0.0.1:5068...
* Connected to localhost (127.0.0.1) port 5068
> GET /api/swagger/v1.json HTTP/1.1
> Host: localhost:5068
> User-Agent: curl/8.5.0
> Accept: */*
> 
< HTTP/1.1 200 OK
< Content-Type: application/json;charset=utf-8
< Date: Wed, 22 Jul 2026 18:54:40 GMT
< Server: Kestrel
< Transfer-Encoding: chunked
< 
{
  "openapi": "3.0.4",
  ...
}

$ curl -v http://localhost:5068/api/swagger/
* Host localhost:5068 was resolved.
* IPv6: ::1
* IPv4: 127.0.0.1
*   Trying 127.0.0.1:5068...
* Connected to localhost (127.0.0.1) port 5068
> GET /api/swagger/ HTTP/1.1
> Host: localhost:5068
> User-Agent: curl/8.5.0
> Accept: */*
> 
< HTTP/1.1 301 Moved Permanently
< Content-Length: 0
< Date: Wed, 22 Jul 2026 18:54:43 GMT
< Server: Kestrel
< Location: index.html
< 

$ curl -v http://localhost:5068/api/swagger/index.html
* Host localhost:5068 was resolved.
* IPv6: ::1
* IPv4: 127.0.0.1
*   Trying 127.0.0.1:5068...
* Connected to localhost (127.0.0.1) port 5068
[...]

PR #4084, opened 2026-07-22
Handle leading slash on route prefix during index.html GET

superstator:
# Pull Request

Minor bug encountered in #4083

This change modifies the regular expression used in `SwaggerUIMiddleware.cs` (line 58) to allow matching when `_options.RoutePrefix` contains a leading `/`. A similar regular expression is used on line 46 to match the route and return a redirect to `index.html`, but it already handles the presence or absence of a leading `/`, so this change simply brings them more in sync.

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerUI/SwaggerUIMiddleware.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerUI/SwaggerUIMiddleware.cs
@@ -55,7 +55,7 @@ internal sealed partial class SwaggerUIMiddleware
                 return;
             }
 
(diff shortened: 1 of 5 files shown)


martincostello (approved):
Thanks - would you mind adding a test somewhere that would have caught this?

superstator:
~Sure, I didn't see a unit test project for SwaggerUI specifically, should this test go somewhere else or would you like me to add a new test project for SwaggerUI?~

Nevermind, I see it's covered by integration tests. I'll try to fit something in there to include a route prefix with leading `/`.

superstator:
I didn't love the idea of creating new test sites just for the route prefix, so instead I added some extra `Startup` classes to the `Basic` site. Let me know if you'd prefer a different approach.

martincostello:
Thanks!

PR #4092, opened 2026-08-03
Adding HTTP new QUERY method support

jaswasin:
# Pull Request

Adds support for the HTTP `QUERY` method (the emerging IETF-draft method for
safe, idempotent requests that carry a request body). Today, an action routed
to `QUERY` causes Swagger generation to throw
`The "QUERY" HTTP method is not supported.`, so the endpoint can't be
documented at all.

This aligns with the project goal of describing *actual* API behavior: if a
route is bound to `QUERY`, that operation should appear in the generated
OpenAPI document.

- **SwaggerGen** — added `"QUERY"` to `OperationTypeMap` in `SwaggerGenerator`,
  so `QUERY`-bound actions are emitted as a `query` path-item operation. Uses
  `HttpMethod.Query` on `net10.0` (where the property exists) and falls back to
  `new HttpMethod("QUERY")` on `net8.0`/`net9.0` via a `NET10_0_OR_GREATER`
  conditional.
- **SwaggerUI** — added `Query` to the `SubmitMethod` enum so the generated
  `supportedSubmitMethods` config can include `"query"` (enabling "Try it out"
  for `QUERY` operations). Public API surface recorded in
  `PublicAPI.Unshipped.txt`.
- **Tests** — `SwaggerGeneratorTests` covers `QUERY` operation generation
  (both `"query"`/`"QUERY"` casing); `RequestValidatorTests` covers `QUERY`
  request validation. Added tests pass locally.

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
+++ b/src/Swashbuckle.AspNetCore.SwaggerGen/SwaggerGenerator/SwaggerGenerator.cs
@@ -1093,6 +1093,11 @@ public class SwaggerGenerator(
         ["HEAD"] = HttpMethod.Head,
         ["PATCH"] = HttpMethod.Patch,
         ["TRACE"] = HttpMethod.Trace,
+#if NET10_0_OR_GREATER
+        ["QUERY"] = HttpMethod.Query,
+#else
+        ["QUERY"] = new HttpMethod("QUERY"),
+#endif
     };
 
     private static readonly Dictionary<BindingSource, ParameterLocation> ParameterLocationMap = new()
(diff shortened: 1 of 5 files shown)


martincostello (commented):
1. Who is the author of this PR? Is it @jaswasin or @jaswantsingh1306?
2. QUERY is only natively supported by OpenAPI 3.2, which Swashbuckle does not yet support - previous versions need to emit as described [here](https://github.com/dotnet/core/blob/main/release-notes/11.0/preview/preview4/aspnetcore.md#http-query-in-generated-openapi-documents) - there are no tests that verify this behaviour in actual rendered/serialized OpenAPI documents.
3. Do SwaggerUI and ReDoc render documents containing QUERY methods correctly?

Issue #3306, opened 2025-03-17
[Bug]: AddServer only adds server to one API version


evochriso:
### Describe the bug

I have 2 versions of my API so I'm using the following code to generate 2 swagger docs, one per version:

``` c#
c.SwaggerDoc("v1", new OpenApiInfo
{
    Title = "API v1",
    Version = "v1",
    Description = "API v1"
});

c.SwaggerDoc("v2", new OpenApiInfo
{
    Title = "API v2",
    Version = "v2",
    Description = "API v2"
});
```

This works as expected and generates 2 different swagger docs, and Swagger UI correctly shows the different docs. When publishing the application, I need to add the server and authentication settings to the generated docs. I am using this code to add the server

``` c#
c.AddServer(new OpenApiServer
{
    Url = server
});
```

and this code to add auth:

``` c#
c.AddSecurityDefinition("oauth2", new OpenApiSecurityScheme
{
    Type = SecuritySchemeType.OAuth2,
    In = ParameterLocation.Header,
    Flows = new OpenApiOAuthFlows
    {
        Implicit = new OpenApiOAuthFlow
        {
            AuthorizationUrl = Configuration.GetValue<Uri>(authUri),
            Scopes = new Dictionary<string, string> { { "user_impersonation", "Access API" } }
        }
    }
});
c.AddSecurityRequirement(new OpenApiSecurityRequirement
{{
    new OpenApiSecurityScheme
    {
        Reference = new OpenApiReference
        {
            Type = ReferenceType.SecurityScheme,
            Id = "oauth2"
        },
        Scheme = "oauth2",
        Name = "oauth2",
        In = ParameterLocation.Header
    }, Array.Empty<string>()
}});
```

This produces the expected v1 doc with the server and auth scheme:

![Image](https://github.com/user-attachments/assets/9f85fc1f-1c87-42ba-be6e-415dbc201865)

However the v2 doc does not have the server or auth URL. 

![Image](https://github.com/user-attachments/assets/ec6dfd9f-2d6c-4a59-bea0-078b0e240eec)

Is this perhaps a bug, or is there additional config code I needed when generating multiple swagger docs for versioning?

### Expected behavior

The current code should produce 2 docs with identical server and authentication info since the AddServer() function doesn't allow targeting a specific doc.

### Swashbuckle.AspNetCore version

7.3.1

### .NET Version

net8.0

RaphaelFakhri:
[...]

PR #4115, opened 2026-08-18
Add coverage for AddServer across multiple documents

RaphaelFakhri:
`GetSwagger_SupportsOption_Servers` generates a single document, so nothing currently pins that servers configured through `AddServer` reach every document rather than only the first one requested. This adds a two-document variant that asserts it for both.

Test only, no source changes. `Swashbuckle.AspNetCore.SwaggerGen.Test` passes on net8.0, net9.0 and net10.0 with 761 tests, 0 failed.

I wrote this while looking into #3306 and thought the coverage was worth keeping regardless of where that issue lands; it asserts current behaviour and is not a fix for anything.


martincostello (approved):
Thanks.

PR #4116, opened 2026-08-18
Use Minimal APIs JSON options when MVC is not registered

### LINKED ISSUE #3097: [Bug]: Property casing policy is ignored when using minimal APIs, camel casing always applied
### Describe the bug

No matter which naming policy is set in general JSON serializer options when using minimal APIs, Swashbuckle applies camel casing. It therefore does not comply with the actual output of the endpoint.

### Expected behavior

With the null naming policy, the property name should have been taken without modification, i.e. it should have been `HelloWorld`.

### Actual behavior

The property name was modified to camel case.

This is what I get in the document:
```json
  "components": {
    "schemas": {
      "MyDto": {
        "type": "object",
        "properties": {
          "helloWorld": {
            "type": "string",
            "nullable": true,
            "readOnly": true
          }
        },
        "additionalProperties": false
      }
    }
  }
```

### Steps to reproduce

Using this code in Program.cs:

```csharp
var builder = WebApplication.CreateSlimBuilder(args);

builder.Services.ConfigureHttpJsonOptions(options =>
{
    options.SerializerOptions.PropertyNamingPolicy = null;
});

builder.Services.AddEndpointsApiExplorer();
builder.Services.AddSwaggerGen();

// This needs to be added because of some other open issue
builder.Services.Configure<RouteOptions>(
    options => options.SetParameterPolicy<RegexInlineRouteConstraint>("regex"));

builder.WebHost.UseKestrelHttpsConfiguration();

var app = builder.Build();

if (app.Environment.IsDevelopment())
{
    app.UseSwagger();
    app.UseSwaggerUI();
}

app.MapGet("/hello", () => new MyDto());

app.Run();

internal sealed class MyDto
{
    public string HelloWorld => "hello";
}
```

### Exception(s) (if any)

_No response_

6.8.1

### .NET Version

8.0.400

### Anything else?

_No response_

RaphaelFakhri:
Fixes #3097.

`IOptions<Microsoft.AspNetCore.Mvc.JsonOptions>` resolves in any application, whether or not MVC was ever registered, so the existing fallback to `IOptions<Microsoft.AspNetCore.Http.Json.JsonOptions>` never ran: in a Minimal APIs application the MVC defaults were returned and the options set through `ConfigureHttpJsonOptions` were ignored. The generated document then contradicted what the endpoints actually serialize, which is the case in the report: the endpoint returns `{"HelloWorld":"hi"}` while the schema documents `helloWorld`.

The MVC options are now preferred only when MVC is actually registered, detected through `IControllerFactory`. I checked that it is registered by `AddControllers()`, `AddMvcCore()` and `AddMvc()`, and by nothing a Minimal APIs application uses, including `AddEndpointsApiExplorer()`, which is why `IActionDescriptorCollectionProvider` is not the marker here.

When an application registers both, MVC keeps precedence exactly as before: one document is generated with one set of serializer options, so a tie-break is unavoidable and this is the one that does not change existing behaviour. An application that configures neither is unaffected, since both option types default to the same web defaults.

Four tests in `SwaggerGenServiceCollectionExtensionsTests`: Minimal APIs only, MVC only, both registered, and neither configured. `Swashbuckle.AspNetCore.SwaggerGen.Test` passes on net8.0, net9.0 and net10.0 with 764 tests, 0 failed.

It turns out the repo's own Minimal APIs sample already demonstrates the bug: `test/WebSites/WebApi` adds a `JsonStringEnumConverter` through `Microsoft.AspNetCore.Http.Json.JsonOptions` and never registers MVC, but the committed snapshots document `DateTimeKind` as an int32 enum (0, 1, 2) while the endpoints actually serialize "Unspecified", "Utc" and "Local". With this change the generated document matches what the sample really returns, and the six snapshots are updated accordingly; every changed hunk is that one schema.

This deliberately takes the no-new-public-API route rather than reviving what #3378 attempted, since that PR was closed with an invitation to start fresh.


test/Swashbuckle.AspNetCore.SwaggerGen.Test/SwaggerGenServiceCollectionExtensionsTests.cs
```diff
@@ -0,0 +1,89 @@
...
+public static class SwaggerGenServiceCollectionExtensionsTests
+{
+    [Fact]
+    public static void AddSwaggerGen_UsesMinimalApiJsonOptions_WhenMvcIsNotRegistered()
+    {
+        // IOptions<Mvc.JsonOptions> resolves even in an application that never registered MVC,
+        // so without a registration check its camelCase default would win over what the
+        // application actually configured for Minimal APIs.
+        var services = CreateServices();
+        services.ConfigureHttpJsonOptions(options => options.SerializerOptions.PropertyNamingPolicy = null);
```
martincostello: I wonder if these tests would be more obvious if they assigned a specific custom policy that would unambiguously indicate it is being used (or not).

For example, something that made it `hElLoWoRlD`, added a constant suffix, etc.

RaphaelFakhri: Good idea. Both paths now assign a policy that suffixes the property name, so the assertions name the source directly: HelloWorld-minimal-api when only Minimal APIs is registered and HelloWorld-mvc when MVC is. The both-registered test now sets a different policy on each, so it shows MVC winning rather than just showing camelCase. Verified the first test fails against master's version of the extension.

test/Swashbuckle.AspNetCore.SwaggerGen.Test/SwaggerGenServiceCollectionExtensionsTests.cs
martincostello, line 1: The changes to the tests in the most recent commit have had a wider scope of change and changed a bunch of snapshots - can you refactor the change to be more scoped to just the specific tests please? I wouldn't expect there to be any impact beyond this test class.

PR #4127, opened 2026-08-25
Document that System.Text.Json ignores [EnumMember]

### LINKED ISSUE #3328: [Bug]: System.Text.Json based serializer doesn't respect EnumMember
### Describe the bug

I have a schema which includes an enum with `JsonStringEnumConverter` and a member annotated with `EnumMember` to supply an alternate value for de/serialization:

```csharp
[JsonConverter(typeof(JsonStringEnumConverter))]
public enum JsonConverterAnnotatedEnum
{
    Value1 = 2,
    Value2 = 4,
    [EnumMember(Value = "X-foo")]
    X = 8
}
```

### Expected behavior

The resulting JSON schema should honor value of `EnumMember`, i.e.:

```json
{
  "type": "string",
  "enum": [
    "Value1",
    "Value2",
    "X-foo"
  ]
}
```

### Actual behavior

`EnumMember` attribute is ignored and the enum name is used instead in the resulting schema:

```json
{
  "type": "string",
  "enum": [
    "Value1",
    "Value2",
    "X"
  ]
}
```

### Steps to reproduce

This can be reproduced using SwaggerGen's tests (in fact, the tests for Newtonsoft.Json based serializer [already cover this use case](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/99ccafac6dac6b1a2e0a11713ae55186ef764b71/test/Swashbuckle.AspNetCore.Newtonsoft.Test/SchemaGenerator/NewtonsoftSchemaGeneratorTests.cs#L757-L766)).

Here is an example commit modifying the tests:

1. Add `[EnumMember(Value = "X-foo")]` attribute to `JsonConverterAnnotatedEnum` in `Swashbuckle.AspNetCore.SwaggerGen.Test`
2. Modify the `GenerateSchema_HonorsSerializerAttribute_StringEnumConverter` in `test/Swashbuckle.AspNetCore.SwaggerGen.Test/SchemaGenerator/JsonSerializerSchemaGeneratorTests.cs` to assert `X-foo` value
3. Run the `Swashbuckle.AspNetCore.SwaggerGen.Test` project

### Exception(s) (if any)

_No response_

8.0.0

### .NET Version

9.0.202

### Anything else?

This issue was previously fixed in #589 for Newtonsoft.Json serializer. So it seems like a missing feature parity in `System.Text.Json`-based serializer.

RaphaelFakhri:
Fixes #3328.

Adds a `Rename Enum Values in Schemas` section to the SwaggerGen docs, covering the difference you agreed was worth writing down: `System.Text.Json` does not implement `[EnumMember]`, `Newtonsoft.Json` does, and `[JsonStringEnumMemberName]` is the native option.

I verified all three claims against `master` rather than trusting the issue, by generating the schema for the same enum through each resolver:

| Serializer | Attribute | `enum` in the generated schema |
|---|---|---|
| `System.Text.Json` | `[EnumMember(Value = "X-foo")]` | `Value1`, `X` |
| `System.Text.Json` | `[JsonStringEnumMemberName("X-foo")]` | `Value1`, `X-foo` |
| `Newtonsoft.Json` | `[EnumMember(Value = "X-foo")]` | `Value1`, `X-foo` |

The C# examples are plain fenced blocks rather than MarkdownSnippets snippets. `JsonStringEnumMemberNameAttribute` is .NET 9 and later, and `DocumentationSnippets` multi-targets down to net8.0, so a compiled snippet would need an `#if NET9_0_OR_GREATER` around it that would then show up in the rendered page. Happy to switch it to a snippet with the guard, or to a net9.0-and-later snippet project, if you would rather have it compiled.

`markdownlint-cli2` with the repo's `.markdownlint.json` reports 0 issues, and `dotnet mdsnippets` leaves the file unchanged.

### DIFF
--- a/docs/configure-and-customize-swaggergen.md
+++ b/docs/configure-and-customize-swaggergen.md
@@ -750,6 +750,62 @@ services.AddSwaggerGen(options =>
 > [!NOTE]
 > See [this GitHub issue](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/2703) for support for nested types.
 
+## Rename Enum Values in Schemas
+
+When an enum is serialized as a string, the values that appear in the generated schema come from the serializer, not from
+Swashbuckle.AspNetCore. Which attribute renames a member therefore depends on which serializer your application uses, and the
+two do not agree.
+
+`System.Text.Json` does not implement `[EnumMember]`. It is ignored for both serialization and schema generation, so this enum:
+
+```csharp
+[JsonConverter(typeof(JsonStringEnumConverter))]
+public enum Priority
+{
+    Low = 1,
+    [EnumMember(Value = "very-high")]
+    VeryHigh = 2
+}
+```
+
+produces the member name, not the value you asked for:
+
+```json
+{
+  "type": "string",
+  "enum": [
+    "Low",
+    "VeryHigh"
+  ]
+}
+```
+
+Use `[JsonStringEnumMemberName]`, added to `System.Text.Json` in .NET 9, which `System.Text.Json` does implement:
+
+```csharp
+[JsonConverter(typeof(JsonStringEnumConverter))]
+public enum Priority
+{
+    Low = 1,
+    [JsonStringEnumMemberName("very-high")]
+    VeryHigh = 2
+}
+```
+
+```json
+{
+  "type": "string",
+  "enum": [
+    "Low",
+    "very-high"
+  ]
+}
+```
+
+> [!NOTE]
+> `Newtonsoft.Json` does honor `[EnumMember]`, so an application using the `Swashbuckle.AspNetCore.Newtonsoft` package gets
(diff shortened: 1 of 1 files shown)


docs/configure-and-customize-swaggergen.md
```diff
@@ -750,6 +750,62 @@ services.AddSwaggerGen(options =>
 > [!NOTE]
 > See [this GitHub issue](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/2703) for support for nested types.
 
+## Rename Enum Values in Schemas
```
martincostello, line 753: I think this is overly verbose - we don't need to repeat and maintain documentation for how the JSON serializer works.

It's sufficient to link to how to upstream documentation on how rename members for both System.Text.Json and Newtonsoft.Json.

PR #4130, opened 2026-08-25
Fix #2951: drop the regex route constraint from the default Swagger route

### LINKED ISSUE #2951: [Bug]: Breaking change in 6.6.1; It now requires you to have the `Regex` Route Constraint even if you don't use it
### Describe the bug

Prior to 6.5.0 you could use `WebApplication.CreateSlimBuilder` (which, among other things, doesn't add the `Regex` Route constraint due to it increasing the file size of the app needlessly), but after upgrading to 6.6.* (Tested with 6.6.1 and 6.6.2) this will throw an exception.

### Expected behavior

SwashBuckle should not require `RegexRouteConstraint` to exist in order to work.

### Actual behavior

SwashBuckle uses the `Regex` Route constraint despite it not existing, and therefore the app no longer starts.

### Steps to reproduce

https://github.com/KennethHoff/Repros/tree/master/SwashbuckleRegexException

### Exception(s) (if any)

```
 Error Message:
   Microsoft.AspNetCore.Routing.RouteCreationException : An error occurred while trying to create an instance of 'Microsoft.AspNetCore.Routing.Constraints.RegexErrorStubRouteConstraint'.
---- System.Reflection.TargetInvocationException : Exception has been thrown by the target of an invocation.
-------- System.InvalidOperationException : A route parameter uses the regex constraint, which isn't registered. If this application was configured using CreateSlimBuilder(...) or AddRoutingCore(...) then this constraint is not registered by default. To use the regex constraint, configure route options at app startup: services.Configure<RouteOptions>(options => options.SetParameterPolicy<RegexInlineRouteConstraint>("regex"));
  Stack Trace:
     at Microsoft.AspNetCore.Routing.ParameterPolicyActivator.ResolveParameterPolicy[T](IDictionary`2 inlineParameterPolicyMap, IServiceProvider serviceProvider, String inlineParameterPolicy, String& parameterPolicyKey)
   at Microsoft.AspNetCore.Routing.DefaultParameterPolicyFactory.Create(RoutePatternParameterPart parameter, String inlineText)
   at Microsoft.AspNetCore.Routing.Template.DefaultTemplateBinderFactory.Create(RoutePattern pattern)
   at Swashbuckle.AspNetCore.Swagger.SwaggerMiddleware..ctor(RequestDelegate next, SwaggerOptions options, TemplateBinderFactory templateBinderFactory)
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)
   at System.Reflection.MethodBaseInvoker.InvokeDirectByRefWithFewArgs(Object obj, Span`1 copyOfArgs, BindingFlags invokeAttr)
   at System.Reflection.MethodBaseInvoker.InvokeWithFewArgs(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
   at Microsoft.Extensions.Internal.ActivatorUtilities.ConstructorMatcher.CreateInstance(IServiceProvider provider)
   at Microsoft.Extensions.Internal.ActivatorUtilities.CreateInstance(IServiceProvider provider, Type instanceType, Object[] parameters)
   at Microsoft.AspNetCore.Builder.UseMiddlewareExtensions.ReflectionMiddlewareBinder.CreateMiddleware(RequestDelegate next)
   at Microsoft.AspNetCore.Builder.ApplicationBuilder.Build()
   at Microsoft.AspNetCore.Builder.ApplicationBuilder.Build()
   at Microsoft.AspNetCore.Hosting.GenericWebHostService.StartAsync(CancellationToken cancellationToken)
   at Microsoft.Extensions.Hosting.Internal.Host.<StartAsync>b__15_1(IHostedService service, CancellationToken token)
   at Microsoft.Extensions.Hosting.Internal.Host.ForeachService[T](IEnumerable`1 services, CancellationToken token, Boolean concurrent, Boolean abortOnFirstException, List`1 exceptions, Func`3 operation)
   at Microsoft.Extensions.Hosting.Internal.Host.StartAsync(CancellationToken cancellationToken)
   at Microsoft.Extensions.Hosting.HostingAbstractionsHostExtensions.RunAsync(IHost host, CancellationToken token)
   at Microsoft.Extensions.Hosting.HostingAbstractionsHostExtensions.RunAsync(IHost host, CancellationToken token)
   at Program.<Main>$(String[] args) in /Users/kennethhoff/Documents/Oxx/TensioLinea/wcp-sharedbackend-app/src/Linea.Api/Program.cs:line 99
   at Program.<Main>(String[] args)
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)
   at System.Reflection.MethodBaseInvoker.InvokeDirectByRefWithFewArgs(Object obj, Span`1 copyOfArgs, BindingFlags invokeAttr)
--- End of stack trace from previous location ---
   at Microsoft.AspNetCore.Mvc.Testing.DeferredHostBuilder.DeferredHost.StartAsync(CancellationToken cancellationToken)
   at Microsoft.Extensions.Hosting.HostingAbstractionsHostExtensions.Start(IHost host)
   at Microsoft.AspNetCore.Mvc.Testing.WebApplicationFactory`1.CreateHost(IHostBuilder builder)
   at Linea.Endpoints.Tests.Integration.LineaWebApplicationFactory.CreateHost(IHostBuilder builder) in /Users/kennethhoff/Documents/Oxx/TensioLinea/wcp-sharedbackend-app/Tests/Linea.Endpoints.Tests.Integration/LineaWebApplicationFactory.cs:line 24
   at Microsoft.AspNetCore.Mvc.Testing.WebApplicationFactory`1.DelegatedWebApplicationFactory.CreateHost(IHostBuilder builder)
   at Microsoft.AspNetCore.Mvc.Testing.WebApplicationFactory`1.ConfigureHostBuilder(IHostBuilder hostBuilder)
   at Microsoft.AspNetCore.Mvc.Testing.WebApplicationFactory`1.EnsureServer()
   at Microsoft.AspNetCore.Mvc.Testing.WebApplicationFactory`1.CreateDefaultClient(DelegatingHandler[] handlers)
   at Microsoft.AspNetCore.Mvc.Testing.WebApplicationFactory`1.CreateDefaultClient(Uri baseAddress, DelegatingHandler[] handlers)
   at Microsoft.AspNetCore.Mvc.Testing.WebApplicationFactory`1.CreateClient(WebApplicationFactoryClientOptions options)
   at Microsoft.AspNetCore.Mvc.Testing.WebApplicationFactory`1.CreateClient()

```

6.6.1, 6.6.2

### .NET Version

8.0.302

### Anything else?

_No response_

RaphaelFakhri:
Fixes #2951. This takes up the offer above: "We'd be happy to take a PR that changes the route template ... that doesn't need a Regex."

It still reproduces on master. `SwaggerOptions.DefaultRouteTemplate` is `/swagger/{documentName}/swagger.{extension:regex(^(json|ya?ml)$)}`, and `SwaggerMiddleware`'s constructor eagerly resolves a `TemplateBinder` for it. An application built with `CreateSlimBuilder` or `AddRoutingCore` does not register the regex constraint, so a standalone app against master throws at startup with the exception in the report, from `SwaggerMiddleware..ctor`.

The default template drops the constraint, and the middleware applies the equivalent allow-list in code, but only when the `extension` parameter has no inline constraint of its own. A custom template that brings its own constraints keeps going through the `TemplateBinder` path unchanged. The allow-list is `json`, `yaml`, `yml`, compared with `OrdinalIgnoreCase`, matching the case-insensitivity of the regex it replaces.

`MapSwagger`'s default parameter value changes with it, so `PublicAPI.Unshipped.txt` carries the `*REMOVED*` and replacement pair the analyzers require.

This also deletes the repo's own workaround: `test/WebSites/WebApi.Aot/Program.cs` was registering `SetParameterPolicy<RegexInlineRouteConstraint>` to get around exactly this, so the in-repo AoT slim-builder site now exercises the fix rather than papering over it.

`SwaggerMiddlewareRouteTemplateTests` covers building the pipeline under `AddRoutingCore()` with no regex constraint registered, plus eight route cases: `.json`, `.yaml`, `.yml`, `.JSON` and `.YAML` serve 200, and `.txt`, `.html` and `.jsonx` fall through to a 404. All nine fail against master's source with the same `RouteCreationException` from the report. With the change, `Swashbuckle.AspNetCore.SwaggerGen.Test` is 769 passed, 0 failed on net8.0. I also confirmed the standalone slim-builder app end to end: it crashed before, and after the change it starts, serves a valid document at `swagger.json`, serves `swagger.yaml` as `text/yaml`, and 404s `swagger.txt`.

Two things I did not change, deliberately. `.YAML` returns 200 with JSON content, because the old regex was case-insensitive while the json/yaml switch downstream is case-sensitive; that is master's existing behaviour and this PR is not the place to change it. And a user with a custom, unconstrained `{extension}` template, say `docs/{documentName}/spec.{extension}`, previously got JSON for any extension and now sees anything outside the allow-list fall through to the next middleware. That is the same tightening #2418 intended for the default template, but it is a behaviour change and worth your call.

### DIFF
--- a/src/Swashbuckle.AspNetCore.Swagger/PublicAPI/PublicAPI.Unshipped.txt
+++ b/src/Swashbuckle.AspNetCore.Swagger/PublicAPI/PublicAPI.Unshipped.txt
@@ -1,1 +1,3 @@
 ﻿
+*REMOVED*static Microsoft.AspNetCore.Builder.SwaggerBuilderExtensions.MapSwagger(this Microsoft.AspNetCore.Routing.IEndpointRouteBuilder endpoints, string pattern = "/swagger/{documentName}/swagger.{extension:regex(^(json|ya?ml)$)}", System.Action<Swashbuckle.AspNetCore.Swagger.SwaggerEndpointOptions> setupAction = null) -> Microsoft.AspNetCore.Builder.IEndpointConventionBuilder
+static Microsoft.AspNetCore.Builder.SwaggerBuilderExtensions.MapSwagger(this Microsoft.AspNetCore.Routing.IEndpointRouteBuilder endpoints, string pattern = "/swagger/{documentName}/swagger.{extension}", System.Action<Swashbuckle.AspNetCore.Swagger.SwaggerEndpointOptions> setupAction = null) -> Microsoft.AspNetCore.Builder.IEndpointConventionBuilder
--- a/test/Swashbuckle.AspNetCore.SwaggerGen.Test/SwaggerMiddlewareRouteTemplateTests.cs
+++ b/test/Swashbuckle.AspNetCore.SwaggerGen.Test/SwaggerMiddlewareRouteTemplateTests.cs
@@ -1,0 +1,76 @@
+using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.OpenApi;
+using NSubstitute;
+using Swashbuckle.AspNetCore.Swagger;
+
+namespace Swashbuckle.AspNetCore.SwaggerGen.Test;
+
+public static class SwaggerMiddlewareRouteTemplateTests
+{
+    [Fact]
+    public static void UseSwagger_Does_Not_Require_Regex_Route_Constraint()
+    {
+        // See https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/2951.
+        // AddRoutingCore(), as used by WebApplication.CreateSlimBuilder(), does not
+        // register the regex route constraint, so building the pipeline must not
+        // require it for the default route template.
+        var (pipeline, _) = CreatePipeline();
+        Assert.NotNull(pipeline);
+    }
+
+    [Theory]
+    [InlineData("/swagger/v1/swagger.json", 200, "application/json;charset=utf-8")]
+    [InlineData("/swagger/v1/swagger.yaml", 200, "text/yaml;charset=utf-8")]
+    [InlineData("/swagger/v1/swagger.yml", 200, "text/yaml;charset=utf-8")]
+    [InlineData("/swagger/v1/swagger.JSON", 200, "application/json;charset=utf-8")]
+    [InlineData("/swagger/v1/swagger.YAML", 200, "application/json;charset=utf-8")]
+    [InlineData("/swagger/v1/swagger.txt", 404, null)]
+    [InlineData("/swagger/v1/swagger.html", 404, null)]
+    [InlineData("/swagger/v1/swagger.jsonx", 404, null)]
+    public static async Task UseSwagger_Default_Route_Template_Only_Matches_Supported_Extensions(
+        string path,
+        int expectedStatusCode,
+        string expectedContentType)
+    {
+        var (pipeline, services) = CreatePipeline();
+
+        using var scope = services.CreateScope();
+
+        var context = new DefaultHttpContext()
+        {
+            RequestServices = scope.ServiceProvider,
+        };
+
+        context.Request.Method = HttpMethods.Get;
+        context.Request.Path = path;
+        context.Response.Body = new MemoryStream();
+
+        await pipeline(context);
+
+        Assert.Equal(expectedStatusCode, context.Response.StatusCode);
+        Assert.Equal(expectedContentType, context.Response.ContentType);
+    }
+
+    private static (RequestDelegate Pipeline, IServiceProvider Services) CreatePipeline()
+    {
+        var swaggerProvider = Substitute.For<ISwaggerProvider>();
+        swaggerProvider
+            .GetSwagger("v1", Arg.Any<string>(), Arg.Any<string>())
+            .Returns(new OpenApiDocument()
+            {
+                Info = new OpenApiInfo() { Title = "Test API", Version = "v1" },
+            });
+        var services = new ServiceCollection()
+            .AddRoutingCore()
+            .AddSingleton(swaggerProvider)
+            .BuildServiceProvider();
+
+        var app = new ApplicationBuilder(services);
+        app.UseSwagger(new SwaggerOptions());
+
+        return (app.Build(), services);
+    }
+}
+
--- a/src/Swashbuckle.AspNetCore.Swagger/SwaggerMiddleware.cs
+++ b/src/Swashbuckle.AspNetCore.Swagger/SwaggerMiddleware.cs
@@ -14,10 +14,13 @@ internal sealed class SwaggerMiddleware
     private static readonly Encoding UTF8WithoutBom = new UTF8Encoding(false);
     private static readonly HashSet<string> AllowedHttpMethods = new(StringComparer.OrdinalIgnoreCase) { HttpMethods.Get, HttpMethods.Head };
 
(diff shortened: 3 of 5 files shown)


src/Swashbuckle.AspNetCore.Swagger/PublicAPI/PublicAPI.Unshipped.txt
```diff
@@ -1 +1,3 @@
-﻿
\ No newline at end of file
+﻿
+*REMOVED*static Microsoft.AspNetCore.Builder.SwaggerBuilderExtensions.MapSwagger(this Microsoft.AspNetCore.Routing.IEndpointRouteBuilder endpoints, string pattern = "/swagger/{documentName}/swagger.{extension:regex(^(json|ya?ml)$)}", System.Action<Swashbuckle.AspNetCore.Swagger.SwaggerEndpointOptions> setupAction = null) -> Microsoft.AspNetCore.Builder.IEndpointConventionBuilder
+static Microsoft.AspNetCore.Builder.SwaggerBuilderExtensions.MapSwagger(this Microsoft.AspNetCore.Routing.IEndpointRouteBuilder endpoints, string pattern = "/swagger/{documentName}/swagger.{extension}", System.Action<Swashbuckle.AspNetCore.Swagger.SwaggerEndpointOptions> setupAction = null) -> Microsoft.AspNetCore.Builder.IEndpointConventionBuilder
```
martincostello, line 3: Looks like a breaking change, but isn't - it just changes the default value for `pattern`.

test/Swashbuckle.AspNetCore.SwaggerGen.Test/SwaggerMiddlewareRouteTemplateTests.cs
```diff
@@ -0,0 +1,76 @@
...
+
+public static class SwaggerMiddlewareRouteTemplateTests
+{
+    [Fact]
+    public static void UseSwagger_Does_Not_Require_Regex_Route_Constraint()
+    {
+        // See https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/2951.
+        // AddRoutingCore(), as used by WebApplication.CreateSlimBuilder(), does not
+        // register the regex route constraint, so building the pipeline must not
+        // require it for the default route template.
```
martincostello, line 18:
```suggestion
```

test/Swashbuckle.AspNetCore.SwaggerGen.Test/SwaggerMiddlewareRouteTemplateTests.cs
```diff
@@ -0,0 +1,76 @@
...
+using Microsoft.OpenApi;
+using NSubstitute;
+using Swashbuckle.AspNetCore.Swagger;
+
+namespace Swashbuckle.AspNetCore.SwaggerGen.Test;
+
+public static class SwaggerMiddlewareRouteTemplateTests
+{
+    [Fact]
+    public static void UseSwagger_Does_Not_Require_Regex_Route_Constraint()
```
martincostello, line 13: Better described as what it works with, than what it doesn't need.

```suggestion
    public static void UseSwagger_Works_With_AddRoutingCore()
```

PR #2890, opened 2024-05-15
Fix serialization of AdditionalItems

martincostello:
- Extend custom `JsonSerializerContext` to cover objects that might be added to the additional JSON property.
- Speed up local development iterations by only running `npm install` when the `node_modules` directory does not already exist.
- Apply code refactoring suggestions from Visual Studio.

Resolves #2884.

PR #4132, opened 2026-08-25
Fix #3153: serialize custom values in the UI ConfigObject AdditionalItems

### LINKED ISSUE #3153: [Bug]: AdditionalItems serialization issue still present
### Describe the bug

originally discussed in https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/2884

the fix was to add a bunch of types, but the approach documented for modifying the theme is not fixed by that approach.

image

a complex anonymous object is not an understood type.

### Expected behavior

_No response_

### Actual behavior

_No response_

### Steps to reproduce

_No response_

### Exception(s) (if any)

_No response_

6.6+

### .NET Version

_No response_

### Anything else?

_No response_

RaphaelFakhri:
Fixes #3153.

This is not fixed by the Microsoft.OpenApi v2 move, which only changed how the OpenAPI document is serialized. The failure is in the UI config serialization and still happens on master. With `c.ConfigObject.AdditionalItems.Add("theme", new { colors = new { primary = "#086eaa" } })`, exercised through the real middleware:

```
NotSupportedException: JsonTypeInfo metadata for type '<>f__AnonymousType0`1[...]' was not
provided by TypeInfoResolver of type 'ReDocOptionsJsonContext'
```

and the identical failure from `SwaggerUIOptionsJsonContext`.

Both middlewares serialize `ConfigObject` with the source-generated context alone whenever the application has not supplied its own `JsonSerializerOptions`. #2890 whitelisted primitives in those contexts, but `AdditionalItems` is `[JsonExtensionData]`, so any complex value in it has no metadata and throws.

When `RuntimeFeature.IsDynamicCodeSupported`, the default options now chain a reflection-based resolver after the source-generated one:

```csharp
TypeInfoResolver = JsonTypeInfoResolver.Combine(
    ReDocOptionsJsonContext.Default,
    new DefaultJsonTypeInfoResolver())
```

Registered types still resolve through source generation first, so the emitted config is byte-identical for everything that worked before, including the camelCase naming. On native AoT the resolver stays source-generated only, so AoT support is untouched. Application-supplied `JsonSerializerOptions` still win, as they did. No public API change.

Two regression tests, one per middleware, styled after the adjacent splice tests. Against master's source both fail with the exception above. With the change, `Swashbuckle.AspNetCore.IntegrationTests` is 224 passed, 0 failed on net8.0.

The tradeoff worth stating: `RuntimeFeature.IsDynamicCodeSupported` distinguishes native AoT, not trimming, so in a trimmed-but-JIT application a user's custom config type could in principle be trimmed away and serialize incompletely. That is the same exposure as the pre-6.6 behaviour this restores, and it matches the suppression pattern already used in these two files.

### DIFF
--- a/src/Swashbuckle.AspNetCore.ReDoc/ReDocMiddleware.cs
+++ b/src/Swashbuckle.AspNetCore.ReDoc/ReDocMiddleware.cs
@@ -1,8 +1,10 @@
 ﻿using System.Diagnostics.CodeAnalysis;
 using System.Reflection;
+using System.Runtime.CompilerServices;
 using System.Security.Cryptography;
 using System.Text;
 using System.Text.Json;
+using System.Text.Json.Serialization.Metadata;
 using System.Text.RegularExpressions;
 using Microsoft.AspNetCore.Http;
 
@@ -12,6 +14,7 @@ internal sealed partial class ReDocMiddleware
 {
     private static readonly HashSet<string> AllowedHttpMethods = new(StringComparer.OrdinalIgnoreCase) { HttpMethods.Get, HttpMethods.Head };
     private static readonly string ReDocVersion = GetReDocVersion();
+    private static readonly JsonSerializerOptions DefaultJsonSerializerOptions = CreateDefaultJsonSerializerOptions();
 
     private readonly RequestDelegate _next;
     private readonly ReDocOptions _options;
@@ -23,10 +26,7 @@ internal sealed partial class ReDocMiddleware
         _next = next;
         _options = options ?? new ReDocOptions();
 
-        if (options.JsonSerializerOptions != null)
-        {
-            _jsonSerializerOptions = options.JsonSerializerOptions;
-        }
+        _jsonSerializerOptions = _options.JsonSerializerOptions ?? DefaultJsonSerializerOptions;
 
         var pathPrefix = options.RoutePrefix.StartsWith('/') ? options.RoutePrefix : $"/{options.RoutePrefix}";
         _resourceProvider = new(
@@ -70,6 +70,32 @@ internal sealed partial class ReDocMiddleware
         }
 
         await _next(httpContext);
+    }
+
+    [UnconditionalSuppressMessage(
+        "AOT",
+        "IL2026:RequiresUnreferencedCode",
+        Justification = "The reflection-based resolver is only used when dynamic code is supported (i.e. not native AoT) to serialize custom values in ConfigObject.AdditionalItems. See https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/3153.")]
+    [UnconditionalSuppressMessage(
+        "AOT",
+        "IL3050:RequiresDynamicCode",
+        Justification = "The reflection-based resolver is only used when dynamic code is supported (i.e. not native AoT) to serialize custom values in ConfigObject.AdditionalItems. See https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/3153.")]
+    private static JsonSerializerOptions CreateDefaultJsonSerializerOptions()
+    {
+        if (!RuntimeFeature.IsDynamicCodeSupported)
+        {
+            // Only source-generated metadata can be used with native AoT
+            return null;
+        }
+
+        // Chain a reflection-based resolver after the source-generated one so that custom
+        // values in ConfigObject.AdditionalItems (e.g. anonymous types) can be serialized.
+        return new JsonSerializerOptions(ReDocOptionsJsonContext.Default.Options)
+        {
+            TypeInfoResolver = JsonTypeInfoResolver.Combine(
+                ReDocOptionsJsonContext.Default,
+                new DefaultJsonTypeInfoResolver()),
+        };
     }
 
     private static string GetReDocVersion()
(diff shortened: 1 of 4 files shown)


RaphaelFakhri:
On the codecov patch check: the two uncovered lines are the `return null;` inside `if (!RuntimeFeature.IsDynamicCodeSupported)`, one in each middleware. That branch only runs under native AoT, so the normal test run cannot reach it, and covering it would mean publishing an AoT app in the test matrix purely to observe a null.

The other side of the branch, the combined resolver, is what the two new tests exercise, and the AoT behaviour is unchanged from before this PR: the resolver stays source-generated only, exactly as it is on master today. Happy to restructure it if you would rather see that path covered some other way.

src/Swashbuckle.AspNetCore.ReDoc/ReDocMiddleware.cs
```diff
@@ -72,6 +72,32 @@ public async Task Invoke(HttpContext httpContext)
...
     }
 
+    [UnconditionalSuppressMessage(
+        "AOT",
+        "IL2026:RequiresUnreferencedCode",
+        Justification = "The reflection-based resolver is only used when dynamic code is supported (i.e. not native AoT) to serialize custom values in ConfigObject.AdditionalItems. See https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/3153.")]
+    [UnconditionalSuppressMessage(
+        "AOT",
+        "IL3050:RequiresDynamicCode",
+        Justification = "The reflection-based resolver is only used when dynamic code is supported (i.e. not native AoT) to serialize custom values in ConfigObject.AdditionalItems. See https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/3153.")]
```
martincostello, line 82:
```suggestion
    [UnconditionalSuppressMessage(
        "AOT",
        "IL2026:RequiresUnreferencedCode",
        Justification = "The reflection-based resolver is only used when dynamic code is supported.")]
    [UnconditionalSuppressMessage(
        "AOT",
        "IL3050:RequiresDynamicCode",
        Justification = "The reflection-based resolver is only used when dynamic code is supported.")]
```

src/Swashbuckle.AspNetCore.SwaggerUI/SwaggerUIMiddleware.cs
```diff
@@ -82,6 +82,32 @@ public async Task Invoke(HttpContext httpContext)
...
     }
 
+    [UnconditionalSuppressMessage(
+        "AOT",
+        "IL2026:RequiresUnreferencedCode",
+        Justification = "The reflection-based resolver is only used when dynamic code is supported (i.e. not native AoT) to serialize custom values in ConfigObject.AdditionalItems. See https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/3153.")]
+    [UnconditionalSuppressMessage(
+        "AOT",
+        "IL3050:RequiresDynamicCode",
+        Justification = "The reflection-based resolver is only used when dynamic code is supported (i.e. not native AoT) to serialize custom values in ConfigObject.AdditionalItems. See https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/3153.")]
```
martincostello, line 92:
```suggestion
        Justification = "The reflection-based resolver is only used when dynamic code is supported.")]
    [UnconditionalSuppressMessage(
        "AOT",
        "IL3050:RequiresDynamicCode",
        Justification = "The reflection-based resolver is only used when dynamic code is supported.")]
```

test/Swashbuckle.AspNetCore.IntegrationTests/ReDocIntegrationTests.cs
```diff
@@ -433,4 +433,32 @@ public async Task ReDocMiddleware_ConfigObject_Is_Not_Spliced_Into_A_String_Lite
...
+
+    [Fact]
+    public async Task ReDocMiddleware_Serializes_Custom_Object_In_AdditionalItems()
+    {
+        // Arrange
+        var cancellationToken = TestContext.Current.CancellationToken;
+
+        using var server = TestSite.CreateServer((app) => app.UseReDoc((options) =>
+        {
+            // See https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/3153
```
martincostello, line 445:
```suggestion
```

test/Swashbuckle.AspNetCore.IntegrationTests/SwaggerUIIntegrationTests.cs
```diff
@@ -807,4 +807,38 @@ public async Task SwaggerUIMiddleware_ConfigObject_Is_Not_Spliced_Into_A_String_
...
+
+    [Fact]
+    public async Task SwaggerUIMiddleware_Serializes_Custom_Object_In_AdditionalItems()
+    {
+        // Arrange
+        var cancellationToken = TestContext.Current.CancellationToken;
+
+        using var server = TestSite.CreateServer((app) => app.UseSwaggerUI((options) =>
+        {
+            // See https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/3153
```
martincostello, line 819:
```suggestion
```

PR #4152, opened 2026-09-11
Strip query string from relative document URLs in Swagger UI

### LINKED ISSUE #2721: Open redirect finding on request for index.html with query parameters
When scanning an API using a Swashbuckle driven Swagger UI, an information level finding for an [open redirect](https://portswigger.net/kb/issues/00500100_open-redirection-reflected) is detected by BurpSuite because any query string is included in the link to the OpenAPI JSON doc.

Browser request
`https://www.mysite.com/swagger/index.html?foo=bar`

Link in resulting index.html
`https://www.mysite.com/swagger/swagger.json?foo=bar`

This [section of code](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/8f363f7359cb1cb8fa5de5195ec6d97aefaa16b3/src/Swashbuckle.AspNetCore.SwaggerUI/index.html#L80C1-L80C1) replaces the configured URLs with absolute paths using window.location.href.  I do not see this same section of code in the Swagger UI repo so I think this belongs here.  

The petstore example does not seem to exhibit the same behavior.
https://petstore.swagger.io/index.html?foo=bar

If there is no reason to include the query string, this change should address the problem.
`window.location.href.replace("index.html", item.url).split(/[?#]/)[0]`

GabeSilvaDev:
# Pull Request

Fixes #2721

The Swagger UI initializer script (`index.js`) resolves relative document URLs against `window.location.href` and only stripped the fragment, so a request for `swagger/index.html?foo=bar` produced a document link of `swagger/swagger.json?foo=bar`. This is what scanners such as Burp Suite flag as a reflected open-redirect finding, as described in the issue.

The page URL now has its query string and fragment removed (`split(/[?#]/)[0]`) *before* the document URL is spliced in, so nothing from the page URL is carried over. The document URL itself is treated exactly as before: a configured query string such as `v1/swagger.json?tenant=foo` is kept, and only a fragment is dropped. Absolute and root-relative document URLs are not touched.

Added an integration test in `SwaggerUIIntegrationTests` (same style as `SwaggerUIMiddleware_ReturnsInitializerScript`) asserting the served script strips the page URL before the replacement; it failed before the change.

### DIFF
--- a/src/Swashbuckle.AspNetCore.SwaggerUI/index.js
+++ b/src/Swashbuckle.AspNetCore.SwaggerUI/index.js
@@ -36,7 +36,7 @@ window.onload = function () {
     // Workaround for https://github.com/swagger-api/swagger-ui/issues/5945
     configObject.urls.forEach(function (item) {
         if (item.url.startsWith("http") || item.url.startsWith("/")) return;
-        item.url = window.location.href.replace("index.html", item.url).split(/[?#]/)[0];
+        item.url = window.location.href.split(/[?#]/)[0].replace("index.html", item.url).split('#')[0];
     });
 
     // If validatorUrl is not explicitly provided, disable the feature by setting to null
--- a/src/Swashbuckle.AspNetCore.SwaggerUI/index.js
+++ b/src/Swashbuckle.AspNetCore.SwaggerUI/index.js
@@ -36,7 +36,7 @@ window.onload = function () {
     // Workaround for https://github.com/swagger-api/swagger-ui/issues/5945
     configObject.urls.forEach(function (item) {
         if (item.url.startsWith("http") || item.url.startsWith("/")) return;
-        item.url = window.location.href.replace("index.html", item.url).split('#')[0];
+        item.url = window.location.href.replace("index.html", item.url).split(/[?#]/)[0];
     });
 
     // If validatorUrl is not explicitly provided, disable the feature by setting to null
--- a/test/Swashbuckle.AspNetCore.IntegrationTests/SwaggerUIIntegrationTests.cs
+++ b/test/Swashbuckle.AspNetCore.IntegrationTests/SwaggerUIIntegrationTests.cs
@@ -133,8 +133,7 @@ public class SwaggerUIIntegrationTests(ITestOutputHelper outputHelper)
 
         var jsContent = await response.Content.ReadAsStringAsync(cancellationToken);
 
-        Assert.Contains(".split(/[?#]/)[0]", jsContent);
-        Assert.DoesNotContain(".split('#')[0]", jsContent);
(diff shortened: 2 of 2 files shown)


src/Swashbuckle.AspNetCore.SwaggerUI/index.js
```diff
@@ -36,7 +36,7 @@ window.onload = function () {
     // Workaround for https://github.com/swagger-api/swagger-ui/issues/5945
     configObject.urls.forEach(function (item) {
         if (item.url.startsWith("http") || item.url.startsWith("/")) return;
-        item.url = window.location.href.replace("index.html", item.url).split('#')[0];
+        item.url = window.location.href.replace("index.html", item.url).split(/[?#]/)[0];
```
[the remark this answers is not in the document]
GabeSilvaDev: Fixed in 825d9728: the query string and fragment are now dropped from `window.location.href` before the document URL is spliced in, and the trailing `split('#')[0]` is kept as before, so a query string configured on the document URL itself (`v1/swagger.json?tenant=foo`) survives. The integration test now asserts that order.

test/Swashbuckle.AspNetCore.IntegrationTests/SwaggerUIIntegrationTests.cs
```diff
@@ -119,6 +119,23 @@ static void AssertResource(HttpResponseMessage response)
...
+        using var client = site.BuildClient();
+
+        using var response = await client.GetAsync("/index.js", cancellationToken);
+
+        Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+
+        var jsContent = await response.Content.ReadAsStringAsync(cancellationToken);
+
+        Assert.Contains("window.location.href.split(/[?#]/)[0].replace(\"index.html\", item.url)", jsContent);
+    }
```
martincostello: This test isn't particularly valuable - it's effectively just checking the code written in the other file is present in the returned file. It's not actually testing what that change itself is fixing.

A UI test (see [UITests](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/master/test/Swashbuckle.AspNetCore.IntegrationTests/UITests.cs)) with a browser that validates the underlying problem the issue wants to address would be better.

GabeSilvaDev: Replaced in c0021c0d with a Playwright test in `UITests`. It loads `swagger/index.html?foo=bar` against the `TopLevelSwaggerDoc` site (relative `v1.json` endpoint, so it goes through the code path in question) and asserts the document URL shown in the info panel ends with `/swagger/v1.json`. With the previous `index.js` the same test fails with `.../swagger/v1.json?foo=bar`. The `index.js` string assertion is gone.

martincostello (approved):

